PR #24, opened 2020-08-27
hw20: - done;

jah701:


src/main/java/core/basesyntax/impl/FruitReturn.java
```diff
@@ -0,0 +1,20 @@
...
+    private static final LocalDate NOW_DATE = LocalDate.now();
+
+    @Override
+    public void returnFruit(Fruit fruit) {
+        if (Storage.fruitStorage.peek() != null && FruitCounter.fruitsSold >= fruit.amount) {
+            if (fruit.getExpDate().isBefore(NOW_DATE)) {
+                return;
+            }
+            Storage.fruitStorage.peek().amount += fruit.amount;
+        }
```
AnaLanda, line 18: Here you're making the 'customer' return fruit to the 'store' when certain conditions are met, but I think we just need to take the data about what was returned from the CSV file and process it just like the two other operations (supply & buy)

src/main/java/core/basesyntax/impl/OrderExecutorService.java
```diff
@@ -0,0 +1,55 @@
...
+                        returner.returnFruit(fruit);
+                        break;
+                    default:
+                        throw new NullPointerException("Empty file");
+                }
+            }
+            System.out.println(Storage.fruitStorage.peek().getName() + ","
+                    + Storage.fruitStorage.peek().amount);
+        } catch (IOException e) {
+            e.printStackTrace();
```
AnaLanda, line 45: It is better to throw a new RunTimeException here

AnaLanda (commented):
Your OrderExecutoeService class performs many actions - it read data from the file, executes the orders based on this data, collects new data into a String. It would be better to create separate classes for each of these operations.
You should also add a class that will write the report about what fruit are left in the store into a new CSV file.

src/main/java/core/basesyntax/impl/FruitReturn.java
jah701: <3 @AnaLanda

PR #52, opened 2020-08-28
not finished yet

KostiantynVorobiov:


src/main/java/core/basesyntax/Storege.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.time.LocalDate;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storege {
```
vremere-a, line 7: поменяй имя на Storage

vremere-a: также Тревис ругается на слабое покрытие кода. 
1. поменяй в pom.xml 
 <element>PACKAGE</element> на  <element>BUNDLE</element>
таким образом, насколько я понял, ты меняешь принцип покрытие на классы, а не по линиям. 

2. также можешь занизить минимальный уровень покрытия теста на 0,7 например(так как у тебя метод Рейн будет с 0 покрытием)

KseniiaMakarova: @vremere-a минимальный уровень покрытия снижать нельзя

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,24 @@
...
+import core.basesyntax.servise.FileReadService;
+import core.basesyntax.servise.FileWriteService;
+import core.basesyntax.servise.FruitService;
+import core.basesyntax.servise.ReadFile;
+import core.basesyntax.servise.WriteFile;
+import java.util.List;
+
+/**
+ * Feel free to remove this class and create your own.
+ */
```
KseniiaMakarova, line 12: let's remove this

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,24 @@
...
+public class Main {
+    public static void main(String[] args) {
+        FileReadService readService = new ReadFile();
+        List<ProductsDto> lists = readService.readFromFile("src/test/resources/FruitTestOk.csv");
```
KseniiaMakarova, line 16:
```suggestion
        List<ProductsDto> productDtos = readService.readFromFile("src/test/resources/FruitTestOk.csv");
```

let's review variable and method names all over the project
so that they were more clear

src/main/java/core/basesyntax/servise/FileReadService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.servise;
+
+import core.basesyntax.ProductsDto;
+import java.util.List;
+
+public interface FileReadService {
+    List<ProductsDto> readFromFile(String fileName);
```
KseniiaMakarova, line 7: it would be even better if you had FileService with methods `read()` and `write()`
and a separate ProductDtoParser service that would parse `List<String>` to `List<ProductDto>`

```suggestion
    List<String> readFromFile(String fileName);
```

src/main/java/core/basesyntax/operations/FruitOperation.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.operations;
+
+import core.basesyntax.ProductsDto;
+
+public interface FruitOperation {
+    void fruitOperation(ProductsDto product);
```
KseniiaMakarova, line 6: let's name this method `apply()`
method names should be verbs

src/main/java/core/basesyntax/operations/ReturnOperation.java
```diff
@@ -0,0 +1,21 @@
...
+    @Override
+    public void fruitOperation(ProductsDto product) {
+        String fruit = product.getName();
+        int quantity = product.getQuantity();
+        LocalDate expirationDate = product.getDate();
+        if (TODAY.isAfter(expirationDate)) {
+            throw new RuntimeException("Spoiled fruit");
+        }
+        Storege.fruitStorage.merge(fruit, quantity, (oldVal, newVal) -> oldVal + newVal);
+        Storege.expirationDateStorage.put(fruit, expirationDate);
```
KseniiaMakarova, line 19: let's move this logic to `StorageService` class

we will have a `StorageService` field in each of our `FruitOperation` implementation
then, in `fruitOperation()` method (I recommended to rename it to `apply()`) we will simply call `storageService.addToStorage()` or `storageService.removeFromStorage()`

src/main/java/core/basesyntax/Storege.java
```diff
@@ -0,0 +1,10 @@
...
+    public static Map<String, Integer> fruitStorage = new HashMap<>();
+    public static Map<String, LocalDate> expirationDateStorage = new HashMap<>();
```
KseniiaMakarova, line 9: you may rename this class `StorageService` and implement the necessary methods here

the storage may remain the static field
I would only recommend to try to make it a single collection of items

you may want to read the task again to understand the storage logic

src/main/java/core/basesyntax/servise/FileWriteService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.servise;
+
+import java.util.Map;
+
+public interface FileWriteService {
+    void writeToFile(Map<String, Integer> fruitStorage);
```
KseniiaMakarova, line 6: let's pass here not `Map<String, Integer>` but `List<String>`, for example (products which are already parsed to a string form)

in your StorageService you can have `getAllRemainingFruit()` method

KseniiaMakarova (changes requested):

PR #54, opened 2020-08-30
added package operation, service and Product with storage. Added test

vladislavshvets:


src/main/java/core/basesyntax/service/FileReader.java
```diff
@@ -0,0 +1,20 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class FileReader {
+
+    public List<String> readLinesFromFile(String fileName) {
+        try {
```
fyarkiy, line 13: You can try to use " try with resources" (reco from Bohdan)

src/main/java/core/basesyntax/service/ParseData.java
```diff
@@ -0,0 +1,39 @@
...
+                } else if (!string.equals("type,fruit,quantity,date")) {
+                    throw new RuntimeException("Incorrect line: " + string);
+                }
+            }
+            return recordList;
+        }
+        throw new RuntimeException("Header line [" + HEADER + "] has not been found");
+    }
+
+    private boolean removeHeader(List<String> stringList) {
```
fyarkiy, line 26: I think  - this a bit complecated - I would just ignore first line. but certainly someone could pass file with no header or different header, so you might be right to check it.

src/main/java/core/basesyntax/service/TypeOfOperation.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.service;
+
+public enum TypeOfOperation {
+    s, b, r
```
fyarkiy, line 4: Enums values should be capital. And considering this - I think no pint to have Enums...

fyarkiy (approved):

KseniiaMakarova (changes requested):

PR #61, opened 2020-12-31
Solid task

ivlytvynets:
This commit is without unit tests. I'll do it later. Check, please, my project.


ivlytvynets:
now, with unit tests

NikolayFatkullin (approved):

src/main/java/core/basesyntax/model/objects/Plant.java
```diff
@@ -0,0 +1,54 @@
+package core.basesyntax.model.objects;
+
+import java.util.Objects;
+
+public class Plant {
+    protected String name;
+    protected Integer balance;
```
dmytro-hryhoruk, line 7: You cannot store information about the quantity of plants inside plant by itself

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.db;
+
+import java.util.List;
+
+public class Storage {
+    public static List<String> data;
```
dmytro-hryhoruk, line 6: these data only a Fruits names? how do you get a quantity?

src/main/java/core/basesyntax/model/impl/FruitStoreImpl.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax.model.impl;
+
+import core.basesyntax.dao.FruitsDao;
+import core.basesyntax.dao.FruitsDaoImpl;
+import core.basesyntax.model.objects.Fruit;
+import core.basesyntax.service.impl.FruitServiceImpl;
+import core.basesyntax.strategy.OperationStrategy;
+import java.util.ArrayList;
+
+public class FruitStoreImpl extends StoreImpl<Fruit> {
```
dmytro-hryhoruk, line 10: let's use only StoreImpl. Too many abstractions here

src/main/java/core/basesyntax/model/impl/StoreImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public void getStatistic(String fromFilePath, String toFilePath) {
+        storeService.getDataFromFile(fromFilePath);
+        plants = storeService.getPlantsBalance(plantsDao.getData());
+        FileWriterService fileWriterService = new FileWriterServiceImpl(toFilePath);
+        Report report = plants -> {
+            StringBuilder result = new StringBuilder();
+            result.append("fruit,quantity").append(System.lineSeparator());
```
dmytro-hryhoruk, line 34: Let's create constant

src/main/java/core/basesyntax/strategy/operation/BalanceOperationHandler.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.strategy.operation;
+
+public class BalanceOperationHandler implements OperationHandler {
+    @Override
+    public Integer updateBalance(Integer balance, Integer value) {
+        if (value < 0) {
+            throw new RuntimeException(String.format("Buyers will not be able to buy %s units. "
+                    + "%s is incorrect input.", value, value));
+        }
+        return balance + value;
```
dmytro-hryhoruk, line 10: You are using these handlers only to check the numbers. Each of this Hanler should has logic which will change Storage depends on operation type

dmytro-hryhoruk (changes requested):
Good job! 
To summarize all comments: Let's make exchanges of information  in main method;
Rad information from file —> pass it to other service wich will update Storage. Then get information from storage(via some service probably) and pass to other service which will write information to file and so on. All through the main method

PR #64, opened 2021-01-02
FruitsShops

vlykhvar:


src/main/java/core/basesyntax/dao/WarehouseImpl.java
```diff
@@ -0,0 +1,41 @@
+package core.basesyntax.dao;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class WarehouseImpl implements Warehouse {
+    private Map<String, Integer> listFruits = new HashMap<>();
```
ivlytvynets, line 7: In my opinion, it would be better to implement class Fruit with fields String name and int amount. Then you should create a list of fruits but not map. Let's imagine that you want to add to your shop vegetables. You need to create one more map, so your program is opened for modifications.

ivlytvynets (changes requested):

src/main/java/core/basesyntax/dao/ReportWorker.java
```diff
@@ -0,0 +1,56 @@
+package core.basesyntax.dao;
+
+import core.basesyntax.fileworker.InputOutputReport;
+import core.basesyntax.service.Validation;
+import java.time.LocalDate;
+import java.util.Map;
+
+public class ReportWorker implements WarehouseDao {
```
dmytro-hryhoruk, line 8: Dao is a layer in the application that is responsible only for operations related to the database. So let's change the name of this interface.

src/main/java/core/basesyntax/dao/WarehouseImpl.java
```diff
@@ -0,0 +1,41 @@
+package core.basesyntax.dao;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class WarehouseImpl implements Warehouse {
+    private final Map<String, Integer> listItems = new HashMap<>();
```
dmytro-hryhoruk, line 7: It's okay that you are using Map, But let's create model Fruit and use it as key to this map, even though it will have only one field

src/main/java/core/basesyntax/dao/ReportWorker.java
```diff
@@ -0,0 +1,56 @@
...
+    private final InputOutputReport inputOutputReport = new InputOutputReport();
```
dmytro-hryhoruk, line 9: Let's remove this dependecy from this class;
InputOutputReport will get the file name and reads information from file.
This information you can later pass to ReportWorker(it's method)(it's better to use create some models which will represent this information) and base on this information update Your DB(WareHouse);
Then ReportWorker will call the method to get a report from the database; And InputOutputReport will use this report and write it to file;
 For that, you need to create class which will represent a controller and from where you will call your services, pass result of one service and it's method to other service method and so on. It may be class with main method

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/model/Fruit.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax.model;
+
+import java.util.Objects;
+
+public class Fruit {
+    String name;
```
dmytro-hryhoruk, line 6: access modifier

src/main/java/core/basesyntax/service/ReportCheck.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.service;
+
+import core.basesyntax.strategy.OperationStrategy;
+import java.util.Map;
+
+public interface ReportCheck {
+    void readFromReport(String filName, Map<Procedure, OperationStrategy> operationStrategyMap);
```
dmytro-hryhoruk, line 7:
```suggestion
    void readFromReport(String fileName, Map<Procedure, OperationStrategy> operationStrategyMap);
```
Let's add a map via constructor

src/main/java/core/basesyntax/service/ReportWorker.java
```diff
@@ -0,0 +1,43 @@
...
+                operationStrategyMap.get(Procedure.findByProcedure(temp[operation]))
+                        .apply(new Fruit(temp[typeFruit]),
+                                Integer.parseInt(temp[amount]));
+            }
+        }
+    }
+
+    @Override
+    public String writeToReport() {
+        String topPattern = "fruit,quantity";
```
dmytro-hryhoruk, line 33: make it constant

src/main/java/core/basesyntax/service/ReportWorker.java
```diff
@@ -0,0 +1,43 @@
...
+        StringBuilder report = new StringBuilder();
+        report.append(topPattern).append(System.lineSeparator());
+        for (Map.Entry<Fruit, Integer> fruit
+                : DataBase.getListItems().entrySet()) {
+            report.append(fruit.getKey().getName()).append(",")
+                    .append(fruit.getValue().toString()).append(System.lineSeparator());
```
dmytro-hryhoruk, line 39: each append from the new line

src/test/java/core/basesyntax/dao/ReportWorkerTest.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ReportWorkerTest {
+    Map<Procedure, OperationStrategy> operationStrategyMap = new HashMap<>();
+    WarehouseImpl warehouseTest = new WarehouseImpl();
+    InputOutputReport inputOutputReport = new InputOutputReport();
```
dmytro-hryhoruk, line 23: access modifiers and let's initialize them in @BeforeClass method

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/service/ReportWorker.java
vlykhvar: done

src/main/java/core/basesyntax/model/Fruit.java
vlykhvar: done

src/main/java/core/basesyntax/service/ReportWorker.java
vlykhvar: done

src/test/java/core/basesyntax/dao/ReportWorkerTest.java
vlykhvar: done

src/main/java/core/basesyntax/service/ReportCheck.java
vlykhvar: done

dmytro-hryhoruk (approved):

PR #75, opened 2021-01-05
Updated solid task

ivlytvynets:


ivlytvynets:
I had some troubles with previous branch, so I created one more.

ivlytvynets:
I didn't create a enum of keys. In my opinion, the application doesn't depend on it.

dmytro-hryhoruk (changes requested):
Good job, almost there. 
Let's also add tests for FileReaderServiceImpl, FileWriterServiceImpl and FruitService

dmytro-hryhoruk (approved):

PR #92, opened 2021-01-06
added Fruit Shop implementation

oroboro-cpu:


src/main/java/core/basesyntax/service/impl/StoreServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public List<T> getFruitsBalance(List<String> data) {
+        List<T> fruits = new ArrayList<>();
+        Integer value;
+        String operation;
+        for (int i = 0; i < data.size(); i++) {
+            Fruit fruit = new Fruit();
+            operation = data.get(i).split(",")[0];
+            fruit.setName(data.get(i).split(",")[1]);
+            value = Integer.parseInt(data.get(i).split(",")[2]);
```
Fedich09, line 41:
```suggestion
            operation = data.get(i).split(",")[0];
            fruit.setName(data.get(i).split(",")[1]);
            value = Integer.parseInt(data.get(i).split(",")[2]);
```
we can add this like constant

Fedich09 (approved):

src/main/java/core/basesyntax/model/Fruit.java
```diff
@@ -0,0 +1,50 @@
+package core.basesyntax.model;
+
+import java.util.Objects;
+
+public class Fruit {
+    private Integer amount;
```
dmytro-hryhoruk, line 6: let's leave for now, but you can store information about objects quantity inside one of these objects

dmytro-hryhoruk (changes requested):
Let's also add tests for reader and writer

dmytro-hryhoruk (approved):

PR #99, opened 2021-01-07
initial

less08:


vicire (changes requested):
You also need to add writerImpl and test for him

PR #101, opened 2021-01-08
My shop

AndreiYazadzhi:


expectedReport
```diff
@@ -0,0 +1 @@
+Expected report
```
KseniiaMakarova, line 1: do we use this file anywhere? let's put in in the `resources` folder

KseniiaMakarova (changes requested):
sorry, the comments from the previous reviews are again ignored

PR #109, opened 2021-04-27
added solution

safordog:


src/main/java/core/basesyntax/service/action/BalanceActionHandler.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.service.action;
+
+public class BalanceActionHandler implements ActionHandler {
+    @Override
+    public int calculateQuantity(int currentQuantity, int inputQuantity) {
+        if (inputQuantity < 0) {
+            throw new RuntimeException("Fruit's balance can't be negative: " + inputQuantity);
+        }
+        return inputQuantity;
```
mankokolya, line 9: why are you not adding to the current balance? if the fruit already has some balance

src/main/java/core/basesyntax/service/reader/ReaderServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ReaderServiceImpl implements ReaderService {
+    @Override
+    public List<String> readFromFile(String inputDataFile) {
+        Path filePath = Paths.get(inputDataFile);
+        List<String> data = Collections.emptyList();
+        try {
+            data = Files.lines(filePath)
```
mankokolya, line 17: return Files.lines(Path.of(inputDataFile))
and you can delete return data because you will throw an exception if you will get exception here

src/main/java/core/basesyntax/service/FruitStoreServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+                .map(entry -> entry.getKey().getName() + COMMA_DELIMITER + entry.getValue())
+                .collect(Collectors.joining(System.lineSeparator())));
+        return stringBuilder.toString();
+    }
+
+    @Override
+    public void addDataToStorage(List<String> data) {
+        for (String record : data) {
+            String[] currentActionData = record.split(COMMA_DELIMITER);
+            String actionType = currentActionData[ACTION_INDEX];
```
mankokolya, line 40: you can insert it directly in action handler

src/main/java/core/basesyntax/service/FruitStoreServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+            String fruitName = currentActionData[FRUIT_NAME_INDEX];
+            Fruit fruit = new Fruit(fruitName);
```
mankokolya, line 42: same with fruit name

mankokolya: you already store these values in currentActionData no need to store them in separate local variable

mankokolya (changes requested):

src/main/java/core/basesyntax/service/FruitStoreServiceImpl.java
safordog: Дякую за перевірку. Корисні зауваження. DTO додавати не буду. Вважаю, що в даному випадку можна без нього.

mankokolya (approved):

src/main/java/core/basesyntax/service/reader/ReaderServiceImpl.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ReaderServiceImpl implements ReaderService {
+    @Override
+    public List<String> readFromFile(String inputDataFile) {
+        Path filePath = Paths.get(inputDataFile);
+        try {
+            return Files.lines(filePath)
+                    .skip(1)
```
KseniiaMakarova, line 16: we should be able to read from any file, not only csv, so let's skip the first line when we have all the lines and are already processing them

KseniiaMakarova (changes requested):
looks very nice, great job! added minor comments

KseniiaMakarova (approved):

PR #115, opened 2021-04-28
something scary

IhorRome:


IhorRome (commented):
ok, I know how to remove 5 files - will do it a little bit later

src/main/java/service/impl/FruitRecordParserServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    public FruitRecordParserServiceImpl(ActivityService activityService) {
+        this.activityService = activityService;
+    }
+
+    @Override
+    public List<FruitRecordDto> getRecord(List<String> stringsFromFile) {
+        List<FruitRecordDto> outputList = new ArrayList<>();
+        for (String string : stringsFromFile) {
+            new DataValidator().test(string);
+            String[] stringElements = string.split(",");
```
vuyko23, line 22: Ad this as the constant

src/main/java/service/impl/FruitRecordParserServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+            outputList.add(new FruitRecordDto(activityService.getActivity(stringElements[0]),
+                    stringElements[1],
+                    Integer.parseInt(stringElements[2])));
```
vuyko23, line 25: What happens if you try to convert "%2" to int?

src/main/java/service/impl/FruitRecordParserServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    public List<FruitRecordDto> getRecord(List<String> stringsFromFile) {
+        List<FruitRecordDto> outputList = new ArrayList<>();
+        for (String string : stringsFromFile) {
+            new DataValidator().test(string);
+            String[] stringElements = string.split(",");
+            outputList.add(new FruitRecordDto(activityService.getActivity(stringElements[0]),
+                    stringElements[1],
```
vuyko23, line 24: Make index as a constant

src/main/java/service/impl/ActivityServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+            case "b":
+                return Activity.BALANCE;
+            case "p":
+                return Activity.PURCHASE;
+            case "s":
+                return Activity.SUPPLY;
+            case "r":
+                return Activity.RETURN;
+            default:
+                throw new InvalidInputException("Invalid activity passed");
```
vuyko23, line 20: Make message constant

src/main/java/service/operation/RemoveOperation.java
```diff
@@ -0,0 +1,21 @@
...
+    private FruitsDao fruitsDao;
+
+    public RemoveOperation(FruitsDao fruitsDao) {
+        this.fruitsDao = fruitsDao;
+    }
+
+    @Override
+    public boolean apply(FruitRecordDto fruitRecordDto) {
+        Fruit fruit = fruitsDao.get(fruitRecordDto.getFruitType());
+        int newAmount = fruitsDao.getAmount(fruit) - fruitRecordDto.getAmount();
```
vuyko23, line 17: What happend if new amount gonna be -3?

vuyko23 (changes requested):

andriiromaniuk (approved):
Version 2.0 when?

src/main/java/service/impl/FruitRecordParserServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+    public static final String COMMA = ",";
+    public static final int LIST_TITLE_INDEX = 0;
+    public static final int OPERATION_INDEX = 0;
+    public static final int FRUIT_INDEX = 1;
+    public static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public List<FruitRecordDto> getRecord(List<String> stringsFromFile) {
+        List<FruitRecordDto> outputList = new ArrayList<>();
+        stringsFromFile.remove(LIST_TITLE_INDEX);
```
andriiromaniuk, line 20: Bad practice to change given object to you in method that is not intended to do that.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,44 @@
...
+import service.operation.AddOrCreateOperation;
+import service.operation.CreateOperation;
+import service.operation.Operation;
+import service.operation.RemoveOperation;
+
+public class Main {
+    public static void main(String[] args) {
+        FruitsDao fruitsDao = new FruitsDaoImpl();
+
+        Map<String, Operation> map = new HashMap<>();
```
RomanDubovskyi, line 28: Ideally, I would move initializing of map to the separate method

src/main/java/service/impl/FruitRecordParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+import exception.InvalidInputException;
+import java.util.ArrayList;
+import java.util.List;
+import model.FruitRecordDto;
+import service.FruitRecordParserService;
+import service.validator.DataValidator;
+
+public class FruitRecordParserServiceImpl implements FruitRecordParserService {
+    public static final String COMMA = ",";
```
RomanDubovskyi, line 11:
```suggestion
    public static final String CSV_PARSER = ",";
```
name variables according to their purpose not the value they hold

src/main/java/service/impl/FruitRecordParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    public static final int PROBABLY_TITLE_INDEX = 0;
```
RomanDubovskyi, line 12:
```suggestion
    public static final int TITLE_INDEX = 0;
```

src/main/java/service/impl/FruitRecordParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    public static final int VALID_STRING_INDEX = 1;
+    public static final int OPERATION_INDEX = 0;
+    public static final int FRUIT_INDEX = 1;
+    public static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public List<FruitRecordDto> getRecord(List<String> stringsFromFile) {
+        List<FruitRecordDto> outputList = new ArrayList<>();
+        DataValidator dataValidator = new DataValidator();
```
RomanDubovskyi, line 21: make this variable belong to class instead of method, we don't need to create new instance per each method call

src/main/java/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+import dao.FruitsDao;
+import java.util.Map;
+import model.Fruit;
+import service.ReportService;
+import service.file.FileWriter;
+
+public class ReportServiceImpl implements ReportService {
+    public static final String COLUMNS = "type,quantity";
+    public static final char COLUMN_SEPARATOR = ',';
```
RomanDubovskyi, line 11: in two classes you have this constant as public. You can reuse it

src/main/java/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+    public ReportServiceImpl(FileWriter fileWriter, FruitsDao fruitsDao) {
+        this.fileWriter = fileWriter;
+        this.fruitsDao = fruitsDao;
+    }
+
+    @Override
+    public void createReport(String filename) {
+        fileWriter.writeToFile(filename, COLUMNS);
+        for (Map.Entry<Fruit, Integer> entry : fruitsDao.getAll()) {
+            String reportLine = System.lineSeparator()
```
RomanDubovskyi, line 25: in the loop it's better to use string builder not string

src/main/java/service/operation/AddOperation.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public AddOperation(FruitsDao fruitsDao) {
+        this.fruitsDao = fruitsDao;
+    }
+
+    @Override
+    public boolean apply(FruitRecordDto fruitRecordDto) {
+        Fruit fruit = fruitsDao.get(fruitRecordDto.getFruitType());
+        int newAmount = fruitsDao.getAmount(fruit) + fruitRecordDto.getAmount();
+        try {
```
RomanDubovskyi, line 19: what happens if you return false ? 
If you process the report anyway it's better not to do this since some records will be ignored but user won't now that

src/store.csv
```diff
@@ -0,0 +1,9 @@
+type,fruit,quantity
+b,banana,20
+b,apple,100
+s,banana,100
+p,banana,13
+r,apple,10
+p,apple,20
+p,banana,5
+s,banana,50
```
RomanDubovskyi, line 9: missing an empty line at the end

RomanDubovskyi (changes requested):

src/main/java/service/impl/FruitRecordParserServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+
+import exception.InvalidInputException;
+import java.util.ArrayList;
+import java.util.List;
+import model.FruitRecordDto;
+import service.FruitRecordParserService;
+import service.validator.DataValidator;
+
+public class FruitRecordParserServiceImpl implements FruitRecordParserService {
+    public static final int TITLE_INDEX = 0;
```
RomanDubovskyi, line 11: If you don't use them outside the class change modifier to private

src/main/java/service/validator/DataValidator.java
```diff
@@ -0,0 +1,36 @@
+package service.validator;
+
+import exception.InvalidInputException;
+import java.util.function.Predicate;
+
+public class DataValidator implements Predicate<String> {
+    public static final int DATA_ELEMENTS = 3;
```
RomanDubovskyi, line 7: same here, reuse existing constants or make yours private

RomanDubovskyi (approved):
don't forget to fix modifiers

PR #119, opened 2021-04-28
jv-fruit-shop, initial commit

fotomassimo:


Reastador (approved):

src/main/java/fruitshop/Main.java
```diff
@@ -0,0 +1,50 @@
...
+import fruitshop.service.shopoperation.StorageIncreaseHandler;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    public static void main(String[] args) {
+        Map<OperationType, OperationHandler> operationMap = new HashMap<>();
+        operationMap.put(OperationType.BALANCE,
+                new StorageIncreaseHandler(new FruitStorageDaoImpl()));
```
RomanDubovskyi, line 28: declare FruitStorageDaoImpl as a seperate local variable, we don't need 4 different instances

src/main/java/fruitshop/model/dto/ReportDto.java
```diff
@@ -0,0 +1,18 @@
+package fruitshop.model.dto;
+
+import fruitshop.model.Fruit;
+import java.math.BigDecimal;
+import java.util.Map;
+import java.util.Set;
+
+public class ReportDto {
```
RomanDubovskyi, line 8: this class looks redundant, you can just get entries from you dao and then do what you need with it

src/main/resources/input.csv
```diff
@@ -0,0 +1,9 @@
+type,fruit,quantity
+b,banana,100
+b,apple,100
+s,banana,100
+p,banana,13
+r,apple,10
+p,apple,20
+p,banana,5
+s,banana,50
```
RomanDubovskyi, line 9: add an empty line at the end

RomanDubovskyi (changes requested):
good job, left minor comments

src/main/java/fruitshop/Main.java
fotomassimo: done

src/main/resources/input.csv
fotomassimo: done

src/main/java/fruitshop/model/dto/ReportDto.java
fotomassimo: done

RomanDubovskyi (approved):

PR #127, opened 2021-05-03
Done Homework.

DeStinyShell:


src/main/java/datavalidation/DataValidator.java
```diff
@@ -0,0 +1,33 @@
...
+        if (!ListOfFruits.contains(dataType.toUpperCase())) {
+            throw new RuntimeException("Such fruit does not exist "
+                    + dataType);
+        }
+        dataType = data.split(SEPARATOR)[AMOUNT_INDEX];
+        if (Integer.parseInt(dataType) < 0) {
+            throw new RuntimeException("Amount can't be less than zero.");
+        }
+        return true;
+    }
```
LayRipper, line 32: what if our amount is "2ef@14!"

src/main/java/datavalidation/DataValidator.java
```diff
@@ -0,0 +1,33 @@
+package datavalidation;
+
+import fruitsassortment.ListOfFruits;
+import shopoperations.ListOfOperations;
+
+public class DataValidator implements DataValidation {
```
LayRipper, line 6:
```suggestion
public class DataValidatorImpl implements DataValidator {
```

src/main/java/fruitsassortment/ListOfFruits.java
```diff
@@ -0,0 +1,15 @@
...
+    APPLE;
+
+    public static boolean contains(String name) {
+        for (ListOfFruits fruitName : ListOfFruits.values()) {
+            if (fruitName.toString().equals(name)) {
+                return true;
+            }
+        }
+        return false;
+    }
```
LayRipper, line 14: this might be called overengineering, we should assume that all the fruits passed are valid

src/main/java/shopdao/FruitDao.java
```diff
@@ -0,0 +1,12 @@
...
+import fruitsassortment.Fruit;
+import java.util.Map;
+
+public interface FruitDao {
+    boolean add(Fruit fruit, int amount);
+
+    int get(Fruit fruit);
+
+    Map<Fruit, Integer> getAll();
+}
```
LayRipper, line 12: We essentially have this method in Storage, why do we need it once again

src/main/java/shopoperations/ListOfOperations.java
```diff
@@ -0,0 +1,17 @@
...
+    R;
+
+    public static boolean contains(String value) {
+        for (ListOfOperations operation : ListOfOperations.values()) {
+            if (operation.toString().equals(value)) {
+                return true;
+            }
+        }
+        return false;
+    }
```
LayRipper, line 16: is it ok having a method within the enum class?

src/main/java/shopservice/CsvWriteServiceImpl.java
```diff
@@ -0,0 +1,16 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class CsvWriteServiceImpl implements WriteService {
+    @Override
+    public void writeToFile(String data, String filePath) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(filePath))) {
+            bufferedWriter.write(data);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to path " + filePath);
```
LayRipper, line 13:
```suggestion
            throw new RuntimeException("Can't write to path " + filePath, e);
```

src/main/java/shopdao/FruitDaoImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import dbimitation.Storage;
+import fruitsassortment.Fruit;
+import java.util.Map;
+
+public class FruitDaoImpl implements FruitDao {
+    @Override
+    public boolean add(Fruit fruit, int amount) {
+        Storage.getFruits().put(fruit, amount);
+        return true;
+    }
```
LayRipper, line 12: will this method ever return false? why do we need it boolean then?

src/main/java/shopservice/StoreServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+        Map<Fruit, Integer> products = fruitDao.getAll();
+        for (Map.Entry<Fruit, Integer> productsEntry : products.entrySet()) {
+            report.append(System.lineSeparator())
+                    .append(productsEntry.getKey().getName())
+                    .append(SEPARATOR)
+                    .append(productsEntry.getValue());
+        }
+        return report.toString();
+    }
+}
```
LayRipper, line 50: One class should be responsible for one logical operation, these two methods should be in separate classes

src/main/java/shopoperations/ListOfOperations.java
```diff
@@ -0,0 +1,17 @@
...
+    P,
+    R;
+
+    public static boolean contains(String value) {
+        for (ListOfOperations operation : ListOfOperations.values()) {
+            if (operation.toString().equals(value)) {
+                return true;
+            }
+        }
+        return false;
```
LayRipper, line 15: contains method has no interface, even though it was used twice

LayRipper (changes requested):

LayRipper (approved):

src/main/java/datavalidation/DataValidatorImpl.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class DataValidatorImpl implements DataValidator {
+    private static final int OPERATION_TYPE_INDEX = 0;
+    private static final int PRODUCT_NAME_INDEX = 1;
+    private static final int AMOUNT_INDEX = 2;
+    private static final String SEPARATOR = ",";
+
+    @Override
+    public boolean validateData(String data) {
+        if ("type,fruit,quantity".equals(data.trim())) {
```
RomanDubovskyi, line 14: let's declare this string as variable

src/main/java/datavalidation/DataValidatorImpl.java
```diff
@@ -0,0 +1,38 @@
...
+            return true;
+        }
+        String dataType = data.split(SEPARATOR)[OPERATION_TYPE_INDEX].toUpperCase().trim();
+        if (!ListOfOperations.contains(dataType)) {
+            throw new RuntimeException("Such operation is not valid "
+                    + dataType);
+        }
+        dataType = data.split(SEPARATOR)[PRODUCT_NAME_INDEX].trim();
+        if (!ListOfFruits.contains(dataType.toUpperCase())) {
```
RomanDubovskyi, line 23: I wouldn't hardcode list of fruits in the code, if you need to change the code each time you add / remove some fruit someone always needs to maintain the code. In future it will be handled by DB but for now I would remove this check

src/main/java/fruitsassortment/ListOfFruits.java
```diff
@@ -0,0 +1,15 @@
+package fruitsassortment;
+
+public enum ListOfFruits {
+    BANANA,
```
RomanDubovskyi, line 4: String is enough here, fruits are not supposed to be constant forever

src/main/java/shopservice/CsvReadServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class CsvReadServiceImpl implements ReadService {
+
+    @Override
+    public List<String> readFromFile(String filePath) {
+        List<String> dataFromFile = new ArrayList<>();
+        try (BufferedReader readFile = new BufferedReader(new FileReader(filePath))) {
+            String data = readFile.readLine();
+            while ((data != null)) {
+                new DataValidatorImpl().validateData(data);
```
RomanDubovskyi, line 18: I would rather parse data after reading to keep open connnection as short as possible

src/main/java/shopservice/CsvWriteServiceImpl.java
```diff
@@ -0,0 +1,16 @@
+package shopservice;
+
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class CsvWriteServiceImpl implements WriteService {
```
RomanDubovskyi, line 7:
```suggestion
public class FileWriterServiceImpl implements WriteService {
```
in this impl you don't do anything special for csv file. It can be used for any file (theoretically only report will differ)

src/main/java/shopservice/ReportCompiler.java
```diff
@@ -0,0 +1,5 @@
+package shopservice;
+
+public interface ReportCompiler {
```
RomanDubovskyi, line 3:
```suggestion
public interface ReportService {
```

RomanDubovskyi (changes requested):

src/main/java/dbimitation/Storage.java
```diff
@@ -0,0 +1,13 @@
+package dbimitation;
+
+import fruitsassortment.Fruit;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    private static final Map<Fruit, Integer> FRUIT_MAP = new HashMap<>();
```
RomanDubovskyi, line 8:
```suggestion
    private static final Map<Fruit, Integer> fruits = new HashMap<>();
```
it's not a constant unless you use Immutable map, so we use regular naming here

src/main/java/shopservice/CsvReadServiceImpl.java
```diff
@@ -0,0 +1,28 @@
...
+public class CsvReadServiceImpl implements ReadService {
+
+    @Override
+    public List<String> readFromFile(String filePath) {
+        List<String> dataFromFile;
+        int dataIndex = 0;
+        try {
+            dataFromFile = Files.readAllLines(Path.of(filePath));
+            String data;
+            while (dataIndex != dataFromFile.size()) {
```
RomanDubovskyi, line 18: move validation to separate class to follow SRP

src/main/resources/report.csv
```diff
@@ -0,0 +1,4 @@
+fruit,amount
+banana,40
+apple,40
+aple,10
```
RomanDubovskyi, line 4:
```suggestion
aple,10

```

src/main/resources/file.csv
```diff
@@ -0,0 +1,6 @@
+type,fruit,quantity
+b, apple ,  60
+b,banana,60
+p,apple,20
+p,banana,20
+b, aple, 10
```
RomanDubovskyi, line 6: mistype
```suggestion
b, apple, 10
```

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #131, opened 2021-05-03
first solution:

andriiromaniuk:


src/main/java/service/DataValidatorImpl.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public void validateRecord(String record) {
+        String[] data = record.split(COMMA);
+        String action = data[0];
+        try {
+            Integer.parseInt(data[2]);
+        } catch (NumberFormatException e) {
+            throw new InvalidInputException("Wrong input, unparsable value of fruits");
+        }
+        int amount = Integer.parseInt(data[2]);
```
IhorRome, line 21:
```suggestion
            int amount = Integer.parseInt(data[2]);
        } catch (NumberFormatException e) {
            throw new InvalidInputException("Wrong input, unparsable value of fruits");
        }
```

src/main/java/service/actions/Return.java
```diff
@@ -0,0 +1,22 @@
+package service.actions;
+
+import dao.FruitDao;
+import dao.FruitDaoImpl;
+import model.Fruit;
+import model.FruitDataDto;
+
+public class Return implements ActivityHandler {
```
IhorRome, line 8: it`s similar to supply - delete one and change map in Main

src/main/resources/fruits.csv
```diff
@@ -0,0 +1,8 @@
+b,banana,20
+b,apple,100
+s,banana,100
+p,banana,13
+r,apple,10
+p,apple,20
+p,banana,5
+s,banana,50
```
IhorRome, line 8: default file has title - you have to deal with it somehow
```suggestion
operation,fruit,quantity
b,banana,20
b,apple,100
s,banana,100
p,banana,13
r,apple,10
p,apple,20
p,banana,5
s,banana,50
```

IhorRome (changes requested):

IhorRome (approved):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,52 @@
...
+import service.file.FileWriter;
+import service.file.FileWriterImpl;
+
+public class Main {
+    public static final String FILE_TO_READ = "src\\main\\resources\\fruits.csv";
+    public static final String FILE_TO_WRITE = "src\\main\\resources\\report.csv";
+    public static final String BALANCE = "b";
+    public static final String PURCHASE = "p";
+    public static final String RETURN = "r";
+    public static final String SUPPLY = "s";
```
RomanDubovskyi, line 31: it's better to move operations initialization to the separate class

src/main/java/dao/FruitDao.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Map;
+import model.Fruit;
+
+public interface FruitDao {
+    void add(Fruit fruit, Integer amount);
+
+    Integer get(Fruit fruit);
+
+    Map<Fruit, Integer> getDB();
```
RomanDubovskyi, line 11: this method looks strange in DAO layer, why do we need it if just can getDB and you put and get directly on Map. I would replace it with getAll() method to return all data if we need smt like this in code

src/main/java/db/Storage.java
```diff
@@ -0,0 +1,21 @@
...
+import model.Fruit;
+
+public class Storage {
+    private static final Map<Fruit, Integer> fruits = new HashMap<>();
+
+    public static Map<Fruit, Integer> getFruits() {
+        return fruits;
+    }
+
+    public static Map<Fruit, Integer> getDB() {
```
RomanDubovskyi, line 14: that's overengineering

src/main/java/service/DataParserImpl.java
```diff
@@ -0,0 +1,33 @@
+package service;
+
+import java.util.ArrayList;
+import java.util.List;
+import model.FruitDataDto;
+
+public class DataParserImpl implements DataParser {
+
+    public static final String COMMA = ",";
```
RomanDubovskyi, line 9:
```suggestion
    public static final String CSV_SEPARATOR = ",";
```
name variables according to their purpose not the value they hold

src/main/java/service/DataParserImpl.java
```diff
@@ -0,0 +1,33 @@
...
+    public static final int FRUIT_INDEX = 1;
+    public static final int QUANTITY_INDEX = 2;
+    public static final int DATA_INDEX = 1;
+    public static final String WHITE_SPACE = " ";
+    public static final int FIRST_LINE_INDEX = 0;
+
+    @Override
+    public List<FruitDataDto> parseData(List<String> data) {
+        List<FruitDataDto> parsedData = new ArrayList<>();
+        DataValidator dataValidator = new DataValidatorImpl();
```
RomanDubovskyi, line 20: make DataValidator class variable instead of local not to create its instance per each method call

src/main/java/service/DataParserImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class DataParserImpl implements DataParser {
+
+    public static final String COMMA = ",";
+    public static final int OPERATION_INDEX = 0;
+    public static final int FRUIT_INDEX = 1;
+    public static final int QUANTITY_INDEX = 2;
+    public static final int DATA_INDEX = 1;
+    public static final String WHITE_SPACE = " ";
+    public static final int FIRST_LINE_INDEX = 0;
```
RomanDubovskyi, line 15: if you don't use these constants outside of this class make them private, if you use them don't redeclare same variable several times

src/main/java/service/DataValidatorImpl.java
```diff
@@ -0,0 +1,47 @@
...
+public class DataValidatorImpl implements DataValidator {
+
+    public static final String COMMA = ",";
+    public static final String BALANCE = "b";
+    public static final String PURCHASE = "p";
+    public static final String RETURN = "r";
+    public static final String SUPPLY = "s";
+    public static final int ACTIVITY_INDEX = 0;
+    public static final int QUANTITY_INDEX = 2;
+    public static final String CORRECT_FIRST_LINE = "activity,fruit,quantity";
```
RomanDubovskyi, line 14:
```suggestion
    public static final String CORRECT_TITLE = "activity,fruit,quantity";
```

src/main/java/dao/FruitDaoImpl.java
```diff
@@ -0,0 +1,22 @@
...
+import model.Fruit;
+
+public class FruitDaoImpl implements FruitDao {
+    @Override
+    public void add(Fruit fruit, Integer amount) {
+        Storage.getFruits().put(fruit, amount);
+    }
+
+    @Override
+    public Integer get(Fruit fruit) {
```
RomanDubovskyi, line 14: It's better to return Optional here to reduce chance of NPE

src/main/resources/fruits.csv
```diff
@@ -0,0 +1,9 @@
+activity,fruit,quantity
+b,banana,20
+b,apple,100
+s,banana,100
+p,banana,13
+r,apple,10
+p,apple,20
+p,banana,5
+s,banana,50
```
RomanDubovskyi, line 9: empty line is missing

RomanDubovskyi (changes requested):

src/main/java/service/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,32 @@
+package service;
+
+import dao.FruitDao;
+import java.util.Map;
+import model.Fruit;
+import service.file.FileWriter;
+
+public class ReportGeneratorImpl implements ReportGenerator {
+    public static final String COMMA = ",";
```
RomanDubovskyi, line 9:
```suggestion
    public static final String CSV_SEPARATOR = ",";
```

src/main/java/service/file/FileReaderImpl.java
```diff
@@ -0,0 +1,19 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+
+public class FileReaderImpl implements FileReader {
+    @Override
+    public List<String> readFile(String fileName) {
+        List<String> data;
+        try {
+            data = Files.readAllLines(Paths.get(fileName));
```
RomanDubovskyi, line 13:
```suggestion
            return Files.readAllLines(Paths.get(fileName));
```

RomanDubovskyi (approved):

PR #139, opened 2021-05-04
Hw 14

MakeDreamCode:


src/main/java/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+import model.Fruit;
+import service.ReportService;
+
+public class ReportServiceImpl implements ReportService {
+    private static final char NEW_LINE_CHAR = '\n';
+    private static final char SEPARATOR_CHAR = ',';
+    private static final String ERROR_MESSAGE = "Can`t write to file ";
+
+    @Override
+    public void makeStockReportToCsvFile(String outputFileName) {
```
mankokolya, line 18: this method should only generate your String report it shouldn't write to file. Rename it and delete writing it is the job of separate service

src/main/java/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,21 @@
+package service.impl;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+import service.ReaderService;
+
+public class ReaderServiceImpl implements ReaderService {
```
mankokolya, line 9: create one FileService that read data from file and write to it

src/main/java/core/basesyntax/ApplicationRun.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static void main(String[] args) {
+        ReaderService readerService = new ReaderServiceImpl();
+        List<String> stringsFromFile = readerService.readFromFile(PATH_SOURCE_FILE);
+        ParserService parserService = new ParserServiceImpl();
+        List<FruitRecordDto> fruitDtos = parserService.parseToDto(stringsFromFile);
+        for (FruitRecordDto dto : fruitDtos) {
+            fruitOperationsServiceMap.get(dto.getOperationType())
+                                                       .apply(dto);
+        }
```
mankokolya, line 40: move this to service that saves your dto to database

mankokolya: this service should take as parameter not map but another interface with method get() that return the proper handler

src/main/java/db/Storage.java
```diff
@@ -0,0 +1,12 @@
+package db;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import model.Fruit;
+
+public class Storage {
+    public static final Map<Fruit, Integer> fruits = new HashMap<>();
+    public static final Set<String> validFruitNames = Set.of("apple", "banana");
```
mankokolya, line 10: what this set is doing? You don't need it here

src/main/java/db/Storage.java
```diff
@@ -0,0 +1,12 @@
...
+    public static final Set<String> validOperations = Set.of("b", "p", "r", "s");
```
mankokolya, line 11: move this operations to enum

src/main/java/model/dto/FruitRecordDto.java
```diff
@@ -0,0 +1,29 @@
+package model.dto;
+
+public class FruitRecordDto {
+    private String operationType;
```
mankokolya, line 4: change String to enum type

src/main/java/model/dto/FruitRecordDto.java
```diff
@@ -0,0 +1,29 @@
...
+        return fruitName;
+    }
+
+    public Integer getQuantity() {
+        return quantity;
+    }
+
+    public void setQuantity(Integer quantity) {
+        this.quantity = quantity;
+    }
```
mankokolya, line 28: it is not a good idea to change you dto

mankokolya: it is already fixed in file. Parse it as it is

src/main/java/service/impl/RemoveOperation.java
```diff
@@ -0,0 +1,19 @@
...
+public class RemoveOperation implements FruitOperationsService {
+    @Override
+    public int apply(FruitRecordDto fruitRecordDto) {
+        if (Storage.fruits
+                .get(new Fruit(fruitRecordDto.getFruitName())) < fruitRecordDto.getQuantity()) {
+            throw new RuntimeException(fruitRecordDto.getFruitName() + " out of Stock!");
+        }
+        fruitRecordDto.setQuantity(-fruitRecordDto.getQuantity());
+        FruitOperationsService fruitOperationsService = new AddOperation();
+        return fruitOperationsService.apply(fruitRecordDto);
```
mankokolya, line 17: it is not readable. It took me some time to understand and you are changing you dto just subtract and save here. The number of lines will be the same but it will be more readable

src/main/java/core/basesyntax/ApplicationRun.java
```diff
@@ -0,0 +1,44 @@
...
+    private static final String SUPPLY = "s";
+    private static final String PURCHASE = "p";
+    private static final String RETURN = "r";
+    private static final String PATH_SOURCE_FILE = "src/main/resources/input.csv";
+    private static final String OUTPUT_FILE_NAME = "src/main/resources/output.csv";
+    private static final Map<String, FruitOperationsService> fruitOperationsServiceMap
+            = new HashMap<>();
+
+    static {
+        fruitOperationsServiceMap.put(SUPPLY, new AddOperation());
```
mankokolya, line 27: You create new AddOperation() two times and will do one more time for balance just create one object and pass where needed

src/main/java/service/impl/AddOperation.java
```diff
@@ -0,0 +1,17 @@
...
+import db.Storage;
+import model.Fruit;
+import model.dto.FruitRecordDto;
+import service.FruitOperationsService;
+
+public class AddOperation implements FruitOperationsService {
+    @Override
+    public int apply(FruitRecordDto fruitRecordDto) {
+        Fruit currentFruit = new Fruit(fruitRecordDto.getFruitName());
+        int currentQuantityInStock = Storage.fruits.get(currentFruit);
```
mankokolya, line 12: what you will get if you call get(not existing in db fruit)? You can use getOrDefault(0);

src/main/java/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final int CORRECT_FIELDS_NUMBER = 3;
+    private static final String BALANCE_OPERATION = "b";
+    private static final String DELIMITER = ",";
+    private static final String ERROR_FIELDS_NUMBER = "Invalid number of fields!";
+    private static final String ERROR_FIELDS = "Invalid one or more fields!";
+    private static final String REGEX_ONLY_DIGITS = "\\d+";
+
+    @Override
+    public List<FruitRecordDto> parseToDto(List<String> strings) {
+        List<FruitRecordDto> fruitRecordDtoList = new ArrayList<>(strings.size());
```
mankokolya, line 23: no need to indicate size here

src/main/java/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        for (String string : strings) {
+            String[] fieldsDto = string.split(DELIMITER);
+            if (fieldsDto.length != CORRECT_FIELDS_NUMBER) {
+                throw new RuntimeException(ERROR_FIELDS_NUMBER);
+            }
+            String operation = fieldsDto[OPERATION_TYPE];
+            String fruitName = fieldsDto[FRUIT_NAME];
+            String fruitQuantity = fieldsDto[QUANTITY];
+            if (Storage.validOperations.contains(operation)
+                    && Storage.validFruitNames.contains(fruitName)
```
mankokolya, line 33: how do you know which fruit names are valid?

src/main/java/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public List<FruitRecordDto> parseToDto(List<String> strings) {
+        List<FruitRecordDto> fruitRecordDtoList = new ArrayList<>(strings.size());
+        for (String string : strings) {
+            String[] fieldsDto = string.split(DELIMITER);
+            if (fieldsDto.length != CORRECT_FIELDS_NUMBER) {
+                throw new RuntimeException(ERROR_FIELDS_NUMBER);
+            }
+            String operation = fieldsDto[OPERATION_TYPE];
```
mankokolya, line 29: using enum will throw you exception automatically if you will try to get by wrong type

src/main/java/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+            String fruitName = fieldsDto[FRUIT_NAME];
+            String fruitQuantity = fieldsDto[QUANTITY];
```
mankokolya, line 31: Integer.parseInt will also throw exception if it is not number

src/main/java/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+            if (Storage.validOperations.contains(operation)
+                    && Storage.validFruitNames.contains(fruitName)
+                    && fruitQuantity.matches(REGEX_ONLY_DIGITS)) {
+                if (BALANCE_OPERATION.equals(operation)) {
+                    Storage.fruits.put(new Fruit(fruitName), Integer.parseInt(fruitQuantity));
+                } else {
```
mankokolya, line 37: remove this from parser to handler

mankokolya (changes requested):

mankokolya (approved):

src/main/java/model/dto/FruitRecordDto.java
```diff
@@ -0,0 +1,27 @@
+package model.dto;
+
+import model.OperationType;
+
+public class FruitRecordDto {
+    private OperationType operationType;
+    private String fruitName;
```
Max-Pochepets, line 7:
```suggestion
    private Fruit fruit;
```

src/main/java/service/impl/AddOperation.java
```diff
@@ -0,0 +1,17 @@
+package service.impl;
+
+import db.Storage;
+import model.Fruit;
+import model.dto.FruitRecordDto;
+import service.OperationHandler;
+
+public class AddOperation implements OperationHandler {
+    @Override
+    public int apply(FruitRecordDto fruitRecordDto) {
```
Max-Pochepets, line 10: what if we add "-10"?

Max-Pochepets (changes requested):
nice

Max-Pochepets (approved):

PR #151, opened 2021-05-05
add solution

KononenkoKonstantin:


src/main/java/core/basesyntax/services/impl/FileReaderImpl.java
```diff
@@ -0,0 +1,28 @@
...
+    private static final String EXCEPTION_MESSAGE = "Can't read file ";
+
+    @Override
+    public List<String> getLinesFromFile(String fileName) {
+        String path = fileName;
+        List<String> dataFromFile = new ArrayList<>();
+        File file = new File(path);
+        if (file.exists()) {
+            try {
+                dataFromFile = Files.readAllLines(Path.of(path));
```
anufriiev, line 21: I think better to read line by line, because it can be null.
```suggestion
                 BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName));
            String line = bufferedReader.readLine();
            while (line != null) {
                System.out.println(line);
                dataFromFile.add(line);
                line = bufferedReader.readLine();
```

src/main/java/core/basesyntax/services/impl/FileWriterImpl.java
```diff
@@ -0,0 +1,29 @@
...
+import java.io.IOException;
+
+public class FileWriterImpl implements FileWriter {
+    private static final String CREATE_FILE_EXCEPTION_MESSAGE = "Can't create new file ";
+    private static final String WRITE_FILE_EXCEPTION_MESSAGE = "Can't write file";
+
+    @Override
+    public boolean write(String dataToWrite, String fileName) {
+        File file = new File(fileName);
+        if (!file.exists()) {
```
anufriiev, line 15: Here same as in FileReader

src/main/java/core/basesyntax/services/impl/FileReaderImpl.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class FileReaderImpl implements FileReader {
+    private static final String EXCEPTION_MESSAGE = "Can't read file ";
+
+    @Override
+    public List<String> getLinesFromFile(String fileName) {
+        String path = fileName;
+        List<String> dataFromFile = new ArrayList<>();
+        File file = new File(path);
+        if (file.exists()) {
```
anufriiev, line 19: I'm not sure we have to check the file for existing here, while we using try-catch construction and throw IOException, but I think it's not a mistake.

anufriiev (changes requested):

src/main/java/core/basesyntax/services/impl/FileReaderImpl.java
KononenkoKonstantin: I think, my code is right

src/main/java/core/basesyntax/services/impl/FileWriterImpl.java
KononenkoKonstantin: I create a new file for report, if it not exists. I would have exception, if file is not exists

src/main/java/core/basesyntax/services/impl/FileReaderImpl.java
KononenkoKonstantin: I agree

anufriiev (approved):

src/main/java/core/basesyntax/model/OperationType.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax.model;
+
+public enum OperationType {
+    BALANCE("b"), SUPPLY("s"), PURCHASE("p"), RETURN("r");
```
Max-Pochepets, line 4: let's make it a column, for better readability

Max-Pochepets (approved):

PR #158, opened 2021-06-14
fruit shop implementation

SakVladyslav:


src/main/java/dao/RecordDaoImpl.java
```diff
@@ -0,0 +1,30 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class RecordDaoImpl implements RecordDao {
+    @Override
+    public List<String> readFile(String fileName) {
+        List<String> result;
+        try {
+            result = Files.readAllLines(Path.of(fileName));
```
RomanDubovskyi, line 15:
```suggestion
            return Files.readAllLines(Path.of(fileName));
```

src/main/java/services/ActionTypeStrategy.java
```diff
@@ -0,0 +1,7 @@
+package services;
+
+import services.actions.ActionHandler;
+
+public interface ActionTypeStrategy {
+    ActionHandler get(String type);
```
RomanDubovskyi, line 6: This basically repeates purpose of the map, I believe we can remove this interface an impl

src/main/java/services/ReportImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    private final ActionTypeStrategy actionTypeStrategy;
+
+    public ReportImpl(ActionTypeStrategy actionTypeStrategy) {
+        this.actionTypeStrategy = actionTypeStrategy;
+    }
+
+    @Override
+    public String createReport(List<String> records) {
+        StringBuilder report = new StringBuilder();
+        RecordValidation recordValidation = new RecordValidation();
```
RomanDubovskyi, line 18: Let's make this object a class field instead of local one (we don't need to create separate object per each method call

src/main/java/services/actions/ActionHandler.java
```diff
@@ -0,0 +1,5 @@
+package services.actions;
+
+public interface ActionHandler {
+    void getResultOfAction(String fruitName, int fruitCount);
```
RomanDubovskyi, line 4: void methods are harder to test, let's make this method return updated result

src/main/java/services/actions/ReturnAndSupplyHandler.java
```diff
@@ -0,0 +1,10 @@
+package services.actions;
+
+import db.Storage;
+
+public class ReturnAndSupplyHandler implements ActionHandler {
```
RomanDubovskyi, line 5:
```suggestion
public class IncreaseFruitHandler implements ActionHandler {
```
We may have more operations handled with this class

RomanDubovskyi (changes requested):

src/main/java/dao/RecordDaoImpl.java
SakVladyslav: Didn`t quite understand why the interface should be removed. + this interface showed it in the lesson as well. Could you give more details, please, for my understand?

RomanDubovskyi (approved):

PR #187, opened 2021-09-02
created a program to for a report for a fruit shop from input file

AlexRozv:
tried to stick with SOLID rules to implement all methods and form a report

### DIFF
--- a/src/main/java/core/basesyntax/model/Operation.java
+++ b/src/main/java/core/basesyntax/model/Operation.java
@@ -1,0 +1,25 @@
+package core.basesyntax.model;
+
+public class Operation {
+    private final String type;
+    private final String fruit;
+    private final int amount;
+
+    public Operation(String type, String fruit, int amount) {
+        this.type = type;
+        this.fruit = fruit;
+        this.amount = amount;
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public String getFruit() {
+        return fruit;
+    }
+
+    public int getAmount() {
+        return amount;
+    }
+}
--- a/src/main/java/core/basesyntax/service/AmountCalculator.java
+++ b/src/main/java/core/basesyntax/service/AmountCalculator.java
@@ -1,0 +1,5 @@
+package core.basesyntax.service;
+
+public interface AmountCalculator {
+    void calculate();
+}
--- a/src/main/java/core/basesyntax/service/OperationStrategyImpl.java
+++ b/src/main/java/core/basesyntax/service/OperationStrategyImpl.java
@@ -1,0 +1,20 @@
+package core.basesyntax.service;
+
+import core.basesyntax.service.operation.OperationHandler;
+import java.util.Map;
+
+public class OperationStrategyImpl implements OperationStrategy {
+    private Map<String, OperationHandler> operationHandlerMap;
+
+    public OperationStrategyImpl(Map<String, OperationHandler> operationHandlerMap) {
+        this.operationHandlerMap = operationHandlerMap;
+    }
+
+    @Override
+    public OperationHandler get(String operation) {
+        if (operationHandlerMap.get(operation) == null) {
+            throw new RuntimeException("Invalid operation");
+        }
+        return operationHandlerMap.get(operation);
+    }
+}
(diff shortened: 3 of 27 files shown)


src/main/java/core/basesyntax/service/OperationStrategyImpl.java
```diff
@@ -0,0 +1,20 @@
...
+public class OperationStrategyImpl implements OperationStrategy {
+    private final Map<String, OperationHandler> operationHandlerMap;
+
+    public OperationStrategyImpl(Map<String, OperationHandler> operationHandlerMap) {
+        this.operationHandlerMap = operationHandlerMap;
+    }
+
+    @Override
+    public OperationHandler get(String operation) {
+        if (operationHandlerMap.get(operation) == null) {
```
mishamolnar, line 15: you can check this in DataValidator (there you have only 1 validation)

mishamolnar (approved):

src/main/java/core/basesyntax/model/Operation.java
```diff
@@ -0,0 +1,42 @@
+package core.basesyntax.model;
+
+public class Operation {
+    private final String type;
+    private final String fruit;
+    private final int amount;
+
+    public Operation(String type, String fruit, int amount) {
```
boroda4436, line 8:
```suggestion
    public Operation (Type type, String fruit, int amount) {
```
can we use type here?

src/main/java/core/basesyntax/service/AmountCalculator.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.Operation;
+import java.util.List;
+import java.util.Map;
+
+public interface AmountCalculator {
+    Map<String, Integer> calculate(List<Operation> operations);
```
boroda4436, line 8: Calculate is a very generic method name. Much more it returns some map, not long or int value. Can we rename this method?

src/main/java/core/basesyntax/service/StoreServiceImpl.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax.service;
+
+import java.util.Map;
+
+public class StoreServiceImpl implements StoreService {
```
boroda4436, line 5:
```suggestion
public class StoreServiceImpl implements ReportService {
```
Maybe report service will be better?

src/main/java/core/basesyntax/service/data/DataToOperationsFormatter.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service.data;
+
+import core.basesyntax.model.Operation;
+import java.util.List;
+
+public interface DataToOperationsFormatter {
+    List<Operation> formatData(List<String> data);
+}
```
boroda4436, line 8: Overall this is a good solution. In many libraries, it could be like
```suggestion
public interface DataParser<T, U> {
    List<T> parse(List<U> data);
}
```
and implementation could be like
```public class CsvOperationParser implements DataParser<Operation, String>```.
You don't need to refactor this code, this is mostly a hint for the future.

File: src/main/resources/output.csv [hunk 6]
```diff
@@ -0,0 +1,3 @@
+fruit,quantity
+banana,152
+apple,90
```
boroda4436, line 3: empty line?

boroda4436 (commented):
Overall looks good! Great job!

src/main/java/core/basesyntax/model/Operation.java
AlexRozv: changed Operation to save Type instead of String

src/main/java/core/basesyntax/service/AmountCalculator.java
AlexRozv: renamed to calculateDataForReport

src/main/java/core/basesyntax/service/StoreServiceImpl.java
AlexRozv: fixed

src/main/java/core/basesyntax/service/data/DataToOperationsFormatter.java
AlexRozv: refactored and renamed

src/main/resources/output.csv
AlexRozv: fixed

Max-Pochepets (approved):

PR #192, opened 2021-09-11
DmytroPetrushenko / jv-fruit-shop

DmytroPetrushenko:


src/main/java/dao/ReaderServiceImp.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import validator.Validator;
+import validator.ValidatorImp;
+
+public class ReaderServiceImp implements ReaderService {
+    private Validator validator = new ValidatorImp();
+
+    @Override
+    public List<String> readFromFile(String filePath) {
```
RomanDubovskyi, line 15: It's not obvious that reader service has inner validation. In this case if we would like to reuse this class in future this may produce some problems. Let's move validation to the upper level (e.g ReportService)

src/main/java/service/ReportServiceImp.java
```diff
@@ -0,0 +1,37 @@
...
+    private static final String MINUS_ACTIVITIES_CODES = "p";
+
+    @Override
+    public List<String> makeBalanceReport(List<String> activitiesList) {
+        int sum;
+        List<String> result = new ArrayList<>();
+        result.add("fruit,quantity");
+
+        Map<String, List<String[]>> stringListMap = activitiesList.stream()
+                .map(line -> line.split(","))
```
RomanDubovskyi, line 18: move ',' to constant SEPARATOR - this way we can reuse it in different classes and it will improve readability

src/main/java/service/ReportServiceImp.java
```diff
@@ -0,0 +1,37 @@
...
+            for (String[] activity:
+                    stringListMap.get(fruit)) {
+                sum += checkType(activity);
+            }
+            result.add(fruit + "," + sum);
+        }
+        return result;
+    }
+
+    private int checkType(String[] activity) {
```
RomanDubovskyi, line 33: this solution will be hard to scale if we will extend types of operation. Instead create handler for each operation and put it in the map where key is operation symbol and value is the object of corresponding implementation. Then you will just call `Handler currentHandler handlers.get(operationLetter);`
and apply it `currentHandler.apply(activity)`. Basically it's nice to use strategy pattern here

RomanDubovskyi (changes requested):

src/main/java/service/ReportServiceImp.java
```diff
@@ -0,0 +1,41 @@
...
+        List<String> result = new ArrayList<>();
+        result.add("fruit,quantity");
+
+        Map<String, List<String[]>> stringListMap = activitiesList.stream()
+                .map(line -> line.split(SEPARATOR))
+                .filter(validator::validationReportLine)
+                .collect(Collectors.groupingBy(strings -> strings[1]));
+
+        for (String fruit:
+                stringListMap.keySet()) {
```
RomanDubovskyi, line 31:
```suggestion
        for (String fruit: stringListMap.keySet()) {

```

src/main/java/service/ReportServiceImp.java
```diff
@@ -0,0 +1,41 @@
...
+            sum = 0;
+            for (String[] activity:
+                    stringListMap.get(fruit)) {
```
RomanDubovskyi, line 34:
```suggestion
            for (String[] activity: stringListMap.get(fruit)) {

```

RomanDubovskyi (approved):

PR #198, opened 2021-09-18
Fruit store hw

AlexPravda1:
- used Strategy pattern
- handler Map is done manually in main()
- checks against wrong initial data and negative fruit amount are implemented

### DIFF
--- a/src/main/java/dao/FileAccessDaoCsv.java
+++ b/src/main/java/dao/FileAccessDaoCsv.java
@@ -1,0 +1,9 @@
+package dao;
+
+import java.util.List;
+
+public interface FileAccessDaoCsv {
+    List<String> readFromFile(String filePath);
+
+    void writeToFile(String stringForReport, String outFilePath);
+}
--- a/src/main/java/dao/FileAccessDaoCsvImpl.java
+++ b/src/main/java/dao/FileAccessDaoCsvImpl.java
@@ -1,0 +1,47 @@
+package dao;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class FileAccessDaoCsvImpl implements FileAccessDaoCsv {
+
+    @Override
+    public List<String> readFromFile(String filePath) {
+        try {
+            return Files.readAllLines(Path.of(filePath));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file located at: " + filePath);
+        }
+    }
+
+    @Override
+    public void writeToFile(String stringForReport, String outFilePath) {
+        File toFile = new File(outFilePath);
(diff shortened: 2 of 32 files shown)


src/main/java/dao/FileAccessDaoCsv.java
```diff
@@ -0,0 +1,9 @@
+package dao;
+
+import java.util.List;
+
+public interface FileAccessDaoCsv {
```
RomanDubovskyi, line 5:
```suggestion
public interface CsvFileService {
```

src/main/java/dao/FileAccessDaoCsvImpl.java
```diff
@@ -0,0 +1,47 @@
...
+            return Files.readAllLines(Path.of(filePath));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file located at: " + filePath);
+        }
+    }
+
+    @Override
+    public void writeToFile(String stringForReport, String outFilePath) {
+        File toFile = new File(outFilePath);
+        ifExists(toFile);
```
RomanDubovskyi, line 25: can we just use file.exists?

src/main/java/service/ActivityStrategyImpl.java
```diff
@@ -0,0 +1,23 @@
...
+    private final Map<String, ActivityHandler> activityHandlerMap;
+
+    public ActivityStrategyImpl(Map<String, ActivityHandler> activityHandlerMap) {
+        this.activityHandlerMap = activityHandlerMap;
+    }
+
+    @Override
+    public ActivityHandler get(String activityType) {
+
+        if (activityType.equals("b") || activityType.equals("p")
```
RomanDubovskyi, line 16: that violates O from SOLID, instead use: `if (!map.containsKey(activityCode)) { throw an exception }`

RomanDubovskyi (changes requested):

src/main/java/dao/FileAccessDaoCsvImpl.java
AlexPravda1: thx, wasn't aware of the method until started "tests" HW. 
Removed this check for current task as non-relevant.

src/main/java/service/ActivityStrategyImpl.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    @Override
+    public ActivityHandler get(String activityType) {
+
+        if (activityHandlerMap.containsKey(activityType)) {
+            return activityHandlerMap.get(activityType);
+        } else {
+            throw new RuntimeException("Wrong activity type in the file");
+        }
```
RomanDubovskyi, line 20:
```suggestion
            return activityHandlerMap.get(activityType);
        } 
        throw new RuntimeException("Wrong activity type in the file");
```
we don't need else if we have return in the `if` block

RomanDubovskyi (approved):

PR #206, opened 2021-09-20
Hw 12 solution

7777vova7777:


src/main/java/service/impl/FileReaderServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import service.FileReaderService;
+
+public class FileReaderServiceImpl implements FileReaderService {
+    @Override
+    public List<String> readFile(String fileName) {
+        List<String> dataInput = new ArrayList<>();
```
RomanDubovskyi, line 13: there is ready to use method Files.readAllLines(), internally it uses BufferedReaer, so we can use it here

src/main/java/service/impl/FileWriterServiceImpl.java
```diff
@@ -0,0 +1,28 @@
+package service.impl;
+
+import db.Storage;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+import service.FileWriterService;
+
+public class FileWriterServiceImpl implements FileWriterService {
```
RomanDubovskyi, line 9: this class is responsible for 2 tasks: form report from storage  and write ready report to file. Let's leave here only writing so we can reuse this class if needed elsewhere

src/main/java/service/impl/InputDataValidatorImpl.java
```diff
@@ -0,0 +1,17 @@
+package service.impl;
+
+import java.util.List;
+import service.InputDataValidator;
+
+public class InputDataValidatorImpl implements InputDataValidator {
+    private static final String VALID_FORMAT = "[sprb],[a-zA-Z]+,[0-9]+";
```
RomanDubovskyi, line 7:
```suggestion
    private static final Pattern VALID_RECORD = Pattern.compile("[sprb],[a-zA-Z]+,[0-9]+");
```
This way it's gonna be better for the performance since constant creation of Pattern object inside of str.get(i).matches may be expensive

src/main/java/service/impl/InputDataValidatorImpl.java
```diff
@@ -0,0 +1,17 @@
...
+
+    @Override
+    public void chekDate(List<String> str) {
+        for (int i = 1; i < str.size(); i++) {
+            if (!str.get(i).matches(VALID_FORMAT)) {
```
RomanDubovskyi, line 12:
```suggestion
            if (!VALID_RECORD.matcher(str.get(i)).matches()) {
```

src/main/java/service/impl/OperationStrategyImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import service.OperationStrategy;
+import service.strategy.BalanceHandler;
+import service.strategy.Handler;
+import service.strategy.PurchaseHandler;
+import service.strategy.ReturnHandler;
+import service.strategy.SupplyHandler;
+
+public class OperationStrategyImpl implements OperationStrategy {
+    @Override
+    public Handler get(String[] str) {
```
RomanDubovskyi, line 12: instead it's better to create and initilize`Map<String, Handler>` and fill it with the Handler implementations not to create new object of Handler per each call.

RomanDubovskyi (changes requested):

src/main/java/service/OperationStrategy.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import service.strategy.BalanceHandler;
+import service.strategy.Handler;
+import service.strategy.PurchaseHandler;
+import service.strategy.ReturnHandler;
+import service.strategy.SupplyHandler;
+
+public interface OperationStrategy {
+    Map<String, Handler> map = new HashMap<>() {{
```
RomanDubovskyi, line 12: мапу краще залишити в класі і зробити приватною щоб напряму не було доступу. Таким чином в методі get за потреби можна буде додавати кастомну логіку

RomanDubovskyi (approved):

PR #221, opened 2021-09-21
Solid fruitshop solution

andrii-demchenko:


src/main/java/operation/AdditionHandler.java
```diff
@@ -0,0 +1,15 @@
+package operation;
+
+import model.Record;
+import report.Report;
+
+public class AdditionHandler implements OperationHandler {
```
pustoj0, line 6: Maybe better name it SupplyHandler, because the operation of addition is used twice, the same in other handlers, but it's not important.

pustoj0 (approved):
I'm not experienced enough to check it =(

src/main/java/service/ReportCreator.java
```diff
@@ -0,0 +1,7 @@
+package service;
+
+import java.util.List;
+
+public interface ReportCreator {
+    List<String> createNewReport();
```
Resci, line 6:
```suggestion
    List<String> createReport();
```

src/main/java/service/WriterService.java
```diff
@@ -0,0 +1,7 @@
+package service;
+
+import java.util.List;
+
+public interface WriterService {
+    void writeReportToCsv(List<String> report, String filename);
```
Resci, line 6:
```suggestion
    void write(List<String> report, String filename);
```

src/main/java/service/WriterService.java
```diff
@@ -0,0 +1,7 @@
+package service;
+
+import java.util.List;
+
+public interface WriterService {
```
Resci, line 5:
```suggestion
public interface FileWriter {
```

Resci (changes requested):

src/main/java/operation/AdditionHandler.java
andrii-demchenko: I think that addition is OK

src/main/java/validation/InputValidatorImpl.java
```diff
@@ -0,0 +1,38 @@
...
+                || !Character.isLetter(operationType.charAt(OPERATION_CHAR_INDEX))) {
+            throw new RuntimeException("Operation type has to be a single letter!");
+        }
+        if (fruit.isEmpty()) {
+            throw new RuntimeException("Fruit name can't be empty!");
+        }
+        if (fruit.matches(DIGITS_CHECK_REGEX)) {
+            throw new RuntimeException("Fruit name can't contain digits!");
+        }
+        int amount = Integer.parseInt(input[AMOUNT_INDEX]);
```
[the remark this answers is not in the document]
andrii-demchenko: It will throw NumberFormatException if you try to put letters in amount

Resci (approved):

PR #241, opened 2021-09-26
jv-fruit-shop solution

kryzhen:


src/main/java/dao/Reader.java
```diff
@@ -0,0 +1,7 @@
+package dao;
+
+import java.util.List;
+
+public interface Reader {
+    List<String> reader(String fileName);
```
RomanDubovskyi, line 6:
```suggestion
    List<String> read(String fileName);
```
Verbs are usually used for method naming

src/main/java/dao/ReaderImpl.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import validator.Validator;
+import validator.ValidatorImpl;
+
+public class ReaderImpl implements Reader {
+    private static final Validator validator = new ValidatorImpl();
+
+    @Override
+    public List<String> reader(String fileName) {
```
RomanDubovskyi, line 15: it's better to divide validation from reading, this way we will be able to reuse our reader whener we need to read from file in other parts of an application

src/main/java/dao/WriterImpl.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.IOException;
+import java.util.Map;
+import service.ReportImpl;
+
+public class WriterImpl implements Writer {
+    private static final String FILE_PATH = "src/main/reports/report.csv";
+    private static final Map<String, Integer> reportMap = new ReportImpl().report();
+
+    @Override
+    public void reportWriter() {
```
RomanDubovskyi, line 14:
```suggestion
    public void reportWriter(String fileName) {
```
it's better to dynamically pass fileName instead of hardcoding it

src/main/java/dao/WriterImpl.java
```diff
@@ -0,0 +1,30 @@
...
+        try {
+            BufferedWriter writer = new BufferedWriter(new FileWriter(FILE_PATH, true));
+            writer.write(header);
+            writer.newLine();
+            for (Map.Entry<String, Integer> entry : reportMap.entrySet()) {
+                writer.write(entry.getKey());
+                writer.write(", " + entry.getValue());
+                writer.newLine();
+            }
+            writer.close();
```
RomanDubovskyi, line 25: use try with resources instead. Now if we get exception in try block write.close() won't be called and we won't close connection to file

src/main/java/service/Report.java
```diff
@@ -0,0 +1,7 @@
+package service;
+
+import java.util.Map;
+
+public interface Report {
+    Map<String, Integer> report();
```
RomanDubovskyi, line 6: same here with method name

src/main/java/service/ReportImpl.java
```diff
@@ -0,0 +1,37 @@
...
+public class ReportImpl implements Report {
+
+    private static final String FILE_PATH = "src/main/resources/input.csv";
+
+    @Override
+    public Map<String, Integer> report() {
+        Reader reader = new ReaderImpl();
+        List<String> data = reader.reader(FILE_PATH);
+        Map<String, Integer> fruitBalance = new HashMap<>();
+        for (String element : data) {
```
RomanDubovskyi, line 18: this algorithm violites OCP from SOLID, use here strategy pattern instead: create separate handler for each operation and initialize Map<String, OperationHandler> with key as operation name and value as handler. You should have similar example in videos

RomanDubovskyi (changes requested):

src/main/java/dao/ReaderImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ReaderImpl implements Reader {
+    private static final String FILE_PATH = "src/main/resources/input.csv";
+
+    @Override
+    public List<String> read() {
```
RomanDubovskyi, line 13:
```suggestion
    public List<String> read(String fileName) {
```
Reader and writer should take path as an input, this way code can be reused

src/main/java/service/ReporterImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private OperationStrategy operationStrategy;
+
+    public ReporterImpl(Validator validator, Reader reader, OperationStrategy operationStrategy) {
+        this.validator = validator;
+        this.reader = reader;
+        this.operationStrategy = operationStrategy;
+    }
+
+    @Override
+    public Map<String, Integer> report() {
```
RomanDubovskyi, line 21:
```suggestion
    public Map<String, Integer> report(List<String> data) {
```
same here, let's improve flexibility

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #242, opened 2021-10-06
solution made and tested

ikhominitch:


src/main/java/service/activity/SupplyHandler.java
```diff
@@ -0,0 +1,12 @@
+package service.activity;
+
+import java.util.Map;
+import model.FruitRecord;
+
+public class SupplyHandler implements ActivityHandler {
```
RomanDubovskyi, line 6: Your supply and return handlers are the same, let's create one generic handler and use for both  return and supply operations

src/main/java/service/Validator.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.function.Function;
+import model.FruitRecord;
+
+public class Validator implements Function<String, FruitRecord> {
+    private static final int ACTIVITY_INDEX = 0;
+    private static final int FRUIT_INDEX = 1;
+    private static final int AMOUNT_INDEX = 2;
+
+    public FruitRecord apply(String rawRecord) {
```
RomanDubovskyi, line 11: to follow SRP from SOLID it's better for Validator to be responsible only for validating rawRecord. Let's move creating of FruitRecord to parser 
```suggestion
    public void apply(String rawRecord) {
```

RomanDubovskyi (changes requested):
great job with code structure, I like it, left 2 minor comments

src/main/java/core/basesyntax/Main.java
```diff
@@ -19,17 +19,17 @@
...
 
 public class Main {
     public static void main(String[] args) {
         Map<String, ActivityHandler> activityHandlersMap = new HashMap<>();
 
         activityHandlersMap.put("b", new BalanceHandler());
-        activityHandlersMap.put("s", new SupplyHandler());
         activityHandlersMap.put("p", new PurchaseHandler());
-        activityHandlersMap.put("r", new ReturnHandler());
+        SupplyHandler supplyHandler = new SupplyHandler();
```
RomanDubovskyi, line 30: we can use common naming IncreaseFruitHandler

RomanDubovskyi (approved):

PR #243, opened 2021-10-19
Home work

Noct2000:


src/main/java/core/basesyntax/service/ReportCreatorImpl.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax.service;
+
+import java.util.Map;
+
+public class ReportCreatorImpl implements ReportCreator {
+    private static final String COLUMN_NAMES = "fruit,quantity";
+    private static final String SEPARATOR = ",";
```
RomanDubovskyi, line 7: this constant is duplicated, make it public in the class that is related to it the most and reuse existing one in other classes

PR #245, opened 2021-11-09
Wrote an assignment jv-fruit-shop.

ASelivonchyk:
- made dao package with ResourcesDao, that response for reading and writing to resources, and StorageDao, that response for working with Storage;
- made db package, that contain fruit storage;
- made OperationStrategy, that response for working with different type op daile operations;
- in package Sevice made ReportService to work with report and StorageService that work with Storage;
- made Validator to validate input files;


MaliukDaria (approved):
Выглядит круто, молодец 🚀

PR #252, opened 2021-11-21
1st commit for fruit-shop

mykolaskrypka:


src/main/java/core/basesyntax/service/ReaderServiceImpl.java
```diff
@@ -0,0 +1,22 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ReaderServiceImpl implements ReaderService {
+
+    @Override
+    public List<String> readFromFile(String filePath) {
+        File file = new File(filePath);
+        try {
```
MaliukDaria, line 15: lets use try with resourses here(same in ReportServiceImpl)

src/main/java/core/basesyntax/model/Fruit.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax.model;
+
+public class Fruit {
+    private int count;
```
MaliukDaria, line 4: I dont think we should store fruit count in Fruit class. We can create some another model smth like TransactionDto, that wil have fruit, operation and quantity. We can parse input file in List<TransactionDto> and perform operations on it

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.db;
+
+import core.basesyntax.model.Fruit;
+import java.util.ArrayList;
+import java.util.List;
+
+public class Storage {
+    private static List<Fruit> fruitsStore = new ArrayList<>();
```
MaliukDaria, line 8: I think its better to store fruits in Map<Fruit, Integer> where we can store fruits and it's quantuty

MaliukDaria (changes requested):

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ReaderServiceImpl implements ReaderService {
+
+    @Override
+    public List<String> readFromFile(String filePath) {
+        try {
```
MaliukDaria, line 15: lets use try with resourses here(same in ReportServiceImpl) https://www.baeldung.com/java-try-with-resources

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+                .append(kv.getKey().getName())
+                .append(",")
+                .append(kv.getValue())
+                .append(System.lineSeparator()));
+        return stringBuilder.toString();
+    }
+
+    @Override
+    public void writeReport(String report, String fileName) {
+        try {
```
MaliukDaria, line 24: lets use try with resourses here https://www.baeldung.com/java-try-with-resources

MaliukDaria (changes requested):

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
mykolaskrypka: Дякую за посилання)

MaliukDaria (approved):

PR #296, opened 2022-01-17
Hw solid main

maxkmaxk:


src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,69 @@
...
+import core.basesyntax.strategy.TransactionHandler;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
 /**
  * Feel free to remove this class and create your own.
  */
 public class HelloWorld {
```
MaliukDaria, line 31: Lets rename this class to MainApp or smth like this

src/main/java/core/basesyntax/dao/FruitDao.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.dao;
+
+import core.basesyntax.model.Fruit;
+import java.util.Optional;
+
+public interface FruitDao {
+    public Optional<Fruit> get(String name);
```
MaliukDaria, line 7: we don't need to specify public access modifier here, all abstract interface methods are public by default (same in other places)

src/main/java/core/basesyntax/model/Transaction.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax.model;
+
+public class Transaction {
+    private final TransactionType type;
+    private final String fruitName;
+    private final int value;
```
MaliukDaria, line 6: not informative field name - `value`

src/main/java/core/basesyntax/service/TransactionService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.Transaction;
+
+public interface TransactionService {
+    void accept(Transaction transaction);
```
MaliukDaria, line 6:
```suggestion
    void apply(Transaction transaction);
```

src/main/java/core/basesyntax/db/FruitsData.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class FruitsData {
+    public static final Map<String, Integer> fruitMap = new HashMap<>();
```
MaliukDaria, line 7: Maybe it's better to Store Fruits here, as your Fruit model has name and quantity?

src/main/java/core/basesyntax/model/Fruit.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax.model;
+
+public class Fruit {
+    private final String name;
+    private int amount;
+
+    public Fruit(String name) {
+        this(name, 0);
+    }
```
MaliukDaria, line 9:
```suggestion
    public Fruit(String name) {
        this.name = name;
    }
```

src/main/java/core/basesyntax/service/impl/FruitsConsumerCsv.java
```diff
@@ -0,0 +1,28 @@
...
+    private final String filePath;
+
+    public FruitsConsumerCsv(String filePath) {
+        this.filePath = filePath;
+    }
+
+    @Override
+    public void accept(List<Fruit> fruits) {
+        FileUtils.writeFile(fruitsToCsvRows(fruits), filePath);
+    }
```
MaliukDaria, line 19: You have `FruitsConsumer` class with the `accept()` method, that writes data to file. Let's somehow rename the class and method, because it is not obvious from their names what exactly they do.

src/main/java/core/basesyntax/service/impl/FruitsConsumerCsv.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private List<String> fruitsToCsvRows(List<Fruit> fruits) {
```
MaliukDaria, line 21: Methods names should starts from verbs

src/main/java/core/basesyntax/service/impl/TransactionServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private final Map<TransactionType, TransactionHandler> strategyMap;
+
+    public TransactionServiceImpl(FruitDao fruitDao,
+                                  Map<TransactionType, TransactionHandler> strategyMap) {
+        this.fruitDao = fruitDao;
+        this.strategyMap = strategyMap;
+    }
+
+    @Override
+    public void accept(Transaction transaction) {
```
MaliukDaria, line 22:
```suggestion
    public void apply(Transaction transaction) {
```

src/main/java/core/basesyntax/service/impl/TransactionsSupplierCsv.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class TransactionsSupplierCsv implements TransactionsSupplier {
+    private final String filePath;
+
+    public TransactionsSupplierCsv(String filePath) {
+        this.filePath = filePath;
+    }
+
+    @Override
+    public List<Transaction> get() {
```
MaliukDaria, line 17: Let's pass filePath as incoming parameter. `get()` it's not informative method name(this method actually reads the data from the file and parse it to some Transaction)

MaliukDaria (changes requested):

src/main/java/core/basesyntax/db/FruitsData.java
maxkmaxk: Yes, and it's already like that

src/main/java/core/basesyntax/service/impl/FruitsConsumerCsv.java
maxkmaxk: It's implementation of interface method, this one implementation accept fruits to write to csv, but we can ...use another implementation with different destination (write to DB, send by email etc.). About class name,  - supposed it's fruits consumer which write to CSV.

src/main/java/core/basesyntax/service/impl/TransactionsSupplierCsv.java
maxkmaxk: it's interface method, interface don't know from where we take transactions

src/main/java/core/basesyntax/db/FruitsData.java
maxkmaxk: FruitDaoImpl
    public void put(Fruit fruit) {
        FruitsData.fruitMap.put(fruit.getName(), fruit.getAmount());
    }

Sorry, i don't know how to quote code here

src/main/java/core/basesyntax/service/impl/TransactionsSupplierCsv.java
maxkmaxk: It's implementation of TransactionSupplier interface (data could be taken from different sources)

src/main/java/core/basesyntax/db/FruitsData.java
MaliukDaria, line 7:
```suggestion
    public static final List<Fruit> fruits = new HashMap<>();
```

Я имею ввиду так

src/main/java/core/basesyntax/service/impl/TransactionsSupplierCsv.java
MaliukDaria: Не важно это метод интерфейса или нет, лучше пусть метод принимает на вход путь к файлу(для этого нужно переписать метод самого интерфейса). У тебя не очень информативное название для метода get(), без контекста не ясно что он делает. Давай переименуем его на getTransactionsFromFile или что-то в этом духе

MaliukDaria (changes requested):

MaliukDaria (approved):
Good job👍🏻

PR #303, opened 2022-02-04
first test version (dont check)

LinkClink:


src/main/java/fruitshop/Application.java
```diff
@@ -0,0 +1,51 @@
...
+public class Application {
+    private static final WriteDataToStorageService writeDataService = new WriteDataToStorageImpl();
+    private static final FileReadServiceImpl fileReadService
+            = new FileReadServiceImpl(writeDataService);
+    private static final FileWriteServiceImpl fileWriteService = new FileWriteServiceImpl();
+    private static final ProcessDataImpl processData = new ProcessDataImpl();
+    private static final GenerateReportDataImpl generateReportData = new GenerateReportDataImpl();
+    private static final TransactionStorage transactionStorage = new TransactionStorage();
+    private static final ReportStorage reportStorage = new ReportStorage();
+    private final Map<Operation, StrategyService> operationsMap = new HashMap<>();
```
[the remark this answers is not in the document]
Sasha-Bachynskyi: let mentor resolve it

input.csv
```diff
@@ -0,0 +1,9 @@
+type,fruit,quantity
+b,banana,20
+b,apple,100
+s,banana,100
+p,banana,13
+r,apple,10
+p,apple,20
+p,banana,5
+s,banana,50
```
Sasha-Bachynskyi, line 9: you should store your files in folder "resources"

Sasha-Bachynskyi (approved):
looks good to me

input.csv
LinkClink: ok

src/main/java/fruitshop/Application.java
```diff
@@ -0,0 +1,51 @@
...
+import fruitshop.strategy.SubtractionHandler;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Application {
+    private static final WriteDataToStorageService writeDataService = new WriteDataToStorageImpl();
+    private static final FileReadServiceImpl fileReadService
+            = new FileReadServiceImpl(writeDataService);
+    private static final FileWriteServiceImpl fileWriteService = new FileWriteServiceImpl();
+    private static final ProcessDataImpl processData = new ProcessDataImpl();
```
AlexRozv, line 24:
```suggestion
    private static final FileWriteService fileWriteService = new FileWriteServiceImpl();
    private static final ProcessData processData = new ProcessDataImpl();
```

let's initialize our services using interfaces, do the same everywhere

src/main/java/fruitshop/service/data/ProcessDataService.java
```diff
@@ -0,0 +1,11 @@
...
+
+import fruitshop.model.FruitTransaction;
+import fruitshop.model.Operation;
+import fruitshop.strategy.StrategyService;
+import java.util.List;
+import java.util.Map;
+
+public interface ProcessDataService {
+    void process(List<FruitTransaction> list, Map<Operation, StrategyService> operationsMap);
+}
```
AlexRozv, line 11: instead of passing operationMap to this method, it is better to create OperationStrategy class that will just contain this map that should be passed using constructor and method get() that will return specific StrategyService (it's better to rename it to OperationHandler)

when you do this you don't have to pass this map here and you can just create an instance of a strategy inside the method and get the needed StrategyService to process the transaction

AlexRozv (changes requested):

src/main/java/fruitshop/service/data/ProcessDataService.java
LinkClink: understood

AlexRozv (approved):
well done ✅

PR #304, opened 2022-02-07
new fruitShop was created, I`m waiting for your comments

PavloHeorh:


src/main/java/core/basesyntax/service/strategy/SubtractOperationHandler.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax.service.strategy;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.DailyActivity;
+import core.basesyntax.model.Fruit;
+
+public class SubtractOperationHandler implements OperationHandler {
+    @Override
+    public boolean operation(DailyActivity dailyActivity) {
```
RomanDubovskyi, line 9: if you always return true by default, you can make the method return void to return false instead of throwing an exception

src/main/java/core/basesyntax/service/impl/DataWriterImpl.java
```diff
@@ -0,0 +1,20 @@
...
+import core.basesyntax.service.DataWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class DataWriterImpl implements DataWriter {
+
+    @Override
+    public boolean writeDataToFile(List<String> list, String filename) {
```
RomanDubovskyi, line 12: same here with the boolean return type

src/main/java/core/basesyntax/service/impl/DataWriterImpl.java
```diff
@@ -0,0 +1,20 @@
...
+        try {
+            Files.write(Path.of(filename), list);
+            return true;
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t write data to file");
```
RomanDubovskyi, line 17:
```suggestion
            throw new RuntimeException("Can`t write data to file " + fileName);
```
to make a message more informative I would recommend to add some kind of identifier of where the problem occurred

src/main/java/core/basesyntax/service/impl/DataWriterImpl.java
RomanDubovskyi, line 12:
```suggestion
    public boolean writeDataToFile(List<String> list, String fileName) {
```
use camelCase for naming

RomanDubovskyi (changes requested):
GJ, left minor comments

src/main/java/core/basesyntax/service/strategy/SubtractOperationHandler.java
PavloHeorh: but, l thought it`s much better to create methods with return type(at least boolean) to avoid troubles during testing of this method

Resci (approved):

PR #307, opened 2022-02-09
created FruitShop which make balance report

liliia-ponomarenko:


src/main/java/core/basesyntax/MainFruitStorage.java
```diff
@@ -0,0 +1,44 @@
...
+    private static final String INPUT_FILE_PATH = "src/main/resources/input_report.csv";
+    private static final String REPORT_FILE_NAME = "src/main/resources/output_report.csv";
+
+    public static void main(String[] args) {
+        Map<FruitTransaction.Operation, OperationService> operationServiceMap = new HashMap<>();
+        operationServiceMap.put(FruitTransaction.Operation.BALANCE, new BalanceOperationService());
+        operationServiceMap.put(FruitTransaction.Operation.SUPPLY, new SupplyOperationService());
+        operationServiceMap.put(FruitTransaction.Operation.PURCHASE,
+                new PurchaseOperationService());
+        operationServiceMap.put(FruitTransaction.Operation.RETURN, new ReturnOperationService());
```
sviat-h, line 32: I think you should to extract map initialization from main method to improve readability

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.db;
+
+import core.basesyntax.model.Fruit;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    public static final Map<Fruit, Integer> fruitBalance = new HashMap<>();
```
sviat-h, line 8: I think this field should be private

src/main/java/core/basesyntax/service/DataProcessingService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface DataProcessingService {
+    void processingDate(List<FruitTransaction> date);
```
sviat-h, line 7:
```suggestion
    void processingData(List<FruitTransaction> data);
```

src/main/java/core/basesyntax/service/impl/DataProcessingServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class DataProcessingServiceImpl implements DataProcessingService {
+    private final OperationStrategy operationStrategy;
+
+    public DataProcessingServiceImpl(OperationStrategy operationStrategy) {
+        this.operationStrategy = operationStrategy;
+    }
+
+    @Override
+    public void processingDate(List<FruitTransaction> date) {
```
sviat-h, line 18:
```suggestion
    public void processingData(List<FruitTransaction> data) {
```

sviat-h (approved):

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,65 @@
+package core.basesyntax.model;
+
+public class FruitTransaction {
+    private static final String BALANCE_MARKING = "b";
+    private static final String SUPPLY_MARKING = "s";
+    private static final String PURCHASE_MARKING = "p";
+    private static final String RETURN_MARKING = "r";
```
AlexRozv, line 7: we don't need it here, see the next comment to understand how to implement it

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,65 @@
...
+                return Operation.BALANCE;
+            case SUPPLY_MARKING :
+                return Operation.SUPPLY;
+            case RETURN_MARKING :
+                return Operation.RETURN;
+            default:
+                return Operation.PURCHASE;
+        }
+    }
+
```
AlexRozv, line 28: let's move this method inside the enum and let's not use switch case there 

look at the method fromString on screenshot
Screenshot 2022-02-18 at 10 48 23

src/main/java/core/basesyntax/strategy/OperationService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.model.FruitTransaction;
+
+public interface OperationService {
+    void processOperation(FruitTransaction fruitTransaction);
+}
```
AlexRozv, line 7: it's better to rename it to OperationHandler and do the same for its implementations, this way will keep our naming consistent and stick to the strategy pattern

AlexRozv (changes requested):

src/main/java/core/basesyntax/model/FruitTransaction.java
liliia-ponomarenko: Thanks, I couldn't  find a good way to realize this method )

AlexRozv (approved):
good job 👍

PR #314, opened 2022-02-17
Interfaces, classes and methods for making fruit store shop's work re…

Liubov-Chernohor:
…port were implemented

### DIFF
--- a/result.csv
+++ b/result.csv
@@ -1,0 +1,3 @@
+fruit,quantity
+banana,152
+apple,70
--- a/src/main/java/core/basesyntax/dao/FruitShopDaoImpl.java
+++ b/src/main/java/core/basesyntax/dao/FruitShopDaoImpl.java
@@ -1,0 +1,44 @@
+package core.basesyntax.dao;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.Operation;
+import core.basesyntax.service.OperationStrategy;
+import core.basesyntax.service.operations.OperationService;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+public class FruitShopDaoImpl implements FruitShopDao {
+    private OperationService operationService;
+
+    @Override
+    public void updateStorage(String data) {
+        List<String> dataToList = Arrays.asList(data.split(System.lineSeparator()));
+        dataToList.stream()
+                .map(s -> s.split(","))
+                .forEach(arr -> {
+                    operationService = OperationStrategy
+                            .getOperationType(getOperationValue(arr[0]));
+                    operationService.operate(arr[1], Integer.parseInt(arr[2]));
+                });
+    }
+
+    @Override
+    public String getDataFromStorage() {
+        StringBuilder dataFromStorage = new StringBuilder();
+        for (Map.Entry<String, Integer> entry : Storage.fruits.entrySet()) {
+            dataFromStorage.append(entry.getKey())
+                    .append(",")
+                    .append(entry.getValue())
+                    .append(System.lineSeparator());
+        }
+        return dataFromStorage.toString();
+    }
+
+    private Operation getOperationValue(String str) {
+        return Arrays.stream(Operation.values())
+                .filter(operation -> operation.getOperation().equals(str))
+                .findFirst()
+                .get();
+    }
+}
--- a/src/main/java/core/basesyntax/service/OperationStrategy.java
+++ b/src/main/java/core/basesyntax/service/OperationStrategy.java
@@ -1,0 +1,22 @@
+package core.basesyntax.service;
+
+import core.basesyntax.service.operations.Expense;
+import core.basesyntax.service.operations.Income;
+import core.basesyntax.service.operations.OperationService;
+import java.util.HashMap;
+import java.util.Map;
+
+public class OperationStrategy {
+    private static Map<Operation, OperationService> events = new HashMap<>(4);
+
+    static {
+        events.put(Operation.RETURN, new Expense());
+        events.put(Operation.BALANCE, new Income());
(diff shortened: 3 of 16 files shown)


result.csv
```diff
@@ -0,0 +1,3 @@
+fruit,quantity
+banana,152
+apple,70
```
LinkClink, line 3: Incorrect

src/main/java/core/basesyntax/service/OperationStrategy.java
```diff
@@ -0,0 +1,22 @@
...
+public class OperationStrategy {
+    private static Map<Operation, OperationService> events = new HashMap<>(4);
+
+    static {
+        events.put(Operation.RETURN, new Expense());
+        events.put(Operation.BALANCE, new Income());
+        events.put(Operation.PURCHASE, new Expense());
+        events.put(Operation.SUPPLY, new Income());
+    }
+
```
LinkClink, line 18: (only recommendation)
 I think you should initialisation this map with operation in main class

src/main/java/core/basesyntax/service/OperationStrategy.java
```diff
@@ -0,0 +1,22 @@
...
+import core.basesyntax.service.operations.Income;
+import core.basesyntax.service.operations.OperationService;
+import java.util.HashMap;
+import java.util.Map;
+
+public class OperationStrategy {
+    private static Map<Operation, OperationService> events = new HashMap<>(4);
+
+    static {
+        events.put(Operation.RETURN, new Expense());
```
LinkClink, line 13: r (RETURN) it is addition operation, you should paste here **Income()** strategy.

r - return, means someone who have bought the fruits now returns them back

src/main/java/core/basesyntax/service/OperationStrategy.java
Liubov-Chernohor: I think filling this Map shouldn't depend on the user. The keys are predefined constants, their values also couldn't change, that's why, I think, this Map should be initialized at class load time.

LinkClink (approved):
Good

src/main/java/core/basesyntax/dao/FruitShopDaoImpl.java
```diff
@@ -0,0 +1,44 @@
...
+                    .append(System.lineSeparator());
+        }
+        return dataFromStorage.toString();
+    }
+
+    private Operation getOperationValue(String str) {
+        return Arrays.stream(Operation.values())
+                .filter(operation -> operation.getOperation().equals(str))
+                .findFirst()
+                .get();
```
AlexRozv, line 42: can we use HashMap mathod get here instead of using stream?
do the same in other methods where possible

src/main/java/core/basesyntax/dao/FruitShopDaoImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Map;
+
+public class FruitShopDaoImpl implements FruitShopDao {
+    private OperationService operationService;
+
+    @Override
+    public void updateStorage(String data) {
+        List<String> dataToList = Arrays.asList(data.split(System.lineSeparator()));
+        dataToList.stream()
+                .map(s -> s.split(","))
```
AlexRozv, line 18: it's better to seprate logic that parses our data because our format of incoming data file may change and will have to change this logic here, but this class should be only responsible for working with storage, fix it in other methods as well

src/main/java/core/basesyntax/dao/FruitShopDaoImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.Operation;
+import core.basesyntax.service.OperationStrategy;
+import core.basesyntax.service.operations.OperationService;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+public class FruitShopDaoImpl implements FruitShopDao {
```
AlexRozv, line 11: let's rename it to FruitDao because we're representing storage of fruits here

src/main/java/core/basesyntax/service/operations/Expense.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.service.operations;
+
+import core.basesyntax.db.Storage;
+
+public class Expense implements OperationService {
```
AlexRozv, line 5: let's rename OperationService to OperationHandler and Expense to ExpenseOperationHandler to keep our naming consistent and stick to the strategy pattern
do the same for other implementations

src/main/java/core/basesyntax/service/report/DailyReport.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax.service.report;
+
+import core.basesyntax.service.file.FIleWriterService;
+import java.io.File;
+
+public class DailyReport implements ReportMaker {
```
AlexRozv, line 6: DailyReport -> ReportMakerImpl
consistant naming

AlexRozv (changes requested):

src/main/java/core/basesyntax/dao/FruitShopDaoImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        }
+        return dataFromStorage.toString();
+    }
+
+    private Operation getOperationValue(String str) {
+        return Arrays.stream(Operation.values())
+                .filter(operation -> operation.getOperation().equals(str))
+                .findFirst()
+                .get();
+    }
```
AlexRozv, line 43: let's move this method inside our enum class

AlexRozv (approved):
good job 👌

PR #319, opened 2022-02-18
Fruit shop storage was implemented

nastyaprokhorenko:


src/main/java/fruitshop/Main.java
```diff
@@ -0,0 +1,23 @@
...
+import fruitshop.service.impl.OperationStrategyImpl;
+import fruitshop.service.impl.ReaderServiceImpl;
+import fruitshop.service.impl.WriterServiceImpl;
+
+public class Main {
+    public static void main(String[] args) {
+        ReaderService readerService = new ReaderServiceImpl();
+        OperationStrategy operationStrategy = new OperationStrategyImpl();
+        WriterService writerService = new WriterServiceImpl();
+        String inputFile = "src/main/resources/input.txt";
```
yurii-tomashevskyi, line 17: I think better use constants

PR #325, opened 2022-02-20
Realised FruitShop task

DenTima:


src/main/java/services/impl/FruitDaoServiceImp.java
```diff
@@ -0,0 +1,46 @@
...
+public class FruitDaoServiceImp implements FruitDaoService {
+    private Storage storage = new Storage();
+
+    public FruitDaoServiceImp(Storage storage) {
+        this.storage = storage;
+    }
+
+    @Override
+    public void put(FruitRecord fruitRecord) {
+        Set<Fruit> fruitsInStorage = storage.getFruitsInStorage();
```
karma-o, line 19: Also you may not understand that `getFruitsInStorage()` returns a link of the object and you can modify that object without creating a local variable of it, that is an advise for the future :)

src/main/java/services/impl/ParseServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import model.Fruit;
+import model.FruitRecord;
+import services.ParseService;
+
+public class ParseServiceImpl implements ParseService {
+    private static final String RECORD_SEPARATOR = "\r\n";
```
karma-o, line 12: remember about `System.lineSeparator()`

src/main/java/services/impl/ParseServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final String DATA_SEPARATOR = ",";
+    private static final int APPEND_TO_REMOVE_TITLE = 2;
+
+    @Override
+    public List<FruitRecord> parseFromCsv(String datInString) {
+        List<FruitRecord> listOfRecords;
+        datInString = datInString.substring(datInString.indexOf(RECORD_SEPARATOR)
```
karma-o, line 19: typo
```suggestion
        dataInString = datInString.substring(datInString.indexOf(RECORD_SEPARATOR)
```

src/main/java/services/impl/ParseServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+                + APPEND_TO_REMOVE_TITLE);
+        String[] arrOfRecords = datInString.split(RECORD_SEPARATOR);
+        listOfRecords = Arrays.stream(arrOfRecords)
+                .map(String::trim)
+                .map(this::parseIntoFruitRecord)
+                .collect(Collectors.toList());
+        return listOfRecords;
```
karma-o, line 26:
```suggestion
        return Arrays.stream(arrOfRecords)
                .map(String::trim)
                .map(this::parseIntoFruitRecord)
                .collect(Collectors.toList());
```

src/main/java/services/impl/ParseServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    private FruitRecord parseIntoFruitRecord(String recordInString) {
+        String[] dataInArray = recordInString.split(DATA_SEPARATOR);
+        char typeOfOperation = dataInArray[0].charAt(0);
+        String nameOfFruit = dataInArray[1];
+        int amountOfFruits = Integer.parseInt(dataInArray[2]);
+        return new FruitRecord(typeOfOperation, nameOfFruit, amountOfFruits);
+    }
+
+    @Override
+    public String parseIntoCsv(Set<Fruit> datInString) {
```
karma-o, line 38: this method just returns a String, this has nothing to do with a csv files
also bad naming of a parameter
```suggestion
    public String toString(Set<Fruit> fruitSet) {
```

karma-o (changes requested):

src/main/java/model/FruitRecord.java
```diff
@@ -0,0 +1,37 @@
+package model;
+
+public class FruitRecord {
```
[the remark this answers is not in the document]
DenTima: so of course we may not use this class, but in the hints it was suggested to use this class for the convenience of data processing 
image

src/main/java/model/Operator.java
```diff
@@ -0,0 +1,26 @@
+package model;
```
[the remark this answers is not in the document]
DenTima: Ok, agree

src/main/java/model/Operator.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import services.FruitDaoService;
+import services.OperationsHandler;
+
+public class Operator {
```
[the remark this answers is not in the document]
DenTima: agree

src/main/java/services/impl/FruitDaoServiceImp.java
```diff
@@ -0,0 +1,46 @@
...
+        for (Fruit fruitInStorage : fruitsInStorage) {
+            if (fruitInStorage.getName().equals(fruitRecord.getNameOfFruit())) {
+                fruitInStorage.setAmount(fruitInStorage.getAmount() + fruitRecord.getAmount());
+                break;
+            }
+        }
+    }
+
+    @Override
+    public void save(FruitRecord fruitRecord) {
```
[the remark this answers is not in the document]
DenTima: method save is used only for Balance Operation. When calling this operation, we do not need to check whether this record already exists. We create a new record every time. When calling the put method, we will always do an extra check for this operation. I think separate methods will optimize this code

src/main/java/services/impl/FruitDaoServiceImp.java
```diff
@@ -0,0 +1,46 @@
...
+        fruitsInStorage.add(new Fruit(fruitRecord.getNameOfFruit(), fruitRecord.getAmount()));
+    }
+
+    @Override
+    public Set<Fruit> get() {
+        return storage.getFruitsInStorage();
+    }
+
+    @Override
+    public Set<Fruit> getAll() {
```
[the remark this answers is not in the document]
DenTima: agree

src/main/java/services/impl/ParseServiceImpl.java
DenTima: agree we can change it

src/main/java/services/impl/ParseServiceImpl.java
DenTima: agree, changed)

src/main/java/services/impl/ParseServiceImpl.java
DenTima: agree

src/main/java/services/impl/ParseServiceImpl.java
DenTima: agree. Refactored method to parceToString()

karma-o (approved):
Looks good to me 🔥

src/main/java/services/impl/ParseServiceImpl.java
```diff
@@ -9,21 +9,19 @@
...
-    private static final String RECORD_SEPARATOR = "\r\n";
     private static final String DATA_SEPARATOR = ",";
     private static final int APPEND_TO_REMOVE_TITLE = 2;
 
     @Override
-    public List<FruitRecord> parseFromCsv(String datInString) {
+    public List<FruitRecord> parseFromCsv(String dataInString) {
         List<FruitRecord> listOfRecords;
-        datInString = datInString.substring(datInString.indexOf(RECORD_SEPARATOR)
+        dataInString = dataInString.substring(dataInString.indexOf(System.lineSeparator())
```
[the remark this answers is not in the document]
DenTima: Yes, it is successfully compiled

KostyaNikiforov (approved):

PR #336, opened 2022-03-12
Fruit storage project was created

liubcheck:


src/main/java/core/basesyntax/service/operation/SupplyOperation.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax.service.operation;
+
+import core.basesyntax.db.FruitStorage;
+import core.basesyntax.exceptions.NoSuchFruitException;
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.OperationType;
+
+public class SupplyOperation implements Operation {
```
KseniiaMakarova, line 8:
```suggestion
public class SupplyOperationHandler implements OperationHandler {
```

src/main/java/core/basesyntax/service/operation/Operation.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service.operation;
+
+import core.basesyntax.model.Fruit;
+
+public interface Operation {
```
KseniiaMakarova, line 5:
```suggestion
public interface OperationHandler {
```

src/main/java/core/basesyntax/service/impl/OperationHandler.java
```diff
@@ -0,0 +1,34 @@
...
+import core.basesyntax.model.OperationType;
+import core.basesyntax.service.operation.BalanceOperation;
+import core.basesyntax.service.operation.Operation;
+import core.basesyntax.service.operation.PurchaseOperation;
+import core.basesyntax.service.operation.ReturnOperation;
+import core.basesyntax.service.operation.SupplyOperation;
+import java.util.HashMap;
+import java.util.Map;
+
+public class OperationHandler {
```
KseniiaMakarova, line 12:
```suggestion
public class OperationHandlerProvider {
```

src/main/java/core/basesyntax/service/impl/OperationHandler.java
```diff
@@ -0,0 +1,34 @@
...
+    private Map<String, Operation> operationMap;
+
+    public OperationHandler() {
+        operationMap = new HashMap<>();
+        putOperations();
+    }
+
+    public Map<String, Operation> getOperations() {
```
KseniiaMakarova, line 20:
```suggestion
    public Map<String, Operation> getOperationHandlers() {
```

src/main/java/core/basesyntax/strategy/OperationStrategy.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.OperationType;
+
+public interface OperationStrategy {
+    void startOperation(OperationType operationType, Fruit fruit);
```
KseniiaMakarova, line 7:
```suggestion
    void process(OperationType operationType, Fruit fruit);
```

src/main/java/core/basesyntax/service/impl/FruitTransaction.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public void doTransaction() {
+        operationStrategy.startOperation(operationType, fruit);
+    }
+
+    public Fruit getFruit() {
+        return fruit;
+    }
+
+    private void defineOperationType(String operationLetter) {
```
KseniiaMakarova, line 28: why do we have field `operation` in enum OperationType? let's iterate over enum values instead of switch case

src/main/java/core/basesyntax/service/impl/FruitTransaction.java
```diff
@@ -0,0 +1,46 @@
...
+
+import core.basesyntax.exceptions.IncorrectOperationException;
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.OperationType;
+import core.basesyntax.strategy.OperationStrategy;
+
+public class FruitTransaction {
+    private OperationType operationType;
+    private final Fruit fruit;
+    private final OperationStrategy operationStrategy;
```
KseniiaMakarova, line 11: let's remove `operationStrategy` field and call its method for each transaction in another place
let this class have only state

src/main/java/core/basesyntax/dao/FruitFileWriter.java
```diff
@@ -0,0 +1,39 @@
...
+            bufferedWriter.write(createOutput());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write in the file " + fileName, e);
+        }
+    }
+
+    private String createOutput() {
+        return fruits.stream()
+                .map(Fruit::toString)
+                .collect(Collectors.joining(System.lineSeparator()));
```
KseniiaMakarova, line 37: let's move this logic into a separate service ReportService

src/main/java/core/basesyntax/dao/FruitFileWriter.java
```diff
@@ -0,0 +1,39 @@
...
+        if (fileName == null) {
+            throw new NullFileNameException("Can't write in a file with NULL name");
+        }
+        this.fileName = fileName;
+    }
+
+    @Override
+    public void write() {
+        try (BufferedWriter bufferedWriter = Files.newBufferedWriter(Paths.get(fileName))) {
+            bufferedWriter.write("fruit,quantity" + System.lineSeparator());
```
KseniiaMakarova, line 27: same for this header line - this is part of the report

writer should only write to file - it is its single responsibility

src/main/java/core/basesyntax/dao/FruitFileReader.java
```diff
@@ -0,0 +1,51 @@
...
+    public FruitFileReader(String fileName) {
+        if (fileName == null) {
+            throw new NullFileNameException("Can't get a file with NULL name");
+        }
+        this.fileName = fileName;
+        operationStrategy = new OperationStrategyImpl(new OperationHandler().getOperations());
+    }
+
+    @Override
+    public void read() {
```
KseniiaMakarova, line 29: same here
reader should better only read
can we return List of Strings from here? and do all further processing in a different service TransactionService?

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+public class Main {
+    private static final String READ_FILE_NAME = "test.cvs";
+    private static final String WRITE_FILE_NAME = "result.cvs";
+
+    public static void main(String[] args) {
+        FruitDataReader fruitReader = new FruitFileReader(READ_FILE_NAME);
+        fruitReader.read();
+        FruitDataWriter fruitWriter = new FruitFileWriter(FruitStorage.fruits, WRITE_FILE_NAME);
+        fruitWriter.write();
+        System.out.println("Here is one more successfull day in our fruit storage!");
```
KseniiaMakarova, line 18: I would suggest to call here services one after another:

reader - read lines

transactionService - process transactions based on strategy (or better ParserService (string -> FruitTransaction) and then TransactionService (accepts FruitTransaction and does manipulation with Storage)

reportService - returns String as report

writer

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,36 @@
...
+        for (String activity : activityList) {
+            FruitTransaction fruitTransaction = parserService.parseToTransaction(activity);
+            TransactionService transactionService = new TransactionServiceImpl(fruitTransaction);
+            transactionService.doOperation();
+        }
+        ReportService reportService = new ReportServiceImpl();
+        String storageOutput = reportService.createOutput(FruitStorage.fruits);
+        FruitDataWriter fruitWriter = new FruitFileWriter(storageOutput, WRITE_FILE_NAME);
+        fruitWriter.write();
+        System.out.println("Here is one more successful day in our fruit storage!");
```
KseniiaMakarova, line 34: very good

src/main/java/core/basesyntax/dao/FruitDataReader.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.dao;
+
+import java.util.List;
+
+public interface FruitDataReader {
+    List<String> readTransactions();
+}
```
KseniiaMakarova, line 7: here I would suggest using more abstract class and method name
in fact, you will read Strings, no matter if it is info about fruits, or something else

src/main/java/core/basesyntax/dao/FruitDataWriter.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.dao;
+
+public interface FruitDataWriter {
```
KseniiaMakarova, line 3: here as well

src/main/java/core/basesyntax/dao/FruitFileWriter.java
```diff
@@ -0,0 +1,30 @@
...
+
+import core.basesyntax.exceptions.NullFileNameException;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public class FruitFileWriter implements FruitDataWriter {
+    private final String storageOutput;
+    private final String fileName;
```
KseniiaMakarova, line 11: it is better for services not to have state (fields)
it is more flexible to pass this information to method parameters

src/main/java/core/basesyntax/db/FruitStorage.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class FruitStorage {
+    public static final List<Fruit> fruits = new ArrayList<>();
+
+    // Check if the storage already contains such type of fruits
+    // without mentioning their quantity, cause it doesn't matter
+    // That's why boolean equals() doesn't compare quantity for fruits
+    // and why usual boolean contains() doesn't pass in this case
```
KseniiaMakarova, line 13: contains() uses equals() under the hood
so if your equals() doesn't take into account quantity, then it should work

src/main/java/core/basesyntax/model/Fruit.java
```diff
@@ -0,0 +1,65 @@
...
+        if (object == null || getClass() != object.getClass()) {
+            return false;
+        }
+        Fruit fruit = (Fruit) object;
+        return name.equals(fruit.getName());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, quantity);
```
KseniiaMakarova, line 59: equals() and hashCode() should work with same fields
so let's remove quantity from here

src/main/java/core/basesyntax/service/impl/OperationHandlerProvider.java
```diff
@@ -0,0 +1,33 @@
...
+import core.basesyntax.model.OperationType;
+import core.basesyntax.service.operation.BalanceOperationHandler;
+import core.basesyntax.service.operation.PurchaseOperationHandler;
+import core.basesyntax.service.operation.ReturnOperationHandler;
+import core.basesyntax.service.operation.SupplyOperationHandler;
+import java.util.HashMap;
+import java.util.Map;
+
+public class OperationHandlerProvider {
+    private Map<String, core.basesyntax.service.operation.OperationHandler> operationMap;
```
KseniiaMakarova, line 12: let's use import

src/main/java/core/basesyntax/service/impl/OperationHandlerProvider.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax.service.impl;
```
KseniiaMakarova, line 1:
```suggestion
package core.basesyntax.service;
```

src/main/java/core/basesyntax/service/impl/FruitTransaction.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.service.impl;
```
KseniiaMakarova, line 1:
```suggestion
package core.basesyntax.model;
```

src/main/java/core/basesyntax/service/operation/PurchaseOperationHandler.java
```diff
@@ -0,0 +1,19 @@
...
+public class PurchaseOperationHandler implements OperationHandler {
+    @Override
+    public void doOperation(Fruit fruit) {
+        if (!FruitStorage.contains(fruit)) {
+            throw new NoSuchFruitException(OperationType.PURCHASE, fruit);
+        }
+        FruitStorage.fruits.stream()
+                .filter(x -> x.equals(fruit))
+                .findFirst().get()
+                .subtract(fruit.getQuantity());
```
KseniiaMakarova, line 17: let's first check if we have enough fruit

for example, if we have in storage 49 bananas, and user wants to buy 50, we should throw exception

src/main/java/core/basesyntax/service/impl/TransactionServiceImpl.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.TransactionService;
+import core.basesyntax.strategy.OperationStrategy;
+
+public class TransactionServiceImpl implements TransactionService {
+    private static final OperationStrategy OPERATION_STRATEGY =
+            new OperationStrategyImpl(new OperationHandlerProvider().getOperationHandlers());
```
KseniiaMakarova, line 8:
```suggestion
    private final OperationStrategy operationStrategy =
            new OperationStrategyImpl(new OperationHandlerProvider().getOperationHandlers());
```
this is not a magic number or string, just a field

src/main/java/core/basesyntax/service/impl/TransactionServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+    private final FruitTransaction fruitTransaction;
```
KseniiaMakarova, line 9: let's pass it into method parameters

src/main/java/core/basesyntax/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    private OperationType findOperationType(String operationLetter) {
+        for (OperationType operation : OperationType.values()) {
+            if (operation.getOperation().equals(operationLetter)) {
+                return operation;
+            }
+        }
+        throw new IncorrectOperationException(operationLetter);
+    }
```
KseniiaMakarova, line 25: great

KseniiaMakarova (changes requested):
almost there, already looking very good

src/main/java/core/basesyntax/dao/DataWriter.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.dao;
+
+public interface DataWriter {
+    void writeData(String storageOutput, String fileName);
```
KseniiaMakarova, line 4:
```suggestion
    void writeData(String data, String fileName);
```

src/main/java/core/basesyntax/service/impl/TransactionServiceImpl.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.TransactionService;
+import core.basesyntax.strategy.OperationStrategy;
+
+public class TransactionServiceImpl implements TransactionService {
+    private final OperationStrategy operationStrategy;
+    private final FruitTransaction fruitTransaction;
```
KseniiaMakarova, line 9: please better pass these values as doOperation() method parameters

KseniiaMakarova (approved):

src/main/java/core/basesyntax/dao/DataWriter.java
liubcheck: done!)

src/main/java/core/basesyntax/service/impl/TransactionServiceImpl.java
liubcheck: done!)

PR #337, opened 2022-03-15
Fruit shop hv

Angela1608:


src/main/java/service/ReaderImpl.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.stream.Collectors;
+import model.Fruit;
+
+public class ReaderImpl implements Reader {
+    private static final String FILE_SEPARATOR = FileSystems.getDefault().getSeparator();
+    private static final String INPUT_FILE = "src" + FILE_SEPARATOR
+            + "main" + FILE_SEPARATOR + "resources" + FILE_SEPARATOR + "input.csv";
+
+    @Override
+    public String read() {
```
MaliukDaria, line 18: Lets pass path to file as incoming parameter, so we can use ReaderImpl to read different files. And I think it's better to return list of lines List<String> instead of joining them

src/main/java/service/ReaderImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        try {
+            lines = Files.readAllLines(Path.of(INPUT_FILE));
+            line = lines.stream().skip(1).collect(Collectors.joining());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from file" + INPUT_FILE);
+        }
+        return line;
+    }
+
+    public List<Fruit> getFromCsvRow(String line) {
```
MaliukDaria, line 30: I think its breaks SRP... lets move this logic to some separate class (FruirParser or smth like this). And I think It's better to work with one line in this method. (and apply it for every line in list)

src/main/java/model/Fruit.java
```diff
@@ -0,0 +1,57 @@
+package model;
+
+public class Fruit {
```
MaliukDaria, line 3:
```suggestion
public class FruitTransaction {
```

src/main/java/Main.java
```diff
@@ -0,0 +1,40 @@
...
+        for (Fruit fruit : fromCsvRow) {
+            Operation operationHandler = operationOperationHandlerMap.get(fruit.getOperation());
+            operationHandler.proceed(fruit);
+        }
+        FruitShopDao fruitShopDao = new FruitShopDaoImpl();
+        List<Fruit> fruitList = fruitShopDao.getAll();
+        ReportCreator reportCreator = new ReportCreatorImpl();
+        String report = reportCreator.createReport(fruitList);
+        Writer writer = new WriterImpl();
+        writer.write(report);
```
MaliukDaria, line 38: same here, let's pass the file path to which we want to write information as an input parameter

src/main/java/operation/BalanceHandler.java
```diff
@@ -0,0 +1,32 @@
...
+    public void proceed(Fruit fruit) {
+        FruitShopDao fruitShopDao = new FruitShopDaoImpl();
+        if (fruitShopDao.getValue(fruit) == null) {
+            fruitShopDao.save(fruit);
+        }
+        if (fruit.getQuantity() == 0
+                || fruit.getFruit().isEmpty()
+                || fruit.getOperation() == null
+                || fruit.getFruit() == null) {
+            throw new NullPointerException("This line cannot be empty");
```
MaliukDaria, line 18: I think it's better to move file validation in Parser logic

src/main/java/operation/BalanceHandler.java
```diff
@@ -0,0 +1,32 @@
...
+        }
+        fruit.setQuantity(fruitShopDao.getValue(fruit) + fruit.getQuantity());
+        fruitShopDao.save(fruit);
+    }
+}
+
+
+
+
+
```
MaliukDaria, line 31: to many redundant empty lines

src/main/java/strategy/FruitStrategyImpl.java
```diff
@@ -0,0 +1,19 @@
...
+    private Map<Fruit.Operation, Operation> operationOperationHandlerMap;
+
+    public FruitStrategyImpl(Map<Fruit.Operation, Operation> operationOperationHandlerMap) {
+        this.operationOperationHandlerMap = operationOperationHandlerMap;
+    }
+
+    @Override
+    public Operation get(Fruit.Operation operation) {
+        return operationOperationHandlerMap.get(operation);
+    }
```
MaliukDaria, line 17:
```suggestion
    @Override
    public void proceed(Fruit fruit) {
        return operationOperationHandlerMap.get(fruit.getOperation()).proceed(fruit);
    }
```

lets move this logic to this method

MaliukDaria (changes requested):

src/main/java/service/ReaderImpl.java
Angela1608: I kept String for now 'cause it's a bit complicated to work with the List in the getFromCsvRow(String line) method

src/main/java/service/ReaderImpl.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ReaderImpl implements Reader {
+    @Override
+    public String read(String fileName) {
```
MaliukDaria, line 11:
```suggestion
    public List<String> read(String fileName) {
```

src/main/java/Main.java
```diff
@@ -0,0 +1,50 @@
...
+        operationHashMap.put(FruitTransaction.Operation.BALANCE, new BalanceHandler());
+        operationHashMap.put(FruitTransaction.Operation.PURCHASE, new PurchaseHandler());
+        operationHashMap.put(FruitTransaction.Operation.RETURN, new ReturnHandler());
+        operationHashMap.put(FruitTransaction.Operation.SUPPLY, new SupplyHandler());
+        Reader reader = new ReaderImpl();
+        FruitParser parser = new FruitParserImpl();
+        String line = reader.read(INPUT_FILE);
+        FruitTransaction fruitTransaction = parser.getFromCsvRow(line);
+        FruitStrategy fruitStrategy = new FruitStrategyImpl(operationHashMap);
+        fruitStrategy.get(fruitTransaction);
```
MaliukDaria, line 42: Давай сделаем так - 
- Reader читает файл и возвращает список из строк файла
- FruitParser принимает на вход String с одной строкой из файла, и мы применяем его к каждой строке из вписка, который возвращает Reader. По итогу получаем список из FruitTransaction
- для каждого элемента списка FruitTransaction мы применяем стратегию FruitStrategy

src/main/java/strategy/FruitStrategyImpl.java
```diff
@@ -0,0 +1,20 @@
...
+public class FruitStrategyImpl implements FruitStrategy {
+    private Map<FruitTransaction.Operation, Operation> operationOperationHandlerMap;
+
+    public FruitStrategyImpl(
+            Map<FruitTransaction.Operation, Operation> operationOperationHandlerMap) {
+        this.operationOperationHandlerMap = operationOperationHandlerMap;
+    }
+
+    @Override
+    public Operation get(FruitTransaction fruitTransaction) {
```
MaliukDaria, line 16:
```suggestion
    public Operation proceed(FruitTransaction fruitTransaction) {
```

src/main/java/model/FruitTransaction.java
```diff
@@ -0,0 +1,57 @@
...
+            return operation;
+        }
+
+        public static Operation findByAbbr(String abbr) {
+            for (Operation o : values()) {
+                if (o.getOperation().equals(abbr)) {
+                    return o;
+                }
+            }
+            return null;
```
MaliukDaria, line 54: it's bad practice to return null, lets make this method to return boolean

MaliukDaria (changes requested):

MaliukDaria (changes requested):
not all comments fixed

MaliukDaria (approved):

PR #338, opened 2022-03-15
First commit. Added jv-fruit-shop solution.

Cergei998:


src/main/java/service/activity/CheckNullBalance.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Map;
+
+public class CheckNullBalance {
+
+    public static Integer get(String fruitName, Map<String, Integer> report) {
+        Integer currentAmount = report.get(fruitName);
+        if (currentAmount == null) {
+            throw new RuntimeException("Activity without initial balance found fot the fruit: "
+                    + fruitName);
+        }
```
KseniiaMakarova, line 12: what if we have some fruit but not enough?

src/main/java/service/activity/SupplyHandler.java
```diff
@@ -0,0 +1,13 @@
+package service.activity;
+
+import java.util.Map;
+import model.FruitRecord;
+
+public class SupplyHandler implements ActivityHandler {
+
```
KseniiaMakarova, line 7: empty line

src/main/java/service/activity/SupplyHandler.java
```diff
@@ -0,0 +1,13 @@
...
+    @Override
+    public void use(FruitRecord fruitRecord, Map<String, Integer> report) {
+        report.put(fruitRecord.getFruitName(),
+                CheckNullBalance.get(fruitRecord.getFruitName(), report) + fruitRecord.getAmount());
```
KseniiaMakarova, line 11: you can check for null here as well
or at least make method non static
static methods are bad practice in most cases

KseniiaMakarova (changes requested):
overall wonderful job! I like that you split logic to different services

PR #339, opened 2022-03-25
source file processed and report file created

oleg-new:


Nick97-git (changes requested):
I would rewrite all this logic step by step:
1. Create a service class that will read from File (you have one)
2. Create a service class that will transform all lines to a specific list of models (example of such model class you can see in the description to the task)
3. Depending on the operation of the transaction you should execute some processing logic (again in the description there are examples for all operations). You should use a strategy pattern here.
4. Create a service class that will form a report
5. Create a service class that will write this report to the file (you have one)
We don't have any DB in this task, so you don't need to use this word in the project.
Please reread the description of the task and checklist, if you'll have questions tag me or come to QnA.

PR #352, opened 2022-04-26
jv_homework_fruit-shop

etezaadi:
What was done:
- Created the Storage class for holding data in HashMap;
- Created Enum class with operations type;
- The CsvReaderService interface has been created to implement the method of reading data from a csv file;
- The FruitTransaction interface has been created to implement method of process data that adds data to our Storage;
- The CsvWriterService interface has been created to implement method of writing report to new csv file.


etezaadi:
the condition indicates that it's recommended to use the pattern Strategy, but I don't understand how to apply it correctly.
I would be grateful for your help, thanks :)

PR #370, opened 2022-04-30
feat: add report from fruit store

mikenikolov:


src/main/java/service/impl/FruitServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    private final ReportService reportService = new ReportServiceImpl();
+    private final WriterService writerService = new WriterServiceImpl();
+    private final ParseService parseService = new ParseServiceImpl();
+
+    public FruitServiceImpl(StrategyService strategyService) {
+        this.strategyService = strategyService;
+    }
+
+    @Override
+    public void getData(String path) {
```
MaliukDaria, line 27:
```suggestion
    public void processData(String path) {
```

as this method void and don't return any data

src/main/java/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+    @Override
+    public List<String> read(String path) {
+        List<String> stringList = new ArrayList<>();
+        try (BufferedReader br = new BufferedReader(new FileReader(path))) {
+            br.readLine();
+            String line;
+            while ((line = br.readLine()) != null) {
+                stringList.add(line);
+            }
+            System.out.println("File read!");
```
MaliukDaria, line 21: I dont think we need to print any info in console, lets remove this (we will use logger for this purpose later). Fix in all places

src/main/java/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Can't find a file");
```
MaliukDaria, line 23:
```suggestion
            throw new RuntimeException("Can't read data from file " + path, e);
```

src/main/java/db/Database.java
```diff
@@ -0,0 +1,9 @@
+package db;
+
+import java.util.ArrayList;
+import java.util.List;
+import model.FruitTransaction;
+
+public class Database {
+    public static final List<FruitTransaction> database = new ArrayList<>();
```
MaliukDaria, line 8: Our db should save some data about fruits not about transactions. Lets implement some Map<String, Integer> that will store fruit name and fruit quantity

src/main/java/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    public String generateReport(List<FruitTransaction> transactions,
+                                 StrategyService strategyService) {
+        Map<String, Integer> reportDataMap = new HashMap<>();
+        StringBuilder reportResult = new StringBuilder();
+        for (FruitTransaction transaction : transactions) {
+            int storeQuantity = reportDataMap.get(transaction.getFruit()) == null ? 0
+                    : reportDataMap.get(transaction.getFruit());
+            reportDataMap.put(transaction.getFruit(),
+                    strategyService.get(transaction.getOperation())
+                            .operationHandler(transaction, storeQuantity));
```
MaliukDaria, line 24: We should process all FruitTransactions after we parse List<FruitTransactions> from file, and store info about fruits in Storage. Than we could get data from Storage and create report

src/main/java/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import service.WriterService;
+
+public class WriterServiceImpl implements WriterService {
+    @Override
+    public void write(String file, String report) {
+        try (OutputStreamWriter osw = new OutputStreamWriter(new FileOutputStream(file))) {
+            osw.write(report);
```
MaliukDaria, line 12: osw - not informative variable name

src/main/java/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,18 @@
+package service.impl;
+
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import service.WriterService;
+
+public class WriterServiceImpl implements WriterService {
+    @Override
+    public void write(String file, String report) {
```
MaliukDaria, line 10:
```suggestion
    public void write(String fileName, String report) {
```

src/main/java/service/impl/FruitServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        List<FruitTransaction> parsedTransactions = parseService.parse(readData);
+        dao.addAllTransaction(parsedTransactions);
+    }
+
+    @Override
+    public void saveReport(String path) {
+        writerService.write(path, getReport());
+    }
+
+    private String getReport() {
```
MaliukDaria, line 38:
```suggestion
    private String createReport() {
```

and lets add this method to FruitService interface

MaliukDaria (changes requested):
We sh

src/main/java/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+public class ReaderServiceImpl implements ReaderService {
+    private static final String SPLIT_SYMBOL = ",";
+    private static final int OPERATION_INDEX = 0;
+    private static final int FRUIT_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public List<FruitTransaction> read(String path) {
+        List<FruitTransaction> stringList = new ArrayList<>();
+        try (BufferedReader br = new BufferedReader(new FileReader(path))) {
```
MaliukDaria, line 21:
```suggestion
        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(path))) {
```

src/main/java/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class ReaderServiceImpl implements ReaderService {
+    private static final String SPLIT_SYMBOL = ",";
+    private static final int OPERATION_INDEX = 0;
+    private static final int FRUIT_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public List<FruitTransaction> read(String path) {
+        List<FruitTransaction> stringList = new ArrayList<>();
```
MaliukDaria, line 20:
```suggestion
        List<FruitTransaction> fruitTransactions = new ArrayList<>();
```

src/main/java/model/Operation.java
```diff
@@ -0,0 +1,27 @@
+package model;
+
+public enum Operation {
```
MaliukDaria, line 3: Lets move this enum into FruitTransaction class

src/main/java/service/impl/ParseServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import model.FruitTransaction;
+import service.ParseService;
+import strategy.StrategyService;
+
+public class ParseServiceImpl implements ParseService {
+
+    @Override
+    public Map<String, Integer> parse(List<FruitTransaction> fruitTransactions,
+                                      StrategyService strategyService) {
+        Map<String, Integer> fruitMap = new HashMap<>();
```
MaliukDaria, line 15: I think it's better not to store data in this map, but provide every transaction with Handlers over Map in Database.database

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #377, opened 2022-05-03
a report has been generated and written into a csv file

Roman-Kachmaryk:


src/main/java/core/basesyntax/strategy/PurchaseOperationHandler.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.model.Transaction;
+import core.basesyntax.storage.Storage;
+
+public class PurchaseOperationHandler implements OperationHandler {
+    @Override
+    public void process(Transaction transaction) {
+        if (Storage.storage.get(transaction.getFruit()) == null) {
```
ViktoriiaIhnatiuk, line 9: You can use initialQuantity instead of Storage.storage.get(transaction.getFruit())

src/main/java/core/basesyntax/strategy/BalanceOperationHandler.java
```diff
@@ -0,0 +1,15 @@
...
+
+import core.basesyntax.model.Transaction;
+import core.basesyntax.storage.Storage;
+
+public class BalanceOperationHandler implements OperationHandler {
+    @Override
+    public void process(Transaction transaction) {
+        Integer initialQuantity = Storage.storage.get(transaction.getFruit());
+        Storage.storage.put(transaction.getFruit(),
+                initialQuantity == null
```
ViktoriiaIhnatiuk, line 11: 1. This check is redundant, 
2. At the same time, in some case we can get incorrect data, so we don`t need to sum balance and initial quantity,
we shoul just put balance value to our HashMap

ViktoriiaIhnatiuk (approved):

src/main/java/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class ReaderServiceImpl implements ReaderService {
+    @Override
+    public List<String> read(String filePath) {
+        List<String> list;
+
+        try {
+            list = new ArrayList<>(Files.readAllLines(Paths.get(filePath)));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file ", e);
```
Sofasmile, line 18: давай додамо шлях до файлу до повідомлення про помилку

src/main/java/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,22 @@
...
+        }
+        return list;
```
Sofasmile, line 20:
```suggestion
        try {
            return new ArrayList<>(Files.readAllLines(Paths.get(filePath)));
        } catch (IOException e) {
            throw new RuntimeException("Can't read the file ", e);
        }
```

Sofasmile (changes requested):

Sofasmile (approved):

PR #391, opened 2022-06-01
implemented services for processing data from fruit store

Anastasia-702:


src/main/java/service/impl/FruitTransactionServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private static final int QUANTITY_INDEX = 2;
+    private FruitTransactionStrategy strategy;
+    private Storage fruitsStorage;
+
+    public FruitTransactionServiceImpl(FruitTransactionStrategy strategy, Storage fruitsStorage) {
+        this.strategy = strategy;
+        this.fruitsStorage = fruitsStorage;
+    }
+
+    public void addToStorage(String[] dataFromFile) {
```
MaliukDaria, line 20: This method responsible for parsing String to FruitTransaction class, for choosing operationHandler and saving data to Storage... it violates SPR. Let's split logic of this method into several methods

src/main/java/core/basesyntax/service/FruitTransactionService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface FruitTransactionService {
+    void setOperationHandler(List<FruitTransaction> transactions);
```
MaliukDaria, line 7:
```suggestion
    void processTransactions(List<FruitTransaction> transactions);
```

MaliukDaria (approved):
Good Job

PR #404, opened 2022-06-15
Strategy implemented but no architecture fixed from mentors made yet

drapail:
Only one commit, because I have accidentally copied original repo to Intelij, and not the forked one

### DIFF
--- a/src/main/java/core/basesyntax/LineProcessor.java
+++ b/src/main/java/core/basesyntax/LineProcessor.java
@@ -1,0 +1,23 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.Operation;
+import core.basesyntax.model.Transaction;
+import core.basesyntax.service.handlers.OperationHandler;
+import core.basesyntax.service.handlers.OperationStrategyImpl;
+import java.util.List;
+
+public class LineProcessor {
+
+    public static void processLine(List<Transaction> transactionList) {
+        for (Transaction transaction: transactionList) {
+            Operation operation = transaction.getOperation();
+            Fruit fruit = transaction.getFruit();
+            Integer quantity = transaction.getQuantity();
+
+            OperationHandler operationHandler = OperationStrategyImpl.getOperationHandlersMap()
+                    .get(operation);
+            operationHandler.processOperation(operation, fruit, quantity);
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/db/TransactionList.java
+++ b/src/main/java/core/basesyntax/db/TransactionList.java
@@ -1,0 +1,13 @@
+package core.basesyntax.db;
+
+import core.basesyntax.model.Transaction;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TransactionList {
+    private static List<Transaction> transactionList = new ArrayList<>();
+
+    public static List<Transaction> getTransactionList() {
+        return transactionList;
+    }
+}
--- a/src/main/java/core/basesyntax/model/Operation.java
+++ b/src/main/java/core/basesyntax/model/Operation.java
@@ -1,0 +1,38 @@
+package core.basesyntax.model;
+
+public enum Operation {
+    BALANCE("b"),
+    SUPPLY("s"),
+    PURCHASE("p"),
+    RETURN("r");
+
+    private String operation;
+
+    Operation(String operation) {
+        this.operation = operation;
+    }
+
+    public String getOperation() {
+        return operation;
+    }
+
+    public void setOperation(String operation) {
+        this.operation = operation;
+    }
+
+    public static Operation getOperationFromString(String operation) {
+        if (operation.equals("b")) {
+            return BALANCE;
+        }
+        if (operation.equals("s")) {
+            return SUPPLY;
+        }
+        if (operation.equals("p")) {
+            return PURCHASE;
+        }
+        if (operation.equals("r")) {
+            return RETURN;
+        }
+        return null;
+    }
+}
--- a/src/main/java/core/basesyntax/model/Transaction.java
+++ b/src/main/java/core/basesyntax/model/Transaction.java
@@ -1,0 +1,45 @@
+package core.basesyntax.model;
+
+public class Transaction {
+    private Operation operation;
+    private Fruit fruit;
+    private Integer quantity;
+
+    public Transaction(String transactionString) {
+        if (transactionString.equals("null")) {
+            this.operation = null;
+            this.fruit = null;
+            this.quantity = null;
+            return;
+        }
+
+        String[] splittedCsvLine = transactionString.split(",");
+        this.operation = Operation.getOperationFromString(splittedCsvLine[0]);
+        this.fruit = new Fruit(splittedCsvLine[1]);
+        this.quantity = Integer.valueOf(splittedCsvLine[2]);
+    }
+
+    public Operation getOperation() {
+        return operation;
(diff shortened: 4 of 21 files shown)


src/main/java/core/basesyntax/service/CsvFormatter.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax.service;
+
+import core.basesyntax.db.TransactionList;
+import core.basesyntax.model.Transaction;
+
+public class CsvFormatter {
+
+    public static void csvToStringArrayList(String csvString) {
```
MaliukDaria, line 8: It's better not to use `static`  https://medium.com/att-israel/should-you-avoid-using-static-ae4b58ca1de5 
Lets remove static from all classes in the programm

src/main/java/core/basesyntax/service/CsvReader.java
```diff
@@ -0,0 +1,27 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+
+public class CsvReader {
+    public static String readFile(String filePath) {
+        File inputFile = new File(filePath);
+        StringBuilder stringBuilder = new StringBuilder();
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(inputFile));
```
MaliukDaria, line 14: I think we should use try-with-resources here

src/main/java/core/basesyntax/db/TransactionList.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.db;
+
+import core.basesyntax.model.Transaction;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TransactionList {
```
MaliukDaria, line 7: I don't think we need to store TransactionList(lets just store data about fruits), lets just get list of transactions from parser and process them

src/main/java/core/basesyntax/model/Transaction.java
```diff
@@ -0,0 +1,45 @@
...
+            this.operation = null;
+            this.fruit = null;
+            this.quantity = null;
+            return;
+        }
+
+        String[] splittedCsvLine = transactionString.split(",");
+        this.operation = Operation.getOperationFromString(splittedCsvLine[0]);
+        this.fruit = new Fruit(splittedCsvLine[1]);
+        this.quantity = Integer.valueOf(splittedCsvLine[2]);
```
MaliukDaria, line 19: Let's move logic of parcing String to transaction in ParserService

src/main/java/core/basesyntax/service/CsvFormatter.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax.service;
+
+import core.basesyntax.db.TransactionList;
+import core.basesyntax.model.Transaction;
+
+public class CsvFormatter {
```
MaliukDaria, line 6:
```suggestion
public class TransactionParserImpl {
```

lets create some interface for this implementation (same for other services)

src/main/java/core/basesyntax/service/handlers/OperationStrategyImpl.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Map;
+
+public class OperationStrategyImpl implements OperationStrategy {
+    private static Map<Operation, OperationHandler> operationHandlersMap = new HashMap<>() {{
+            put(Operation.BALANCE, new BalanceOperationHandler());
+            put(Operation.SUPPLY, new SupplyOperationHandler());
+            put(Operation.PURCHASE, new PurchaseOperationHandler());
+            put(Operation.RETURN, new ReturnOperationHandler());
+        }};
+
```
MaliukDaria, line 14: Lets path this Map through OperationStrategyImpl constructor

src/main/java/core/basesyntax/LineProcessor.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.Operation;
+import core.basesyntax.model.Transaction;
+import core.basesyntax.service.handlers.OperationHandler;
+import core.basesyntax.service.handlers.OperationStrategyImpl;
+import java.util.List;
+
+public class LineProcessor {
```
MaliukDaria, line 10:
```suggestion
public class TransactionProcessor {
```

and lets move this class to service package(same for ReportFormatter)

src/main/java/core/basesyntax/LineProcessor.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public static void processLine(List<Transaction> transactionList) {
```
MaliukDaria, line 12:
```suggestion
    public static void process(List<Transaction> transactionList) {
```

src/main/java/core/basesyntax/service/handlers/OperationStrategyImpl.java
```diff
@@ -0,0 +1,28 @@
...
+    public OperationStrategyImpl(Map<Operation, OperationHandler> operationHandlersMap) {
+        OperationStrategyImpl.operationHandlersMap = operationHandlersMap;
+    }
+
+    public static Map<Operation, OperationHandler> getOperationHandlersMap() {
+        return operationHandlersMap;
+    }
```
MaliukDaria, line 21: I don't think we need this method as ve have `get(Operation operation)` method

src/main/java/core/basesyntax/model/Operation.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax.model;
+
+public enum Operation {
```
MaliukDaria, line 3: lets move this emun to Transaction class

src/main/java/core/basesyntax/model/Operation.java
```diff
@@ -0,0 +1,38 @@
...
+            return SUPPLY;
+        }
+        if (operation.equals("p")) {
+            return PURCHASE;
+        }
+        if (operation.equals("r")) {
+            return RETURN;
+        }
+        return null;
+    }
```
MaliukDaria, line 37: I think it's better to use loop here to find enum by String, or build some map. Otherwise, if we add a new type of operation, we will have to change this method. Let's try to implement more universal solutions

src/main/java/core/basesyntax/service/CsvWriter.java
```diff
@@ -0,0 +1,17 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class CsvWriter {
+    public static void generateCsvReport(String pathname, String report) {
+        File file = new File(pathname);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file));) {
+            bufferedWriter.write(report);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file \"" + file + "\"", e);
```
MaliukDaria, line 14:
```suggestion
            throw new RuntimeException("Can't write to file \"" + file + "\"", e);
```

MaliukDaria (changes requested):

drapail:
Here I made several importal changes to the project's logic:
1) CSV file is now parsed to the List<String> directly, this allowed to get rid of one extra class that was parsing CSV to String and then String to String[] by "," separator;
2) No class calls Strategy, Strategy is now called in Main method
3) getOperationFromString is now populated in a separate class, instead of inside enum Class.

PR #405, opened 2022-06-15
FruitShop implemented

antonEEmus:


NikitaSemisoshenko (approved):
Good job!

src/main/java/core/basesyntax/strategy/transaction/FruitTransaction.java
```diff
@@ -0,0 +1,69 @@
...
+        setOperation(operation);
+        setFruit(fruit);
+        setQuantity(quantity);
+    }
+
+    public FruitTransaction(String operation, String fruit, int quantity) {
+        setOperation(operation);
+        setFruit(fruit);
+        setQuantity(quantity);
+    }
```
KostyaNikiforov, line 18:
```suggestion
```

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class Main {
+    public static void main(String[] args) {
+        Map<String, TransactionHandler> transactionHandlersMap = new HashMap<>();
+        transactionHandlersMap.put("b", new BalanceTransactionHandler());
+        transactionHandlersMap.put("s", new SupplyTransactionHandler());
+        transactionHandlersMap.put("p", new PurchaseTransactionHandler());
+        transactionHandlersMap.put("r", new ReturnTransactionHandler());
```
KostyaNikiforov, line 29: actually, it would be better to store these values as an enum, but let it be) I like your style!

KostyaNikiforov (approved):
🔥

PR #421, opened 2022-06-23
created fruitshop

DimaPodakin:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,54 @@
...
+                new PurchaseOperationHandlerImpl());
+        operationHandlerMap.put(FruitTransaction.Operation.RETURN,
+                new ReturnOperationHandlerImpl());
+        operationHandlerMap.put(FruitTransaction.Operation.SUPPLY,
+                new SupplyOperationHandlerImpl());
+
+        Reader reader = new ReaderImpl();
+        List<String> dataFromFile = reader.fileReader(FILEPATH_DAILY_RECORDS);
+        Parser parser = new ParserImpl();
+        List<FruitTransaction> fruitTransactionList = parser.parseDataFromFile(dataFromFile);
```
gdpbalt, line 44: I think you should rename parseDataFromFile() to something like parse(),
because parseDataFromFile() actually not work with file

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,54 @@
...
+
+        OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlerMap);
+        FruitService fruitService = new FruitServiceImpl(operationStrategy);
+        fruitService.processData(fruitTransactionList);
```
gdpbalt, line 48: You should move action from fruitService.processData(fruitTransactionList) to Main class
Mentors say do that

gdpbalt (changes requested):
You should add Dao level and work with storage via Dao
Now you work with Storage directly from Service level, this is wrong

liliia-ponomarenko (approved):
LGTM 🔥

PR #439, opened 2022-08-02
Implemented FruitShop logic, using strategy pattern and SOLID principes

ssplot56:


src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+public class ReportServiceImpl implements ReportService {
+    private final FruitDao fruitDao;
+
+    public ReportServiceImpl(FruitDao fruitDao) {
+        this.fruitDao = fruitDao;
+    }
+
+    @Override
+    public String getReport() {
+        StringBuilder report = new StringBuilder("fruit,quantity");
```
aaananas, line 17: It is a special service for reports, it would be good to move the header to constants

src/main/java/core/basesyntax/service/impl/ParseServiceImpl.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.ParserService;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ParseServiceImpl implements ParserService {
```
aaananas, line 8:
```suggestion
public class ScvParseServiceImpl implements ParserService {
```
In fact you are parsing data from SCV file in this service

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.dao.FruitDao;
+import core.basesyntax.model.Fruit;
+import core.basesyntax.service.ReportService;
+import java.util.List;
+
+public class ReportServiceImpl implements ReportService {
```
aaananas, line 8:
```suggestion
public class ScvReportServiceImpl implements ReportService {
```
Result is also in scv format

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+    private final FruitDao fruitDao;
+
+    public ReportServiceImpl(FruitDao fruitDao) {
+        this.fruitDao = fruitDao;
+    }
+
+    @Override
+    public String getReport() {
```
aaananas, line 16: The solution is ok
But you can consider making this service less dependent on Dao. What about creating a method in FruitShopService that can return a Map<String, Integer> with all the data? And passing the Map to this method as a parameter?

aaananas (changes requested):
Really great job, just a few adjustments

src/main/java/core/basesyntax/service/impl/FruitShopServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    public void process(List<FruitTransaction> transactions) {
+        for (FruitTransaction transaction : transactions) {
+            OperationHandler handler
+                    = strategy.get(transaction.getOperation());
+            handler.handle(transaction);
+        }
+    }
+
+    @Override
+    public Map<String, Integer> getAll() {
```
aaananas, line 30:
```suggestion
    public Map<String, Integer> getAllData() {
```

aaananas (approved):

Nick97-git (approved):

PR #443, opened 2022-08-06
hw-jv-fruit-shop

serhii-davydov:


src/main/java/core/basesyntax/dao/impl/FruitDaoImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public void updateQuantity(String fruit, int quantity) {
+        if (!contains(fruit)) {
+            addFruit(fruit, quantity);
+        } else {
+            Storage.fruits.remove(fruit);
+            addFruit(fruit, quantity);
+        }
```
aaananas, line 30: if you put key-value pair in map, if key already exists there it changes the value, no need to remove, before putting new quantity

aaananas (changes requested):

aaananas (approved):

MaliukDaria (approved):

PR #512, opened 2022-09-05
fruit shop was implemented

not4und:


src/main/java/Main.java
```diff
@@ -0,0 +1,37 @@
...
+import service.strategy.impl.ReturnHandler;
+import service.strategy.impl.SupplyHandler;
+
+public class Main {
+    private static final String INPUT_CSV_PATH = "src/main/resources/input.csv";
+    private static final String REPORT_CSV_PATH = "src/main/resources/report.csv";
+    private static final FruitDao FRUIT_DAO = new FruitDaoImpl();
+    private static final ReaderService READER = new ReaderServiceImpl();
+    private static final ParserService PARSER = new ParserServiceImpl();
+    private static final WriterService WRITER = new WriteServiceImpl();
```
Rommelua, line 25: These 4 fields are not constants, better make them local and create them in the main(). 
FYI [constant naming](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

src/main/java/Main.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public static void main(String[] args) {
+        Map<Transaction.Operation, TransactionHandler> transactionMap = new HashMap<>();
+        transactionMap.put(Transaction.Operation.BALANCE, new BalanceHandler(FRUIT_DAO));
+        transactionMap.put(Transaction.Operation.PURCHASE, new PurchaseHandler(FRUIT_DAO));
+        transactionMap.put(Transaction.Operation.RETURN, new ReturnHandler(FRUIT_DAO));
+        transactionMap.put(Transaction.Operation.SUPPLY, new SupplyHandler(FRUIT_DAO));
```
Rommelua, line 32: We should not create this map here in main()
This logic should be in a separate method 
or (better) create `public class TransactionExecutor` with Map inside and `public void execute(Transaction transaction)` method. In this `execute(Transaction transaction)` move this logic `transactionMap.get(transaction.getOperation()).apply(transaction))`

src/main/java/model/Transaction.java
```diff
@@ -0,0 +1,51 @@
...
+
+        Operation(String label) {
+            this.label = label;
+        }
+
+        public String getLabel() {
+            return label;
+        }
+
+        public static Operation valueOfLabel(String label) {
```
Rommelua, line 44:
```suggestion
        public static Operation getByLabel(String label) {
```

src/main/java/model/Transaction.java
```diff
@@ -0,0 +1,51 @@
...
+            return Arrays.stream(values())
+                    .filter(o -> o.label.equals(label))
+                    .findFirst()
```
Rommelua, line 47: Better use findAny() always when order does not matter.

src/main/java/service/ParserService.java
```diff
@@ -0,0 +1,8 @@
+package service;
+
+import java.util.List;
+import model.Transaction;
+
+public interface ParserService {
```
Rommelua, line 6:
```suggestion
public interface TransactionParser {
```

src/main/java/service/ReaderService.java
```diff
@@ -0,0 +1,7 @@
+package service;
+
+import java.util.List;
+
+public interface ReaderService {
```
Rommelua, line 5: There is `Files.readAllLines()` in java core and I do not think we should reinvent the wheel creating our custom interface.

src/main/java/service/WriterService.java
```diff
@@ -0,0 +1,5 @@
+package service;
+
+public interface WriterService {
```
Rommelua, line 3: The same as for reader - we can use Files class

src/main/java/service/strategy/TransactionHandler.java
```diff
@@ -0,0 +1,7 @@
+package service.strategy;
+
+import model.Transaction;
+
+public interface TransactionHandler {
+    void apply(Transaction transaction);
```
Rommelua, line 6: better call this method _execute_ or _run_

src/main/java/dao/FruitDao.java
```diff
@@ -0,0 +1,12 @@
+package dao;
+
+import java.util.Map;
+import model.Fruit;
+
+public interface FruitDao {
+    void update(Fruit fruit, Integer quantity);
+
+    Integer getQuantityOf(Fruit fruit);
```
Rommelua, line 9:
```suggestion
    Integer getQuantity(Fruit fruit);
```

src/main/java/service/strategy/impl/PurchaseHandler.java
```diff
@@ -0,0 +1,25 @@
...
+        this.fruitDao = fruitDao;
+    }
+
+    @Override
+    public void apply(Transaction transaction) {
+        Fruit fruit = transaction.getFruit();
+        Integer fruitQuantity = fruitDao.getQuantityOf(fruit);
+        Integer transactionFruitQuantity = transaction.getQuantity();
+        if (fruitQuantity < transactionFruitQuantity) {
+            throw new RuntimeException("Not enough " + fruit.getName() + "s in stock.");
```
Rommelua, line 21: add actual and required quantity to message

Rommelua (changes requested):

src/main/java/service/impl/TransactionParserImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public List<Transaction> parse(List<String> inputList) {
+        inputList.remove(INFO_INPUT_LINE_INDEX);
+        List<Transaction> transactions = new ArrayList<>();
+        for (String input : inputList) {
+            String[] inputSplited = input.split(",");
+            transactions.add(new Transaction(
+                    Transaction.Operation.getByLabel(inputSplited[OPERATION_INDEX]),
+                    new Fruit(inputSplited[FRUIT_INDEX]),
+                    Integer.parseInt(inputSplited[QUANTITY_INDEX])
```
MaliukDaria, line 24: as advise: move logic of creating transection into separate private method to increase code readability

src/main/java/service/impl/TransactionExecutorImpl.java
```diff
@@ -0,0 +1,29 @@
...
+                break;
+            case PURCHASE:
+                new PurchaseHandler(fruitDao).execute(transaction);
+                break;
+            case SUPPLY:
+            case RETURN:
+                new AddHandler(fruitDao).execute(transaction);
+                break;
+            default:
+        }
```
MaliukDaria, line 27: we should use "Strategy" pattern here with use of Map<Operation, OperationHandler>

src/main/java/Main.java
```diff
@@ -0,0 +1,29 @@
...
+public class Main {
+    private static final String INPUT_CSV_PATH = "src/main/resources/input.csv";
+    private static final String REPORT_CSV_PATH = "src/main/resources/report.csv";
+
+    public static void main(String[] args) {
+        FruitDao fruitDao = new FruitDaoImpl();
+        TransactionParser transactionParser = new TransactionParserImpl();
+        TransactionExecutor transactionExecutor = new TransactionExecutorImpl();
+        try {
+            transactionParser.parse(Files.readAllLines(Path.of(INPUT_CSV_PATH)))
```
MaliukDaria, line 21: We should move logic or reading/writing files in separate services

MaliukDaria (changes requested):

src/main/java/Main.java
not4und: Another mentor says like "There is Files.readAllLines() in java core and I do not think we should reinvent the wheel creating our custom interface.". So i deleted Writer and Reader services.

aaananas (approved):

PR #519, opened 2022-09-16
Home work

andriy-linkov:


src/main/java/dao/ShopDaoImpl.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public void writeToFile(File file, String data) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, true))) {
+            bufferedWriter.write(FIRST_LINE_REPORT);
+            bufferedWriter.write(System.lineSeparator());
+            bufferedWriter.write(data);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write the file" + file, e);
+        }
+    }
```
zachar-chorny, line 45: Reading and writing to a file has nothing to do with the database or the Dao layer. Create two services, Writer and Reader, and put these methods there. Also, by combining these methods in one class, you violate the first principle of Solid

src/main/java/dao/ShopDaoImpl.java
```diff
@@ -0,0 +1,46 @@
...
+                value = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file" + file, e);
+        }
+        return result;
+    }
+
+    @Override
+    public void writeToFile(File file, String data) {
```
zachar-chorny, line 37: It is better to have the path to the file as the input data, not the file

src/main/java/dao/ShopDaoImpl.java
```diff
@@ -0,0 +1,46 @@
...
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, true))) {
```
zachar-chorny, line 38:
```suggestion
        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file))) {
```

src/main/java/dao/ShopDaoImpl.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class ShopDaoImpl implements ShopDao {
+    private static final String FIRST_LINE_REPORT = "fruit,quantity";
+
+    @Override
+    public List<String> readFromFile(File file) {
```
zachar-chorny, line 17: The same is better to use the path

src/main/java/dao/ShopDaoImpl.java
```diff
@@ -0,0 +1,46 @@
...
+                return Collections.emptyList();
+            }
+            while (value != null) {
+                result.add(value);
+                value = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file" + file, e);
+        }
+        return result;
```
zachar-chorny, line 33:
```suggestion
        try {
            return Files.readAllLines(Path.of(file.getPath()));
        } catch (IOException e) {
            throw new RuntimeException("Can't read the file" + file, e);
        }
```
Better use this method instead of writing logic from scratch

src/main/java/service/transaction/TransactionStrategy.java
```diff
@@ -0,0 +1,7 @@
+package service.transaction;
+
+import model.FruitTransaction;
+
+public interface TransactionStrategy {
```
zachar-chorny, line 5: You don't need this interface

src/main/java/service/ShopServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    @Override
+    public void process(File fromFile, File toFile) {
+        List<String> data = shopDao.readFromFile(fromFile);
+        transactionService.processData(data);
+        String report = transactionService.generateReport();
+        shopDao.writeToFile(toFile, report);
+    }
+}
```
zachar-chorny, line 23: You don't need this service either. You have to create methods according to the solid principle. They shouldn't call each other in a chain instead you should create services with independent logic and call them in the main method of the Main class you didn't create

src/main/java/service/TransactionServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import model.FruitTransaction;
+import service.transaction.TransactionHandler;
+import service.transaction.TransactionStrategy;
+import service.transaction.TransactionStrategyImpl;
+
+public class TransactionServiceImpl implements TransactionService {
```
zachar-chorny, line 12: This service needs to be broken down into several. You should not have different tasks in one service. Again, these are all solid principles

zachar-chorny (changes requested):

kozhukhovsky (approved):

PR #533, opened 2022-10-12
Fruit shop solid v1

EvgenyTyshchenko:


SoulARC (approved):
Looks good!

src/main/java/core/basesyntax/service/DataSupplierService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.Transaction;
+
+public interface DataSupplierService {
+    Transaction.Operation getOperationByName(String operation);
```
hronograf, line 6:
```suggestion
    Transaction.Operation mapToEnumValue(String operation);
```

src/main/java/core/basesyntax/service/impl/FileReaderServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class FileReaderServiceImpl implements FileReaderService {
+    @Override
+    public List<String> readFromFile(String pathToFile) {
+        List<String> data;
+        try {
+            data = Files.readAllLines(Path.of(pathToFile));
```
hronograf, line 14: looks like `data` is redundant variable and we can just return the value

src/main/java/core/basesyntax/service/FileWriterService.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.service;
+
+public interface FileWriterService {
+    void writeToFile(String report, String pathToFile);
```
hronograf, line 4:
```suggestion
public interface FileWriterService {
    void writeToFile(String text, String pathToFile);
```
not critical, but `FileWriterService` shouldn't know anything about reports

src/main/java/core/basesyntax/service/ReporterService.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax.service;
+
+// TODO: create a list and write it to file
+public interface ReporterService {
```
hronograf, line 4: can we remove this TODO?
also, for me `ReportService` sounds better

src/main/java/core/basesyntax/service/impl/ReporterServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Override
+    public String getReport() {
+        StringBuilder builder = new StringBuilder();
+        builder.append(FIRST_LINE).append(System.lineSeparator());
+        for (Map.Entry<Fruit, Integer> row : Storage.storage.entrySet()) {
+            String reportLine = row.getKey().getName() + CSV_COLUMNS_SPLITTER
+                    + row.getValue() + System.lineSeparator();
+            builder.append(reportLine);
+        }
```
hronograf, line 20: can we use `builder.append` instead of `+` here? (so we don't need `reportLine` variable)

src/main/java/core/basesyntax/strategy/operation/impl/Balance.java
```diff
@@ -0,0 +1,19 @@
...
+    public void apply(Transaction transaction) {
+        Fruit fruit = transaction.getFruit();
+        if (!Storage.storage.containsKey(fruit)) {
+            Storage.storage.put(fruit, transaction.getSum());
+        } else {
+            Integer currentQuantity = Storage.storage.get(fruit) + transaction.getSum();
+            Storage.storage.put(fruit, currentQuantity);
+        }
+    }
+}
```
hronograf, line 19: looks like we have code duplication here and in other implementations of `Operation`
can we move common logic to a separate class like `FruitService`?

hronograf (changes requested):
Goood!

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,39 @@
...
+        map.put(Transaction.Operation.SUPPLY, new Supply());
+        map.put(Transaction.Operation.RETURN, new Return());
+        map.put(Transaction.Operation.PURCHASE, new Purchase());
+        Strategy operationStrategy = new StrategyImpl(map);
+        List<String> data = new FileReaderServiceImpl().readFromFile(INPUT_FILE_NAME);
+        List<Transaction> transactions = new TransactionParserImpl().parse(data);
+        for (Transaction transaction : transactions) {
+            Operation handler = operationStrategy.getByOperation(transaction.getOperation());
+            handler.apply(transaction);
+        }
```
hronograf, line 35: we can move this `for` loop inside `FruitService`
also, when I left comment about `FruitService`, I meant that we can use `Balance` and other operations inside it

hronograf (changes requested):
Nice!

hronograf (approved):
Great! 🚀

PR #535, opened 2022-10-24
 fruit store by SOLID rules

Bodddd:


nazariipertsak (approved):
Good done

src/main/java/core/basesyntax/service/impl/FruitTransactionParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Optional;
+
+public class FruitTransactionParserServiceImpl implements DataParserService {
+    private static final String SEPARATED_VALUE = ",";
+    private static final int OPERATION_INDEX = 0;
+    private static final int FRUIT_NAME_INDEX = 1;
+    private static final int FRUIT_COUNT_INDEX = 2;
+
+    @Override
+    public List<FruitTransaction> analyse(List<String> fruitInfo) {
```
ChumakAn, line 18:
```suggestion
    public List<FruitTransaction> parse(List<String> fruitInfo) {
```

src/main/java/core/basesyntax/service/impl/FruitTransactionParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+        return fruitTransactions;
+    }
+
+    private FruitTransaction.Operation getOperationByStringValue(String value) {
+        Optional<FruitTransaction.Operation> operation =
+                Arrays.stream(FruitTransaction.Operation.values())
+                        .filter(x -> x.getOperation().equals(value))
+                        .findFirst();
+        return operation.orElseThrow(
+                () -> new RuntimeException("Can't found operation by this value" + value));
```
ChumakAn, line 37: would be a good practice to create custom exception and reuse, but optional

src/main/java/core/basesyntax/operation/impl/OperationServicePurchaseImpl.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax.operation.impl;
+
+import core.basesyntax.db.FruitStorage;
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.operation.OperationService;
+
+public class OperationServicePurchaseImpl implements OperationService {
```
ChumakAn, line 8: it's better to create separate classes for each operation instead of having only two - for addition and subtraction, cause it will make your code maintainable for future changes - fro example if the client changes the requirements and two addition operations will have different logic

ChumakAn (changes requested):
Good solution! Let's fix some issues and make it better 🔥

src/main/java/core/basesyntax/operation/impl/OperationServicePurchaseImpl.java
Bodddd: okay, but they will be identical

ChumakAn: for this situation you're right, but it's better to make such decision here even if we won't escape from code duplication

ChumakAn (approved):

PR #536, opened 2022-10-24
Report service impl

shmex-j:


src/main/java/core/basesyntax/service/impl/FruitReportService.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class FruitReportService implements ReportService {
+    private static final String HEADER = "fruit,quantity";
+
+    @Override
+    public String calculateReport(Map<String, Integer> data) {
+        StringBuilder report = new StringBuilder(HEADER);
+        data.forEach((key, value)
+                -> report.append(System.lineSeparator())
+                .append(key).append(",").append(value)
```
SoulARC, line 14: The separator should make constant.

SoulARC (changes requested):

SoulARC (approved):
LGTM!

src/main/java/core/basesyntax/dao/FruitShopStorageDao.java
```diff
@@ -0,0 +1,31 @@
...
+        if (!fruitStorage.getContent().containsKey(key)) {
+            throw new RuntimeException("Theres no such fruit: " + key);
+        }
+        int oldValue = fruitStorage.getContent().get(key);
+        if (quantity > oldValue) {
+            throw new RuntimeException("You can't remove from storage more than it have (have "
+                    + oldValue + ")");
+        }
+        fruitStorage.getContent().put(key, oldValue - quantity);
+    }
```
ChumakAn, line 30: don't you think this method is overcomplicated with the logic? Why don't we move this logic to handlers for operations you've added? It will help us make a maintainable code in case in future you will need to modify operations, add new ones, etc
what do you think of it?

ChumakAn (changes requested):
Interesting solution 🔥 Let's think of the idea I've wrote

PR #537, opened 2022-10-24
Fruit Shop implemented

hpavlo:


src/main/java/core/basesyntax/service/impl/ReportWriterImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    public void writeReport(String path) {
+        writeToFile(Path.of(path), createReport());
+    }
+
+    private String createReport() {
+        StringBuilder reportBuilder = new StringBuilder();
+        reportBuilder.append(FRUIT_HEADER)
+                .append(CSV_SPLITTER)
+                .append(QUANTITY_HEADER)
+                .append(System.lineSeparator());
```
CyberDogFK, line 26: Maybe here we can make one header, not separeated, and don`t make multiple call of append(), from StringBuilder?

src/main/java/core/basesyntax/service/impl/DataReaderImpl.java
```diff
@@ -0,0 +1,46 @@
...
+        for (String row : rows) {
+            String[] columns = row.split(CSV_SPLITTER);
+            fruitTransactions.add(new FruitTransaction(
+                    FruitTransaction.Operation.getOperationById(columns[TYPE_COLUMN_INDEX]),
+                    columns[FRUIT_COLUMN_INDEX],
+                    Integer.parseInt(columns[QUANTITY_COLUMN_INDEX])
+            ));
+        }
+        return fruitTransactions;
+    }
```
CyberDogFK, line 45: I think, maybe, it will be better to move this logic to FruitServiceImpl.
Becouse, in this class you have logic only about reading from file. Maybe it was more logical make Service do parsing of this data

src/main/java/core/basesyntax/strategy/activity/ActivityHandler.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.strategy.activity;
+
+import core.basesyntax.model.FruitTransaction;
+
+public interface ActivityHandler {
+    void doActivity(FruitTransaction fruitTransaction);
+}
```
CyberDogFK, line 7: Move all implementation of this interfase to package impl

src/main/resources/data.csv
```diff
@@ -0,0 +1,9 @@
+type,fruit,quantity
+b,banana,20
+b,apple,100
+s,banana,100
+p,banana,13
+r,apple,10
+p,apple,20
+p,banana,5
+s,banana,50
```
CyberDogFK, line 9: I think m we don`t need those files. So remove it. Or, add a new line in the end of files)
(new line in the end of files it's really important, you can find info about it at internet)

src/main/java/core/basesyntax/service/impl/ReportWriterImpl.java
```diff
@@ -0,0 +1,43 @@
...
+        for (Map.Entry<String, Integer> entry : Storage.fruits.entrySet()) {
```
CyberDogFK, line 27: Don`t use Storage directly. It`s bad practice. Create StorageDao class, and work with Storage through him.

CyberDogFK (changes requested):

CyberDogFK (approved):
LGTM

src/main/java/core/basesyntax/service/impl/activity/ActivityHandler.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service.impl.activity;
```
liliia-ponomarenko, line 1: the same about package ```activity```  it should  be subpackage ```strategy``` and you can name it ```impl``` because  it is the implementation of your strategy

liliia-ponomarenko (changes requested):
good job! Let’s improve your solution😉

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    private final Map<String, Integer> fruits = new HashMap<>();
```
liliia-ponomarenko, line 7: чому б одразу не зробити цю мапу статичною? тобі не потрібно буде повертати її в StorageDao і ти зможеш звертатис до неї напряму через назву класу

liliia-ponomarenko (approved):
Great🔥🔥🔥 but fix my comment before start to test your solution 😉

PR #546, opened 2022-10-30
Homework FruitShop

PavloPolovyi:
Created Storage for holding products and their amounts. Created StorageDao interface and its implementation for managing access to Storage. Created FileWorker interface and its implementation for managing read and write operations with files. Created DataExtractor interface and its implementation for managing parsing of CSV files lines with input data. Created custom exceptions classes - for better clarification of exceptions. Created intermediate model for storing current transaction info - FruitTransaction. Implemented strategy pattern for better managing of all operations that are needed to be done. Created FruitShopService interface and its implementation to provide two method for processing data from input file and write report to report file.


alex-anofriev (approved):

src/main/java/core/fruitshop/OperationType.java
```diff
@@ -0,0 +1,28 @@
+package core.fruitshop;
+
+import core.fruitshop.exceptions.IncorrectOperationTypeException;
+
+public enum OperationType {
+    BALANCE("b"), PURCHASE("p"), RETURN("r"), SUPPLY("s");
```
ChumakAn, line 6: it's better to write each value from separate line for better readability

src/main/java/core/fruitshop/OperationType.java
```diff
@@ -0,0 +1,28 @@
...
+    private static final String INCORRECT_TYPE_EXCEPTION = "Incorrect operation type. "
```
ChumakAn, line 7:
```suggestion
    private static final String INCORRECT_TYPE_EXCEPTION_MESSAGE = "Incorrect operation type. "
```

src/main/java/core/fruitshop/OperationType.java
```diff
@@ -0,0 +1,28 @@
...
+            + "Expected: " + BALANCE.getType() + ","
+            + SUPPLY.getType() + "," + RETURN.getType() + "or"
+            + PURCHASE.getType() + ", but was: ";
+    private final String type;
+    OperationType(String type) {
```
ChumakAn, line 12: you should separate this lines

src/main/java/core/fruitshop/dao/StorageDao.java
```diff
@@ -0,0 +1,9 @@
+package core.fruitshop.dao;
+
+public interface StorageDao {
+    void addProduct(String productName, int amount);
+
+    void minusAmount(String productName, int amount);
```
ChumakAn, line 6:
```suggestion
    void substractAmount(String productName, int amount);
```

src/main/java/core/fruitshop/services/impl/CsvDataExtractorImpl.java
```diff
@@ -0,0 +1,32 @@
+package core.fruitshop.services.impl;
+
+import core.fruitshop.OperationType;
+import core.fruitshop.model.FruitTransaction;
+import core.fruitshop.services.DataExtractor;
+
+public class CsvDataExtractorImpl implements DataExtractor {
```
ChumakAn, line 7:
```suggestion
public class CsvDataParserImpl implements DataParser {
```
what about it?

src/main/java/core/fruitshop/dao/StorageDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+        if (!contains(productName)) {
+            throw new NoSuchProductException(productName + NOT_EXISTS_EXCEPTION);
+        }
+        return Storage.fruitsStorage.get(productName);
+    }
+
+    private boolean contains(String productName) {
+        return Storage.fruitsStorage.containsKey(productName);
+    }
+}
```
ChumakAn, line 55: you could notice that this class is overcomplicated with the logic what violates the SOLID principles. Why don't we implement the logic of changing the amount in the OperationHandlers you have implemented?

src/main/java/core/fruitshop/services/impl/FruitShopServiceImpl.java
```diff
@@ -0,0 +1,45 @@
+package core.fruitshop.services.impl;
```
ChumakAn, line 1: it's better to call this package service, not services

src/main/java/core/fruitshop/services/impl/FruitShopServiceImpl.java
```diff
@@ -0,0 +1,45 @@
...
+    public FruitShopServiceImpl(FileWorker fileWorker, DataExtractor dataExtractor,
+                                OperationStrategy operationStrategy) {
+        this.fileWorker = fileWorker;
+        this.dataExtractor = dataExtractor;
+        this.operationStrategy = operationStrategy;
+    }
+
+    @Override
+    public void processData(String fileName) {
+        for (String line : fileWorker.readFromFile(fileName)) {
```
ChumakAn, line 28: you can process the result of the reading from file, this method should have one responsibility

ChumakAn: I mean, you're passing the file name , not the data you have read from the file, but the method is called processData, so it should actually process data, not work with file

ChumakAn (changes requested):
Let's make your solution better and re-read the common mistakes 

image

src/main/java/Main.java
```diff
@@ -0,0 +1,62 @@
...
+                        OperationType.SUPPLY, returnSupplyHandler);
+        FileWorker fileWorker = new FileWorkerImpl();
+        DataParser dataExtractor = new CsvDataParserImpl(COLUMN_SEPARATOR);
+        OperationStrategy operationStrategy = new OperationStrategyImpl(handlerMap);
+        List<String> data;
+        try {
+            data = fileWorker.readFromFile(DATA_FILE_PATH);
+        } catch (IOException e) {
+            throw new RuntimeException("Something went wrong "
+                    + "when reading data from " + DATA_FILE_PATH);
```
ChumakAn, line 42: catch this exception in the place you implement this

src/main/java/Main.java
```diff
@@ -0,0 +1,62 @@
...
+                            transaction.getAmount());
+        }
+        try {
+            fileWorker.writeToFile(REPORT_FILE_PATH, REPORT_HEADER,
+                    COLUMN_SEPARATOR, storageDao.dumpData());
+        } catch (IOException e) {
+            throw new RuntimeException("Something went wrong "
+                    + "when writing report to "
+                    + REPORT_FILE_PATH);
+        }
```
ChumakAn, line 60: same here

ChumakAn (changes requested):

ChumakAn (approved):

PR #551, opened 2022-10-31
Wrote FileReader, FileWriter, OperationHandler, OperationValidator se…

marchTooMuch:
…rvices, implemented them all, wrote FruitTransaction class

### DIFF
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,0 +1,34 @@
+package core.basesyntax;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.services.OperationHandler;
+import core.basesyntax.services.TransactionService;
+import core.basesyntax.services.impl.FileReaderImpl;
+import core.basesyntax.services.impl.FileWriterImpl;
+import core.basesyntax.services.impl.ReportServiceImpl;
+import core.basesyntax.services.impl.TransactionParserImpl;
+import core.basesyntax.services.impl.TransactionServiceImpl;
+import core.basesyntax.strategy.OperationHandlerBalanceImpl;
+import core.basesyntax.strategy.OperationHandlerImplSupply;
+import core.basesyntax.strategy.OperationHandlerPurchaseImpl;
+import core.basesyntax.strategy.OperationHandlerReturnImpl;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    public static void main(String[] args) {
+        Map<String, OperationHandler> operations = new HashMap<>();
+        operations.put("b", new OperationHandlerBalanceImpl());
+        operations.put("s", new OperationHandlerImplSupply());
+        operations.put("p", new OperationHandlerPurchaseImpl());
+        operations.put("r", new OperationHandlerReturnImpl());
+        List<String> fruitTransations = new FileReaderImpl().read("src/main/resources/input.csv");
+        List<FruitTransaction> parse = new TransactionParserImpl().parse(fruitTransations);
+        TransactionService service = new TransactionServiceImpl(operations);
+        service.handleTransaction(parse);
+        String report = new ReportServiceImpl().generateReport();
+        String path = "output.csv";
+        new FileWriterImpl().writeDate(path,report);
+    }
+}
--- a/src/main/java/core/basesyntax/services/FileWriter.java
+++ b/src/main/java/core/basesyntax/services/FileWriter.java
@@ -1,0 +1,5 @@
+package core.basesyntax.services;
+
+public interface FileWriter {
+    void writeDate(String path, String message);
+}
--- a/src/main/java/core/basesyntax/services/TransactionParser.java
+++ b/src/main/java/core/basesyntax/services/TransactionParser.java
@@ -1,0 +1,8 @@
+package core.basesyntax.services;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface TransactionParser {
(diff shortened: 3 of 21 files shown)


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    public static void main(String[] args) {
+        Map<String, OperationHandler> operations = new HashMap<>();
+        operations.put("b", new OperationHandlerBalanceImpl());
+        operations.put("s", new OperationHandlerImplSupply());
+        operations.put("p", new OperationHandlerPurchaseImpl());
+        operations.put("r", new OperationHandlerReturnImpl());
```
Bohdan20b, line 25: you have these values in operation enum, so use them instead of hardcode

src/main/java/core/basesyntax/services/impl/OperationValidatorImpl.java
```diff
@@ -0,0 +1,18 @@
...
+public class OperationValidatorImpl implements OperationValidator {
+
+    @Override
+    public FruitTransaction.Operation validate(String operation) {
+        switch (operation) {
+            case "b" : return FruitTransaction.Operation.BALANCE;
+            case "s" : return FruitTransaction.Operation.SUPPLY;
+            case "p" : return FruitTransaction.Operation.PURCHASE;
+            case "r" : return FruitTransaction.Operation.RETURN;
+            default: throw new RuntimeException("Unknow operation");
```
Bohdan20b, line 15: you can simplify this as you already have method getOperation in your FruitTransaction class.

Bohdan20b (changes requested):
Solution could be improved

src/main/java/core/basesyntax/services/impl/OperationValidatorImpl.java
marchTooMuch: I was trying to use FruitTransaction.Operation.getOperation() method immediately at case block but I cant do it

Bohdan20b: I've implemented this in kind of different approach, with accepting the object, not a string. Nevertheless, your code is working as well, so I approve it :D

Bohdan20b (approved):

src/main/java/core/basesyntax/services/FileWriter.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.services;
+
+public interface FileWriter {
+    void writeDate(String path, String message);
```
ChumakAn, line 4:
```suggestion
    void writeData(String path, String message);
```

src/main/java/core/basesyntax/services/TransactionParser.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.services;
```
ChumakAn, line 1: it's better to rename package to service

src/main/java/core/basesyntax/services/impl/FileReaderImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import java.nio.file.Path;
+import java.util.List;
+
+public class FileReaderImpl implements FileReader {
+    @Override
+    public List<String> read(String path) {
+        try {
+            return Files.readAllLines(Path.of(path));
+        } catch (IOException e) {
+            throw new RuntimeException("Cant read from file");
```
ChumakAn, line 15: it's better to make it more informational
image

src/main/java/core/basesyntax/services/impl/FileWriterImpl.java
```diff
@@ -0,0 +1,17 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class FileWriterImpl implements FileWriter {
+    @Override
+    public void writeDate(String path, String report) {
+        try {
+            Files.write(Path.of(path), report.getBytes());
+        } catch (IOException e) {
+            throw new RuntimeException("Cant write to file");
```
ChumakAn, line 14: same for exception here

src/main/java/core/basesyntax/services/impl/OperationValidatorImpl.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @Override
+    public FruitTransaction.Operation validate(String operation) {
+        switch (operation) {
+            case BALANCE : return FruitTransaction.Operation.BALANCE;
+            case SUPPLY : return FruitTransaction.Operation.SUPPLY;
+            case PURCHASE : return FruitTransaction.Operation.PURCHASE;
+            case RETURN : return FruitTransaction.Operation.RETURN;
+            default: throw new RuntimeException("Unknow operation");
+        }
```
ChumakAn, line 20: image

src/main/java/core/basesyntax/services/impl/TransactionParserImpl.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax.services.impl;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.services.OperationValidator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TransactionParserImpl implements core.basesyntax.services.TransactionParser {
```
ChumakAn, line 8:
```suggestion
public class TransactionParserImpl implements TransactionParser {
```

ChumakAn (changes requested):
Good 👍 Let's finish up and make the solution better

ChumakAn (approved):

PR #553, opened 2022-10-31
solution

vyshnevsky:


Bodddd (approved):
Looks good!

src/main/java/core/basesyntax/service/ReportService.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.service;
+
+public interface ReportService {
+    String makeReport();
```
ChumakAn, line 4:
```suggestion
    String generateReport();
```

ChumakAn: or create

src/main/java/core/basesyntax/service/FileReaderService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import java.io.File;
+import java.util.List;
+
+public interface FileReaderService {
+    List<String> readFromFile(File file);
```
ChumakAn, line 7: what do you think of passing a String as a parameter here? In this case, you won't need to make a file from path in Main

ChumakAn: same for file IO methods

ChumakAn (changes requested):
Good! Let's finish up 🚀

PR #554, opened 2022-11-01
task completed following SOLID principles

yehbreus:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,56 @@
...
+        FileReader readDB = new FileReaderImpl();
+        List<String> dataBaseFromFile = readDB.dataFromFile(READ_FROM);
+        TransitionParser transitionParser = new TransitionParserImpl(validator);
+        List<FruitTransition> fruitTransitionList
+                = transitionParser.parseTransition(dataBaseFromFile);
+        TransitionService insert
+                = new TransitionServiceImpl(new OperationHandlerStrategyImpl(operationHandlerMap));
+        insert.doTransition(fruitTransitionList);
+        ReportCreator reportCreator = new ReportCreatorImpl();
+        String report = reportCreator.createReport(FruitStorage.storage);
```
Iryna-a, line 52: Do you need to pass FruitStorage.storage to createReport() method? You have acces to it from every part of your program, you can just call it in the method

Iryna-a (changes requested):
Great job!

src/main/java/core/basesyntax/Main.java
yehbreus: Yes, you are right, I can not do it. But let's imagine that I need to make a report of another storage?) In this case, I will have to change the logic of creating a report. Therefore, I believe that passing the storage as a parameter to this method is correct.

Iryna-a (approved):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final String WRITE_TO = "src/main/resources/output.csv";
+    private static final OperationValidator validator = new OperationValidatorImpl();
+
+    public static void main(String[] args) {
+        Map<String, OperationHandler> operationHandlerMap = new HashMap<>();
+        operationHandlerMap.put(BALANCE.getOperation(), new BalanceHandlerImpl());
+        operationHandlerMap.put(PURCHASE.getOperation(), new PurchaseHandlerImpl());
+        operationHandlerMap.put(RETURN.getOperation(), new ReturnHandlerImpl());
+        operationHandlerMap.put(SUPPLY.getOperation(), new SupplyHandlerImpl());
+        FileReader readDB = new FileReaderImpl();
```
KostyaNikiforov, line 43:
```suggestion
        FileReader fileReader = new FileReaderImpl();
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,56 @@
...
+        List<String> dataBaseFromFile = readDB.dataFromFile(READ_FROM);
+        TransitionParser transitionParser = new TransitionParserImpl(validator);
+        List<FruitTransition> fruitTransitionList
+                = transitionParser.parseTransition(dataBaseFromFile);
+        TransitionService insert
```
KostyaNikiforov, line 48:
```suggestion
        TransitionService transitionService
```

src/main/java/core/basesyntax/service/FileReader.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+import java.util.List;
+
+public interface FileReader {
+    List<String> dataFromFile(String filePath);
```
KostyaNikiforov, line 6: method names must be verbs
```suggestion
    List<String> readFromFile(String filePath);
```

KostyaNikiforov (changes requested):
Awesome solution:fire:

KostyaNikiforov (approved):

PR #568, opened 2022-11-03
fruit shop - completed

vazonchik:


marchTooMuch (approved):

src/main/java/strategy/impl/PurchaseHandler.java
```diff
@@ -0,0 +1,18 @@
+package strategy.impl;
+
+import model.FruitTransaction;
+import strategy.GeneralOperation;
+import strategy.TransactionHandler;
+
+public class PurchaseHandler implements GeneralOperation {
+    private final TransactionHandler processTransaction;
```
ChumakAn, line 8:
```suggestion
    private final TransactionHandler transactionHandler;
```

src/main/java/strategy/impl/PurchaseHandler.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public PurchaseHandler() {
+        processTransaction = new TransactionHandlerImpl();
```
ChumakAn, line 11:
```suggestion
        transactionHandler = new TransactionHandlerImpl();
```

ChumakAn: same for all handlers

ChumakAn (changes requested):
Good! Minor comments

ChumakAn (approved):

PR #578, opened 2022-11-03
task complited

ZoreslavT:


src/main/java/sevice/impl/TransactionToDataBaseImpl.java
```diff
@@ -0,0 +1,31 @@
...
+    private static final int INDEX_OF_QUANTITY = 2;
+    private String spliterator;
+
+    public TransactionToDataBaseImpl(String spliterator) {
+        this.spliterator = spliterator;
+    }
+
+    @Override
+    public boolean transactionToDataBase(List<String> data,
+                                         Map<String, TransactionStrategy> strategyMap) {
```
Pavlo-Gook, line 20: Interesting solution!
I'm not sure you can use a map here, but let the mentors guide you)

Pavlo-Gook (approved):
LGTM👌
But about TransactionToDataBaseImpl I'm not sure ;)

pom.xml
```diff
@@ -19,6 +19,11 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>net.sf.opencsv</groupId>
+            <artifactId>opencsv</artifactId>
+            <version>2.3</version>
+        </dependency>
```
ChumakAn, line 26: I don't think you should use additional utils for this HW

src/main/java/dao/impl/StorageDaoImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public int get(String fruit) {
+        return DataBase.storage.get(fruit);
+    }
+
+    @Override
+    public List<String> getAll() {
+        return DataBase.storage.entrySet()
+                .stream()
+                .map(e -> e.getKey() + "," + e.getValue() + System.lineSeparator())
```
ChumakAn, line 20: "," should be a constant

src/main/java/service/CsvReaderService.java
```diff
@@ -0,0 +1,7 @@
+package service;
+
+import java.util.List;
+
+public interface CsvReaderService {
```
ChumakAn, line 5:
```suggestion
public interface FileReaderService {
```

src/main/java/service/CsvWriterService.java
```diff
@@ -0,0 +1,7 @@
+package service;
+
+import java.util.List;
+
+public interface CsvWriterService {
```
ChumakAn, line 5:
```suggestion
public interface FileWriterService {
```

src/main/java/service/ReportCreateService.java
```diff
@@ -0,0 +1,7 @@
+package service;
+
+import java.util.List;
+
+public interface ReportCreateService {
```
ChumakAn, line 5:
```suggestion
public interface ReportService {
```

src/main/java/sevice/impl/TransactionToDataBaseImpl.java
```diff
@@ -0,0 +1,31 @@
+package sevice.impl;
+
+import java.util.List;
+import java.util.Map;
+import service.TransactionToDataBase;
+import strategy.TransactionStrategy;
+
+public class TransactionToDataBaseImpl implements TransactionToDataBase {
```
ChumakAn, line 8:
```suggestion
public class TransactionServiceImpl implements TransactionService {
```

src/main/java/strategy/TransactionStrategy.java
```diff
@@ -0,0 +1,6 @@
+package strategy;
+
+public interface TransactionStrategy {
+
+    void doTransaction(String fruit, int quantity);
```
ChumakAn, line 5:
```suggestion
    void apply(String fruit, int quantity);
```

src/main/resources/report.csv
```diff
@@ -0,0 +1 @@
+
```
ChumakAn, line 1: it's better to commit a test file and a report it's generated

src/main/java/sevice/impl/TransactionToDataBaseImpl.java
ChumakAn: It's better to handle the choice of operation in the strategy class, having a map to choose from, it's better not to handle this in the service class

ChumakAn (changes requested):
Good! Let's make it better

src/main/java/strategy/ChooserStrategy.java
```diff
@@ -0,0 +1,16 @@
+package strategy;
+
+import java.util.Map;
+
+public class ChooserStrategy {
```
ChumakAn, line 5: let's rename it. Strategy it's about choosing the right option, so we are getting line 'buttery butter' aka 'масло масляне')
It can be OperationStrategy

ChumakAn (changes requested):
Good 🚀 Minor comment

ChumakAn (approved):

PR #587, opened 2022-12-07
Hw fruit shop

romanovosad87:


src/main/java/strategy/OperationStrategy.java
```diff
@@ -0,0 +1,8 @@
+package strategy;
+
+import strategy.handler.OperationHandler;
+
+public interface OperationStrategy {
+
+    OperationHandler get(String operation);
```
Rommelua, line 7: Why `String operation`? Better use enum.

src/main/java/strategy/handler/BalanceHandler.java
```diff
@@ -0,0 +1,11 @@
+package strategy.handler;
+
+import model.FruitTransaction;
+
+public class BalanceHandler implements OperationHandler {
+
```
Rommelua, line 6: line

Rommelua (changes requested):

aleksandr-bevza (approved):

PR #591, opened 2022-12-13
Hw solid solution

AntonKalenskiy:


src/main/java/service/impl/FileReaderServiceImpl.java
```diff
@@ -0,0 +1,21 @@
+package service.impl;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+import service.FileReaderService;
+
+public class FileReaderServiceImpl implements FileReaderService {
+    private static final String PATH_NAME = "src\\resources\\file.csv";
```
aleksandr-bevza, line 10: File path should not be hardcoded here. Move it to main method.

src/main/java/service/impl/FileReaderServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+    @Override
+    public List<String> readFromFile(String fileName) {
+        File file = new File(PATH_NAME);
```
aleksandr-bevza, line 14:
```suggestion
        File file = new File(fileName);
```

src/main/java/service/impl/FileWriterServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+import service.FileWriterService;
+
+public class FileWriterServiceImpl implements FileWriterService {
+    private static final String PATH_NAME = "src\\resources\\result.csv";
```
aleksandr-bevza, line 11: File path should not be hardcoded here. Move it to main method.

aleksandr-bevza (changes requested):

zachar-chorny (approved):

PR #593, opened 2022-12-18
added jv-fruit-shop homework solution

roman-yuskevych:


src/main/java/model/FruitTransaction.java
```diff
@@ -0,0 +1,57 @@
...
+            return operation;
+        }
+
+        public static Operation getByCode(String code) {
+            for (Operation operation : Operation.values()) {
+                if (operation.getOperation().equals(code)) {
+                    return operation;
+                }
+            }
+            return null;
```
aleksandr-bevza, line 54: It's better to throw exception here

aleksandr-bevza (changes requested):

PR #635, opened 2023-01-16
commit

olei23:


src/main/java/core/basesyntax/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+public class ReaderServiceImpl implements ReaderService {
+    private static final String FILE_NAME
+            = "src/main/java/core/basesyntax/resources/inputDataForPeriod.csv";
+    private static final String SEPARATE = ",";
+    private static final int INDEX_OPERATION = 0;
+    private static final int INDEX_FRUIT = 1;
+    private static final int INDEX_QUANTITY = 2;
+
+    private Transaction getFruitTransaction(String line) {
+        List<Operation> operations = new OperationDaoImpl().getListOperations();
```
karma-o, line 22: There are a couple of problems with this OperationDao and the task itself:
1. Your class is in fact not a DAO, you should not have one in this task, something like this should be more of a constant field and not a whole class with some interface.
2. you created some sort of a list, of `string operations` and their `arithmetic operations` and then use if to handle it, but it should be a proper strategy pattern implementation with: 
* `Operation` as an enum(represents our `balance`, `supply` etc.) with method `Operation forValue(String)`
* `OperationHandler` interface (has method `handle` or `apply` or `calculate`, which accepts our `Transaction` parameter) 
* 2 or more implementations of this hander(addition and subtraction) that will be processing the transactions in the right way and saving the results of that processing to the DB(which you don't have yet so let's add some mock like a `Storage` class with a Map of String fruit name and number of that fruits)
* and `OperationStrategy` interface with its implementation with a map filed of `Map<Operation, OperationHandler>` and method `OperationHandler get(Operation)` inside it

After all that you will have a clean and scalable code in other methods that are using this strategy

src/main/java/core/basesyntax/resources/inputDataForPeriod.csv
```diff
@@ -0,0 +1,8 @@
+    b,banana,20
+    b,apple,100
+    s,banana,100
+    p,banana,13
+    r,apple,10
+    p,apple,20
+    p,banana,5
+    s,banana,50
```
karma-o, line 8: Thas does not look like a proper path
image

src/main/java/core/basesyntax/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+import core.basesyntax.model.Transaction;
+import core.basesyntax.service.ReaderService;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ReaderServiceImpl implements ReaderService {
+    private static final String FILE_NAME
```
karma-o, line 14: do not hardcode the file paths into the services, what if you will need to read another file?
image
In general, any field that is not a constant we should be getting from outside the class, in our case let's try to separate all of the functionality as much as we can and call all of the needed methods and pass returned values down the road in the main method

src/main/java/core/basesyntax/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+            = "src/main/java/core/basesyntax/resources/inputDataForPeriod.csv";
+    private static final String SEPARATE = ",";
+    private static final int INDEX_OPERATION = 0;
+    private static final int INDEX_FRUIT = 1;
+    private static final int INDEX_QUANTITY = 2;
+
+    private Transaction getFruitTransaction(String line) {
```
karma-o, line 21: this is more of a `parseTransaction()` method, and this is not the responsibility of the file reader service
Create a separate service for this functionality

src/main/java/core/basesyntax/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+            if (operation.getShortName().equals(fields[INDEX_OPERATION].trim())) {
+                transaction.setOperation(operation);
+                break;
+            }
+        }
+        return transaction;
+    }
+
+    @Override
+    public List<Transaction> getListTransaction() {
```
karma-o, line 36: The responsibility of the File reader is only to read and return the data, that's it
```suggestion
    public List<String> readData() {
```

src/main/java/core/basesyntax/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import core.basesyntax.model.Transaction;
+import core.basesyntax.service.WriterService;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.PrintWriter;
+import java.util.List;
+
+public class WriterServiceImpl implements WriterService {
+    private static final String CSV_FILE_NAME
+            = "src/main/java/core/basesyntax/resources/finalReport.csv";
```
karma-o, line 12: the file name should be passed as a method parameter
```suggestion
```

src/main/java/core/basesyntax/strategy/TransactionCalculate.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax.strategy;
```
karma-o, line 1: This is not a strategy by any means.
I recommend check out the theory again

src/main/java/core/basesyntax/strategy/TransactionCalculate.java
```diff
@@ -0,0 +1,38 @@
...
+
+import core.basesyntax.model.Operation;
+import core.basesyntax.model.Transaction;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class TransactionCalculate {
```
karma-o, line 10: remove this class and move all of the testing logic to the main method

src/main/java/core/basesyntax/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,49 @@
+package core.basesyntax.impl;
```
karma-o, line 1:
```suggestion
package core.basesyntax.service.impl;
```

src/main/java/core/basesyntax/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax.impl;
```
karma-o, line 1:
```suggestion
package core.basesyntax.service.impl;
```

karma-o (changes requested):

src/main/java/core/basesyntax/impl/SeparateServiceImpl.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax.impl;
+
+import core.basesyntax.model.Transaction;
+import core.basesyntax.service.SeparateService;
+import java.util.ArrayList;
+import java.util.List;
+
+public class SeparateServiceImpl implements SeparateService {
```
kozhukhovsky, line 8: Give me a more correct name for this service

src/main/java/core/basesyntax/impl/ReaderServiceImpl.java
kozhukhovsky: move implementation to the impl package into service package

src/main/java/core/basesyntax/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+import core.basesyntax.service.ReaderService;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class ReaderServiceImpl implements ReaderService {
+
```
kozhukhovsky, line 10: remove empty line

src/main/java/core/basesyntax/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.WriterService;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.PrintWriter;
+import java.util.Map;
+
+public class WriterServiceImpl implements WriterService {
+
```
kozhukhovsky, line 11: remove empty line

src/main/java/core/basesyntax/model/Transaction.java
```diff
@@ -0,0 +1,55 @@
+package core.basesyntax.model;
+
+public class Transaction {
+
```
kozhukhovsky, line 4: remove empty line

src/main/java/core/basesyntax/strategy/PurchaseOperationHandler.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.Transaction;
+
+public class PurchaseOperationHandler implements OperationHandler {
+
```
kozhukhovsky, line 7: remove empty line

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/service/WriterServiceImpl.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.service;
+
+import core.basesyntax.db.Storage;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.PrintWriter;
+import java.util.Map;
+
+public class WriterServiceImpl implements WriterService {
```
aaananas, line 9: better create 2 separate services 
- to create report 
- to write String to file

aaananas (changes requested):

karma-o (approved):
:fire:

PR #641, opened 2023-01-22
Implemented reading from a file, data processing, creating and writin…

andrey156:
…g a report to a new file.

### DIFF
--- a/src/main/java/core/service/impl/DataParserServiceImpl.java
+++ b/src/main/java/core/service/impl/DataParserServiceImpl.java
@@ -1,0 +1,33 @@
+package core.service.impl;
+
+import core.model.FruitTransaction;
+import core.service.DataParserService;
+import java.util.ArrayList;
+import java.util.List;
+
+public class DataParserServiceImpl implements DataParserService {
+    private static final String COMA = ",";
+    private static final int OPERATION_INDEX = 0;
+    private static final int FRUIT_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+    private static final int FIRST_DATA_LINE = 1;
+
+    @Override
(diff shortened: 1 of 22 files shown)


src/main/java/core/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,22 @@
...
+    public OperationHandler getOperationHandler(FruitTransaction.Operation operation) {
+        switch (operation) {
+            case BALANCE:
+                return new BalanceOperationHandler();
+            case SUPPLY:
+                return new SupplyOperationHandler();
+            case PURCHASE:
+                return new PurchaseOperationHandler();
+            case RETURN:
+                return new ReturnOperationHandler();
```
aaananas, line 17: this should be a map
when creating OperationStrategyImpl instance pass that map with a constructor

src/main/java/core/service/impl/DataParserServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    public List<FruitTransaction> parseList(List<String> dataFromFile) {
+        List<FruitTransaction> transactions = new ArrayList<>();
+        for (int i = FIRST_DATA_LINE; i < dataFromFile.size(); i++) {
+            transactions.add(parseLine(dataFromFile.get(i)));
+        }
+        return transactions;
+    }
+
+    public FruitTransaction parseLine(String line) {
```
aaananas, line 24:
```suggestion
    private FruitTransaction parseLine(String line) {
```

aaananas (changes requested):

aaananas (approved):

PR #648, opened 2023-01-28
My solution fruit shop

viktoriia-maherko:


src/main/java/core/basesyntax/service/impl/TransactionServiceImpl.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.TransactionService;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TransactionServiceImpl implements TransactionService {
```
karma-o, line 8: This is  not a service really, more of a mapper or just a parser since it is only one-way thing
```suggestion
public class TransactionParserImpl implements TransactionParser {
```

src/main/java/core/basesyntax/service/impl/TransactionHandlerImpl.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.TransactionHandler;
+import core.basesyntax.strategy.OperationStrategy;
+import java.util.List;
+
+public class TransactionHandlerImpl implements TransactionHandler {
```
karma-o, line 8: this is more of a central service that contains the main logic, so I think this name fits more, considering that we have `OperationHandlers` with method `handle()` which can confuse us
```suggestion
public class FruitShopServiceImpl implements FruitShopService {
```

src/main/java/core/basesyntax/service/impl/TransactionHandlerImpl.java
```diff
@@ -0,0 +1,21 @@
...
+    private final OperationStrategy operationStrategy;
+
+    public TransactionHandlerImpl(OperationStrategy operationStrategy) {
+        this.operationStrategy = operationStrategy;
+    }
+
+    @Override
+    public void parse(List<FruitTransaction> parseList) {
```
karma-o, line 16:
```suggestion
    public void processTransactions(List<FruitTransaction> transactions) {
```

karma-o (commented):
Good job! :fire: :fire_engine: just left some minor naming comments

karma-o (changes requested):
missed the request changes button :nerd_face:

karma-o (approved):

PR #671, opened 2023-02-23
Hw jv fruit shop

yaaariiik:


src/main/java/service/CsvFileWriterService.java
```diff
@@ -0,0 +1,26 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.List;
+
+public class CsvFileWriterService implements FileWriterService {
+
+    @Override
+    public File saveToFile(File reportFile, List<String> report) {
+        report.stream()
+                .forEach(l -> write(reportFile, l));
```
olekskov, line 14: We can call .forEach() method of list without stream

src/main/java/service/transaction/BalanceTransactionHandler.java
```diff
@@ -0,0 +1,9 @@
+package service.transaction;
+
+public class BalanceTransactionHandler implements TransactionHandler {
+
+    @Override
+    public Integer calculateRemnant(Integer remnant, Integer amount) {
+        return remnant + amount;
```
olekskov, line 7: balance operation replaces current remnant, not added

src/main/java/Main.java
```diff
@@ -0,0 +1,64 @@
...
+    public static final Map<FruitTransaction.Operation,
+                        TransactionHandler> transactionHandlerMap = new HashMap<>();
+    public static final TransactionHandler balanceTransactionHandler =
+                                        new BalanceTransactionHandler();
+    public static final TransactionHandler purchaseTransactionHandler =
+                                        new PurchaseTransactionHandler();
+    public static final TransactionHandler returnTransactionHandler =
+                                        new ReturnTransactionHandler();
+    public static final TransactionHandler supplyTransactionHandler =
+                                        new SupplyTransactionHandler();
```
olekskov, line 34: can't see real reason to extract handlers to class variables, they used only once, you may just inline new ProperTransactionHandler() when initializing handlerMap

src/main/java/Main.java
```diff
@@ -0,0 +1,64 @@
...
+    public static void main(String[] args) {
+        transactionHandlerMap.put(FruitTransaction.Operation.BALANCE,
+                                    balanceTransactionHandler);
+        transactionHandlerMap.put(FruitTransaction.Operation.PURCHASE,
+                                    purchaseTransactionHandler);
+        transactionHandlerMap.put(FruitTransaction.Operation.RETURN,
+                                    returnTransactionHandler);
+        transactionHandlerMap.put(FruitTransaction.Operation.SUPPLY,
+                                    supplyTransactionHandler);
+        String pathToInputDataFile = "src/main/resources/inputData.csv";
```
olekskov, line 45: And magic values can be extracted to class constants :)

src/main/java/Main.java
```diff
@@ -0,0 +1,64 @@
...
+                            new TransactionStrategyImpl(transactionHandlerMap);
+        ReportService reportService = new ReportServiceImpl(transactionStrategy);
+        List<String> report = reportService.createReport(fruitTransactions);
+        String pathToReportFile = "src/main/resources/reportFile.csv";
+        File reportFile = new File(pathToReportFile);
+        FileWriterService fileWriterService = new CsvFileWriterService();
+        fileWriterService.saveToFile(reportFile, report);
+        List<String> readReport = csvFileReader.readFile(reportFile);
+        Printer printer = new PrinterImpl();
+        printer.print(readReport);
```
olekskov, line 62: Look, you have a plenty of raw logic in Main method. 
Please try to rethink your solution so from Main method you call some service and it's making required magic from just one command like:

fruitShopService.generateDailyReport(INPUT_FILENAME, REPORT_FILENAME);

think of what other services it will require to be initialized, add these required services to be set via constructor

and in the end your Main method will look like:
- Initialize handlers map
- Create instances of required services / dao instances / etc.
- Create your masterService with constructor that will accept lesser services
- Run just 1 command to generate report

olekskov (changes requested):
Please recheck comment in Main method, let's make solution prettier

kozhukhovsky (approved):

PR #677, opened 2023-03-05
I write solution with SOLID rules. Write WriterImpl, ReaderImpl and R…

NazarZva:
…eportCounterImpl. Also, I use pattern strategy in StrategyOptionsImpl

### DIFF
--- a/src/main/java/core/basesyntax/HelloWorld.java
+++ b/src/main/java/core/basesyntax/HelloWorld.java
@@ -1,8 +1,39 @@
 package core.basesyntax;
 
+import core.basesyntax.actions.BalanceAction;
+import core.basesyntax.actions.DoingAction;
+import core.basesyntax.actions.PurchaseAction;
+import core.basesyntax.actions.ReturnAction;
+import core.basesyntax.actions.SupplyAction;
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.ReportCounter;
+import core.basesyntax.service.Writer;
+import core.basesyntax.service.impl.ReaderImpl;
+import core.basesyntax.service.impl.ReportCounterImpl;
+import core.basesyntax.service.impl.WriterImpl;
+import core.basesyntax.strategy.StrategyOptions;
+import core.basesyntax.strategy.StrategyOptionsImpl;
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Feel free to remove this class and create your own.
  */
 public class HelloWorld {
+    public static void main(String[] args) {
+        Map<FruitTransaction.Operation, DoingAction> doingStrategy = new HashMap<>();
+        doingStrategy.put(FruitTransaction.Operation.BALANCE, new BalanceAction());
+        doingStrategy.put(FruitTransaction.Operation.SUPPLY, new SupplyAction());
+        doingStrategy.put(FruitTransaction.Operation.PURCHASE, new PurchaseAction());
+        doingStrategy.put(FruitTransaction.Operation.RETURN, new ReturnAction());
 
+        ReaderImpl reader = new ReaderImpl();
+        reader.readerTransaction("src/main/resources/testInput.csv");
+        StrategyOptions strategyOptions = new StrategyOptionsImpl(doingStrategy);
+        ReportCounter reportCounter = new ReportCounterImpl(strategyOptions);
+        Fruit[] fruits = reportCounter.countReport();
+        Writer writer = new WriterImpl();
+        writer.writeReport("src/main/resources/testOutput.csv", fruits);
+    }
 }
--- a/src/main/java/core/basesyntax/actions/DoingAction.java
+++ b/src/main/java/core/basesyntax/actions/DoingAction.java
@@ -1,0 +1,8 @@
+package core.basesyntax.actions;
+
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.FruitTransaction;
+
+public interface DoingAction {
+    Fruit getCount(FruitTransaction transaction, Fruit fruit);
+}
--- a/src/main/java/core/basesyntax/actions/PurchaseAction.java
+++ b/src/main/java/core/basesyntax/actions/PurchaseAction.java
@@ -1,0 +1,12 @@
+package core.basesyntax.actions;
+
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.FruitTransaction;
+
+public class PurchaseAction implements DoingAction {
+    @Override
+    public Fruit getCount(FruitTransaction transaction, Fruit fruit) {
+        fruit.setQuantity(fruit.getQuantity() - transaction.getQuantity());
+        return fruit;
+    }
+}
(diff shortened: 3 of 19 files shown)


src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,8 +1,39 @@
...
 public class HelloWorld {
+    public static void main(String[] args) {
+        Map<FruitTransaction.Operation, DoingAction> doingStrategy = new HashMap<>();
+        doingStrategy.put(FruitTransaction.Operation.BALANCE, new BalanceAction());
+        doingStrategy.put(FruitTransaction.Operation.SUPPLY, new SupplyAction());
+        doingStrategy.put(FruitTransaction.Operation.PURCHASE, new PurchaseAction());
+        doingStrategy.put(FruitTransaction.Operation.RETURN, new ReturnAction());
 
+        ReaderImpl reader = new ReaderImpl();
+        reader.readerTransaction("src/main/resources/testInput.csv");
```
GalubaGSV, line 32: Add to constant

src/main/java/core/basesyntax/service/impl/ReaderImpl.java
```diff
@@ -0,0 +1,50 @@
...
+            text = bufferedReader.readLine();
+            while (text != null) {
+                addFruitTransaction(text);
+                text = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can`open file" + fileName, e);
+        }
+    }
+
```
GalubaGSV, line 26: You must separate different logic. In ReaderImpl you must only read from file and save to list, and return this list. Non process data.

GalubaGSV (approved):
There are some remarks

src/main/java/core/basesyntax/actions/PurchaseAction.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.actions;
+
+import core.basesyntax.model.Fruit;
+import core.basesyntax.model.FruitTransaction;
+
+public class PurchaseAction implements DoingAction {
+    @Override
+    public Fruit getCount(FruitTransaction transaction, Fruit fruit) {
```
maxline, line 8: rename this method (here and all other classes that implements DoingAction) .
name it like `applyOretation()`, or `applyAction()`. method should be void (no returns the result) and should save the result of the operation to the Storage class.

maxline (changes requested):
good job for the first attempt! let's improve the solution

src/main/java/core/basesyntax/service/impl/SaveServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.SaveService;
+import java.util.ArrayList;
+import java.util.List;
+
+public class SaveServiceImpl implements SaveService {
+    private static final String DELIMITER_BY_SENTENCE = ",";
+
+    @Override
+    public List<FruitTransaction> saveToDb(List<String> string) {
```
karma-o, line 13:
```suggestion
    public List<FruitTransaction> saveToDb(List<String> strings) {
```

src/main/java/core/basesyntax/service/impl/SaveServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+        List<FruitTransaction> transactions = new ArrayList<>();
+        String[] message;
+        for (int i = 1; i < string.size(); i++) {
+            message = string.get(i).split(DELIMITER_BY_SENTENCE);
+            FruitTransaction fruitTransaction = new FruitTransaction();
+            fruitTransaction.setOperation(getOperation(message[0]));
+            fruitTransaction.setFruit(message[1]);
+            fruitTransaction.setQuantity(Integer.parseInt(message[2]));
```
karma-o, line 21: magic numbers

src/main/java/core/basesyntax/service/impl/SaveServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+                return FruitTransaction.Operation.BALANCE;
+            case "s":
+                return FruitTransaction.Operation.SUPPLY;
+            case "p":
+                return FruitTransaction.Operation.PURCHASE;
+            case "r":
+                return FruitTransaction.Operation.RETURN;
+            default:
+                throw new RuntimeException("Don`t have this category: " + operation);
+        }
```
karma-o, line 40: this is not the responsibility of this service, enum class itself should have some method like `FruitTransaction.Operation getByCode(String code)`
And can we do this without the switch hardcode?
https://stackoverflow.com/a/2965252

src/main/java/core/basesyntax/service/impl/FruitCounterImpl.java
```diff
@@ -0,0 +1,35 @@
...
+import core.basesyntax.model.FruitDto;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.FruitCounter;
+import core.basesyntax.strategy.StrategyOptions;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class FruitCounterImpl implements FruitCounter {
```
karma-o, line 13: this service is basically redundant now, so I think we can remove it

src/main/java/core/basesyntax/MainClass.java
```diff
@@ -0,0 +1,53 @@
...
+
+        ReaderService reader = new ReaderServiceImpl();
+        List<String> strings = reader.readTransactionWithFile(FILE_INPUT);
+
+        SaveService saved = new SaveServiceImpl();
+        saved.saveToDb(strings);
+
+        StrategyOptions strategyOptions = new StrategyOptionsImpl(operationStrategy);
+        FruitCounter fruitCounter = new FruitCounterImpl(strategyOptions);
+        List<FruitDto> fruits = fruitCounter.countReport();
```
karma-o, line 45: instead of this, we can use our strategy right here just in a for loop iterating over the transactions that we have just parsed

src/main/java/core/basesyntax/MainClass.java
```diff
@@ -0,0 +1,53 @@
...
+import core.basesyntax.strategy.StrategyOptionsImpl;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class MainClass {
+    private static final String FILE_INPUT = "src/main/resources/testInput.csv";
+    private static final String FILE_OUTPUT = "src/main/resources/testOutput.csv";
+
+    public static void main(String[] args) {
```
karma-o, line 30: would be nice to move our services out of the main method to class fields

src/main/java/core/basesyntax/actions/DoingAction.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.actions;
```
karma-o, line 1: same for the other handlers
```suggestion
package core.basesyntax.strategy.handler;
```

karma-o (changes requested):

sarakhmen (approved):
Good job :)

PR #687, opened 2023-03-11
Create solution

oleksandrCod:


src/main/java/core/basesyntax/service/impl/CalculateServiceImpl.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.StorageTransaction;
+import core.basesyntax.service.CalculateService;
+import core.basesyntax.strategy.TransactionActivity;
+import java.util.List;
+
+public class CalculateServiceImpl implements CalculateService {
+    private final TransactionActivity activity = new TransactionActivity();
```
Ivan95kos, line 9: It is better to use the constructor
```suggestion
    private final TransactionActivity activity;
```

src/main/java/core/basesyntax/service/impl/DataParseServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+public class DataParseServiceImpl implements DataParseService {
+    public static final int ACTIVITY_INDEX = 0;
+    public static final int FRUIT_INDEX = 1;
+    public static final int QUANTITY_INDEX = 2;
+    public static final String PATTERN = ",";
+
+    @Override
+    public List<StorageTransaction> getParsedData(List<String> data) {
+        return data.stream()
+                .map(d -> d.split(PATTERN))
```
Ivan95kos, line 20: do not use abbreviated names, 
fix everywhere
```suggestion
                .map(line -> line.split(PATTERN))
```

src/main/java/core/basesyntax/service/impl/DataParseServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+                .map(strings -> new StorageTransaction(getTypeActivity(strings[ACTIVITY_INDEX]),
+                        strings[FRUIT_INDEX], parseInt(strings[QUANTITY_INDEX])))
+                .collect(Collectors.toList());
+
+    }
+
+    public StorageTransaction.Operation getTypeActivity(String code) {
+        return Arrays.stream(StorageTransaction.Operation.values())
+                .filter(t -> t.getCode().equals(code))
```
Ivan95kos, line 29: same

src/main/java/core/basesyntax/service/impl/DataParseServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+                .findFirst()
+                .get();
```
Ivan95kos, line 31: it is better to throw an exception

src/main/java/core/basesyntax/strategy/TransactionActivity.java
```diff
@@ -0,0 +1,22 @@
...
+                return new BalanceActivity();
+            case RETURN:
+                return new ReturnActivity();
+            case PURCHASE:
+                return new PurchaseActivity();
+            default:
+                return new SupplyActivity();
+        }
+    }
+}
```
Ivan95kos, line 22: It is better to rewrite it like this
```suggestion
public class OperationStrategyImpl implements OperationStrategy {
    private Map<FruitTransaction.Operation, OperationHandler> operationHandlerMap;

    public OperationStrategyImpl(Map<FruitTransaction.Operation,
            OperationHandler> operationHandlerMap) {
        this.operationHandlerMap = operationHandlerMap;
    }

    @Override
    public OperationHandler get(FruitTransaction.Operation operation) {
        return operationHandlerMap.get(operation);
    }
}
```
so as not to violate solid principles

src/main/java/core/basesyntax/model/StorageTransaction.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax.model;
+
+public class StorageTransaction {
```
Ivan95kos, line 3:
```suggestion
public class FruitTransaction {
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/service/impl/ReportMakerServiceImpl.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.ReportMakerService;
+import java.util.Map;
+
+public class ReportMakerServiceImpl implements ReportMakerService {
+    private static final String HEADER = "fruit,quantity";
+    private static final String NEW_LINE = "\n";
```
olekskov, line 9: we should use System.lineSeparator() instead of "\n" to have operating system independent solution

olekskov (approved):
Good job, approved, please pay attention to comment below

PR #688, opened 2023-03-12
v1

yaroslav-ometiukh:
-- Created class Activity which holds information about item in store its amoun and operation;
-- Created ReaderService for reading data from file; 
-- Created ActivityService for converting data from file to Activity objects and collect to List<Activity>; 
-- Created OperationStrategy and OperationHandler for implementing Strategy pattern;
 -- Created ReportService class, that create report about items after processing operations; 
-- Created WriterService, which save report to file;


olekskov (approved):
Good job!

PR #689, opened 2023-03-12
SOLID FruitShop Build Success!

IlkoGood:
Please see the @Deprecated annotation.
That classes left to observe logic's flexibility and are not involved in workflow!


src/main/java/processor/DataProcessorWithStrategy.java
```diff
@@ -0,0 +1,29 @@
...
+    private final DataStorageService storageService = new DataStorageServiceImpl();
+    private final ProcessingStrategy processingStrategy = new ProcessingStrategyImpl();
+
+    @Override
+    public void makeProcessedData(List<String> input) {
+        for (String line: input) {
+            String[] values = line.split(DATA_SEPARATOR);
+            OperationalStorage.setOperationAmount(Integer.valueOf(values[VALUE_INDEX]));
+            OperationalStorage.setFruit(values[NAME_INDEX]);
+            OperationalStorage.setStoredAmount(storageService.getValue(values[NAME_INDEX]));
```
outref, line 25: in my opinion using static OperationStorage is not a good practice. You need to use some Transaction class (see hint at the bottom of readme file) to store fruitName, amount and operation fields and pass them to your strategies.

src/main/java/processor/DataProcessorWithStrategy.java
```diff
@@ -0,0 +1,29 @@
...
+import service.DataStorageService;
+import service.impl.DataStorageServiceImpl;
+import storage.OperationalStorage;
+
+public class DataProcessorWithStrategy implements DataProcessor {
+    private static final String DATA_SEPARATOR = ",";
+    private static final int OPERATION_INDEX = 0;
+    private static final int NAME_INDEX = 1;
+    private static final int VALUE_INDEX = 2;
+    private final DataStorageService storageService = new DataStorageServiceImpl();
```
outref, line 16: Common Mistakes: "All services should be independent. We shouldn’t have Strategy and call its methods in CsvFileReaderService, or we shouldn’t have CsvFileWriterService and call its methods in the Strategy class."
I think you need Dao here.

src/main/java/processor/FalseCheckUnitImpl.java
```diff
@@ -0,0 +1,22 @@
+package processor;
+
+import java.util.Map;
+import java.util.Set;
+import model.ReportException;
+import service.impl.DataStorageServiceImpl;
+
+@Deprecated
```
outref, line 8: redundant classes don't look good in a ready solution)

src/main/java/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final String NEW_LINE = System.lineSeparator();
+    private static final Set<Map.Entry<String, Integer>> ENTRIES =
+            new DataStorageServiceImpl().getFruitMap().entrySet();
+    private ReportStorage reportStorage;
+
+    @Override
+    public String getReport() {
+        makeReport();
+        return reportStorage.getReport();
+    }
```
outref, line 20:
```suggestion
```
I don't see a reason to have this method + ReportStorage, if you have a perfectly working makeReport method that returns String report

src/main/java/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+        } catch (Exception e) {
+            System.out.println("Unable to perform operations with destination " + reportPath
+                    + ", switching to default destination");
+            reportPath = defaultReportPath;
+            filePath = Path.of(reportPath);
+        }
+
+        try {
+            Files.deleteIfExists(filePath);
+            Files.write(filePath, report.getBytes());
```
outref, line 24:
```suggestion
            try {
            Path filePath = Path.of(reportPath);
            Files.deleteIfExists(filePath);
            Files.write(filePath, report.getBytes());
```
will this work?

outref (changes requested):
wow, what an interesting solution! respect to your hard work)
here are some suggestions from me:

src/main/java/service/impl/ReportServiceImpl.java
IlkoGood: That left from alfa version maybe)
I was playing with storage services and direct accessing and so on.
Thanks.

src/main/java/service/impl/WriterServiceImpl.java
IlkoGood: Of course will, but my goal was to make extra logic to handle situation, when destination folder is unavailable but there is no other reason not to create report.
So program puts report in default destination.
Leave this please to mentor)

outref (approved):
disagree with some things, but maybe I'm wrong
good job anyways!

src/main/java/processor/strategy/operation/Balance.java
```diff
@@ -0,0 +1,20 @@
...
+    @Override
+    public void handleOperation() {
+        String fruit = OperationBufferManager.getFruit();
+        Integer balance = OperationBufferManager.getOperationAmount();
+        if (dataDao.getValue(fruit) != null) {
+            System.out.println("ATTENTION! Balance operation occurred more than once for: "
+                    + fruit + "!\nCheck your storage workflow!");
+        }
+        dataDao.putValue(fruit, balance);
+    }
```
karma-o, line 19: let's just accept the transaction that we need to process as a parameter in this method
same for all the handlers

src/main/java/storage/InputStorage.java
```diff
@@ -0,0 +1,15 @@
...
+public class InputStorage {
+    private static List<String> inputData;
+
+    public static List<String> getInputData() {
+        return inputData;
+    }
+
+    public static void setInputData(List<String> inputData) {
+        InputStorage.inputData = inputData;
+    }
```
karma-o, line 14: we don't need to store some intermediate data, we can just send it straight to the strategy to handle it and then use already valid data in the report, so lets remove this

src/main/java/service/impl/LegendFilterImpl.java
```diff
@@ -0,0 +1,22 @@
...
+    private final InputDao storageService = new InputDaoImpl();
+
+    @Override
+    public void getLegendFiltered() {
+        List<String> inputData = storageService.getStorageData();
+        String legend = storageService.getStorageData().get(LEGEND_INDEX);
+        if (legend.matches(LEGEND_REGEX) && !legend.matches(CORRECT_DATA_REGEX)) {
+            inputData.remove(LEGEND_INDEX);
+        }
+    }
```
karma-o, line 21: Looks like you are just filtering out the invalid lines, but we should not skip any parts of the data that the user sent to us this is inconsistent behavior and we should not make decisions for the user on how to handle that. 
In most cases, if we encounter some error - we throw an exception and cancel the whole transaction if anything went wrong, this is called 'atomic operations' went from the databases and 'ACID', you can google that if you want, and we will learn that in the future

Overall lets modify it and move to the `TransactionParser`, to check the format of the lines

src/main/java/processor/strategy/buffer/OperationBuffer.java
```diff
@@ -0,0 +1,7 @@
+package processor.strategy.buffer;
+
+public class OperationBuffer {
+    protected static String fruit;
+    protected static Integer operationAmount;
+    protected static Integer storedAmount;
+}
```
karma-o, line 7: that looks more like a model `FruitTransaction` class than the service, so let's do that

src/main/java/processor/DataProcessorImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public void makeProcessedData(List<String> input) {
+        for (String line: input) {
+            String[] values = line.split(DATA_SEPARATOR);
+            OperationBufferManager.setOperationAmount(Integer.valueOf(values[VALUE_INDEX]));
+            OperationBufferManager.setFruit(values[NAME_INDEX]);
+            OperationBufferManager.setStoredAmount(storageService.getValue(values[NAME_INDEX]));
+            processingStrategy.makeTransaction(Operation.getByLabel(values[OPERATION_INDEX]));
+        }
+    }
```
karma-o, line 28: I dont think this class has some separate functionality, so I'd move part of it to the `TransactionParser` and the rest to the main logic

src/main/java/core/basesyntax/Launcher.java
```diff
@@ -0,0 +1,23 @@
...
+    + Check lines for consistent data "[bspr],\w++,\d++";
+    + Form a String report from calculated data in storage;
+    + Write the report in "./res/OutGoingReports/Report.csv", make sure to overwrite it;
+    + Perform additional checks for input/output files, negative values, etc.
+*/
+
+import logic.MainLogic;
+import logic.MainLogicImpl;
+
+public class Launcher {
```
karma-o, line 14: this is more of a ReadMe.md or a Javadoc type of information but we will not focus on that now

src/main/java/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import service.WriterService;
+
+public class WriterServiceImpl implements WriterService {
+
+    @Override
+    public String writeToFile(String report, String reportPath, String defaultReportPath) {
```
karma-o, line 11: default sounds like a constant var to me :)

src/main/java/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+        Path filePath;
+        try {
+            filePath = Path.of(reportPath);
+        } catch (Exception e) {
```
karma-o, line 15: specify the exception type

src/main/java/service/impl/WriterServiceImpl.java
karma-o: I would say that this is also a non-consistent behavior, and we are making a decision instead of the user, he can try to generate a report specifying the path to it, get some exceptions, and not look in the logs of your application. 
So he won't be able to find the report + this functionality was not specified in the requirements, although you did a good job, I think it is better to just throw an exception and have some constant default path in the main logic

karma-o (changes requested):
Total overkill, but I would say that it is a good job and a practice 👍 
Left some comments on simplifying and improving your solution

src/main/java/storage/InputStorage.java
IlkoGood: That was for checks ands filters, but you say it has no sense, so ok)

src/main/java/service/impl/LegendFilterImpl.java
IlkoGood: okay(

IlkoGood:
> 

Thanks)
Bogdan said we need to imagine all the possible variants... That was good for practice, yes)

src/main/java/service/impl/WriterServiceImpl.java
IlkoGood: That should go in loading screen / pop-ups instead of logs, but I don't know how to do it yet)
I'll remove everything extra.

petro-ushchuk (approved):
Good work 🔥

src/main/java/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+            throw new RuntimeException("Unable to perform operations with " + reportPath, e);
+        }
+
+        try {
+            Files.deleteIfExists(filePath);
+            Files.write(filePath, report.getBytes());
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to perform operations with " + reportPath, e);
+        }
+    }
```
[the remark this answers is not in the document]
IlkoGood: Of course!
Missed it, thanks

PR #703, opened 2023-03-14
Hw/java-fruit-shop

EliSoll:


src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,64 @@
...
+        private String code;
+
+        Operation(String code) {
+            this.code = code;
+        }
+
+        public String getCode() {
+            return code;
+        }
+
```
yurihhhT, line 56: you can add one more method for your convenience when you need to use parser.

public static Operation getByCode(String operationCode) {
            for (Operation o : Operation.values()) {
                if (o.getCode().equals(operationCode)) {
                    return o;
                }
            }
            throw new RuntimeException("Operation code is not supported: " + operationCode);
        }

src/main/java/core/basesyntax/dao/TransactionDaoCsvImpl.java
```diff
@@ -0,0 +1,59 @@
...
+                break;
+            case ("r"):
+                fruitTransaction.setOperation(FruitTransaction.Operation.RETURN);
+                break;
+            case ("p"):
+                fruitTransaction.setOperation(FruitTransaction.Operation.PURCHASE);
+                break;
+            default:
+                throw new IllegalStateException("Unexpected value: " + fields[INDEX_OPERATION]);
+        }
```
yurihhhT, line 54: You can use method getByCode to remove this switch construction.
when you split your line it will look like "b,banana,10"
String[] fields = line.split(COLUMN_SEPARATOR);
// fields[0] = "b";
// fields[1] = "banana";
// fields[2] = 20;
You can make it like
String[] row = s.split(COLUMNS_SEPARATOR);
String operation = row[TYPE];
String fruit = row[FRUIT];
int quantity = Integer.parseInt(row[QUANTITY]);
retern new FruitTransaction(FruitTransaction.Operation.getByCode(operation), fruit, quantity));

Its something about open/closed when you need just add new operation on your enum and don't need to change other classes.

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "FruitTransaction{"
+                + "operation=" + operation
+                + ", fruit='" + fruit + '\''
+                + ", quantity=" + quantity
+                + '}';
+    }
```
yurihhhT, line 39: I'm not sure that we need this method.

src/main/java/core/basesyntax/repository/FruitDB.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.repository;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class FruitDB {
+    public static final Map<String, Integer> fruitsOnStock = new LinkedHashMap<>();
+}
```
yurihhhT, line 8: We can use just HashMap
Because we have fruit like a key, it will be always change quantity depends about operation.
But its only my opinion, because I don't understand why we need linkedhashmap  -_-

src/main/java/core/basesyntax/repository/TransactionDB.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.repository;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TransactionDB {
+    public static final List<FruitTransaction> transactions = new ArrayList<>();
+}
```
yurihhhT, line 9: You can make it like one Service "TransactionParseService" with method -
List<FruitTransaction> parse(String dataFromCsv);

in impl just parse data from file and save into List<FruitTransaction>
You did it in TransactionDaoImpl

src/main/java/core/basesyntax/service/impl/ReportBuilderImpl.java
```diff
@@ -0,0 +1,25 @@
...
+        StringBuilder sb = new StringBuilder();
+        sb.append(HEADER).append(System.lineSeparator());
+        for (Map.Entry<String, Integer> entry : fruitDao.getStock().entrySet()) {
+            sb.append(entry.getKey())
+                    .append(COMMA)
+                    .append(entry.getValue())
+                    .append(System.lineSeparator());
+        }
+        return sb.toString();
+    }
```
yurihhhT, line 24: return sb.toString().trim();

Just to remove separate line after report

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,64 @@
...
+    private int quantity;
+
+    public Operation getOperation() {
+        return operation;
+    }
+
+    public void setOperation(Operation operation) {
+        this.operation = operation;
+    }
+
```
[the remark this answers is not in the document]
EliSoll: Redid my logic without setters, thanks!

src/main/java/core/basesyntax/model/FruitTransaction.java
EliSoll: Yes, thanks, did this method - try to do it with stream))

src/main/java/core/basesyntax/service/impl/ReportBuilderImpl.java
EliSoll: yes, added trim to method

yurihhhT (approved):
Good job!

src/main/java/core/basesyntax/service/operation/BalanceOperationHandler.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.service.operation;
+
+public class BalanceOperationHandler implements OperationHandler {
```
Andriy-dz, line 3: you have several identical transactions. Let's combine them

src/main/java/core/basesyntax/service/FruitService.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.service;
```
Andriy-dz, line 1: let's change the interfaces a little, because they are not completely clear right now
You should have 5 interfaces:
ReaderService - reading from a file
WriterService - writing to a file
ReportCreatorService - report creation
TransactionParserService - from list<String> makes list<Transaction>
TransactionsProcessingService - performs certain actions, depending on the input data
This will also solve the problem of dependencies of services on each other - they can be called one by one in the Main class

Andriy-dz (changes requested):
Very good solution, let's structure it a bit

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.ReaderService;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ReaderServiceImpl implements ReaderService {
```
nacenik, line 11: now your reader not only read data, but it also parses data. Follow all principles from SOLID.
Create one more interface and realization to parse data

nacenik (changes requested):
Let’s improve your solution ;)

src/main/java/core/basesyntax/service/operation/OperationHandler.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service.operation;
+
+public interface OperationHandler {
+    int doCalculation(int amount);
+
+    boolean isApplicable(String operation);
```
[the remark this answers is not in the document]
EliSoll: This method isApplicable() was adviced for us by Bohdan Chupika for fruit shop handlers during the lection yesterday - before I had map in my code.  Should I redo it one more time? [https://pastenow.ru/MF1A1](url)

src/main/java/core/basesyntax/strategy/impl/BalanceHandler.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.strategy.impl;
+
+import core.basesyntax.strategy.OperationHandler;
+
+public class BalanceHandler implements OperationHandler {
```
nacenik, line 5: Don't forget about the naming convention. Your realization should named `BalanceOperationHandler`
fix for all realizations

nacenik (approved):
Great job! You can improve your solution if you want 😉

PR #704, opened 2023-03-14
jv-fruit-shop SOLID practice

Deschna:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,43 @@
...
+import core.basesyntax.strategy.operation.SupplyOperationHandler;
+import java.io.File;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    public static final File ACTIVITIES = new File("src/main/resources/activities.csv");
+    public static final File REPORT = new File("src/main/resources/report.csv");
+    public static final Map<String, OperationHandler> OPERATIONS = new HashMap<>();
```
potrap, line 22: i think it's better to do these vars private

src/main/java/core/basesyntax/strategy/OperationHandler.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.strategy;
+
+public interface OperationHandler {
+    String COMMA_SEPARATOR = ",";
```
potrap, line 4: there’s no need to write delimiter static variables in interface

src/main/java/core/basesyntax/strategy/operation/SupplyOperationHandler.java
```diff
@@ -0,0 +1,15 @@
...
+import core.basesyntax.strategy.OperationHandler;
+
+public class SupplyOperationHandler implements OperationHandler {
+
+    @Override
+    public void accept(String data) {
+        int indexOfSeparator = data.indexOf(OperationHandler.COMMA_SEPARATOR);
+        int amount = Storage.STORAGE.get(data.substring(0, indexOfSeparator));
+        Storage.STORAGE.put(data.substring(0, indexOfSeparator),
+                amount + Integer.parseInt(data.substring(indexOfSeparator + 1)));
```
potrap, line 13: i think you need to separate this method and 3 other, because you have here two purpose: parsing and pushing data into storage. so it's better to create class which will parse data to some fruit object with some fields(operation, fruitName, quantity)

potrap (changes requested):

src/main/java/core/basesyntax/strategy/OperationStrategyImpl.java
```diff
@@ -4,13 +4,15 @@
...
-        data.remove(0);
-        for (String value : data) {
-            int indexOfSeparator = value.indexOf(OperationHandler.COMMA_SEPARATOR);
-            operationHandlerMap.get(value.substring(0, indexOfSeparator))
-                    .accept(value.substring(indexOfSeparator + 1));
+    public void accept(List<String[]> data, Map<String, OperationHandler> operationHandlerMap) {
+        for (String[] value : data) {
+            operationHandlerMap.get(value[OPERATION_INDEX])
+                    .accept(value[FRUIT_INDEX], value[QUANTITY_INDEX]);
         }
```
potrap: yeah, now it looks better. you can change this piece of code on .stream(), but overall is good. approve :)

potrap (approved):

src/main/java/core/basesyntax/service/impl/DataParseServiceImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import core.basesyntax.service.DataParseService;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class DataParseServiceImpl implements DataParseService {
+    private static final String COMMA_SEPARATOR = ",";
+    private static final int HEADER_INDEX = 0;
+
+    @Override
+    public List<String[]> parseData(List<String> data) {
```
karma-o, line 12: introduce model class `FruitTransaction` to store operation(along with `getOperationByCode()` method), fruit name, and amount as its fields. 
It is a bad practice to operate raw strings

src/main/java/core/basesyntax/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class OperationStrategyImpl implements OperationStrategy {
+    private static final int OPERATION_INDEX = 0;
+    private static final int FRUIT_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public void accept(List<String[]> data, Map<String, OperationHandler> operationHandlerMap) {
```
karma-o, line 12: map should be class-level field initialized through the constructor and this method should just return the corresponding handler and nothing else
Handlers `accept()` method may be called in another service like `TransactionProcessingService` where you will iterate over the transactions list
```suggestion
    public OpearationHandler getHandlerByOperation(FruitTransaction.Operation operation) {
```

src/main/java/core/basesyntax/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,19 @@
...
+        for (String[] value : data) {
+            operationHandlerMap.get(value[OPERATION_INDEX])
+                    .accept(value[FRUIT_INDEX], value[QUANTITY_INDEX]);
+        }
+    }
+
```
karma-o, line 18:
```suggestion
```

src/main/java/core/basesyntax/strategy/OperationHandler.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.strategy;
+
+public interface OperationHandler {
+    void accept(String fruit, String quantity);
```
karma-o, line 4:
```suggestion
    void evaluateTransaction(FruitTransaction transaction);
```

src/main/java/core/basesyntax/strategy/operation/PurchaseOperationHandler.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.strategy.operation;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.strategy.OperationHandler;
+
+public class PurchaseOperationHandler implements OperationHandler {
+    @Override
+    public void accept(String fruit, String quantity) {
+        Storage.STORAGE.put(fruit, Storage.STORAGE.get(fruit) - Integer.parseInt(quantity));
```
karma-o, line 9: what you will try to buy more than we have stocked?

src/main/java/core/basesyntax/strategy/operation/SupplyOperationHandler.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.strategy.operation;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.strategy.OperationHandler;
+
+public class SupplyOperationHandler implements OperationHandler {
+    @Override
+    public void accept(String fruit, String quantity) {
+        Storage.STORAGE.put(fruit, Storage.STORAGE.get(fruit) + Integer.parseInt(quantity));
```
karma-o, line 9: what if we do not have such fruit in storage yet?
add null checks for addition operations

karma-o (changes requested):

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    public static final Map<String, Integer> STORAGE = new HashMap<>();
```
karma-o, line 7: this is also not a constant, it is in fact constantly being changed by our application
```suggestion
    public static final Map<String, Integer> storage = new HashMap<>();
```

src/main/java/core/basesyntax/service/impl/CsvFileWriter.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.List;
+
+public class CsvFileWriter implements FileWriterService {
+    @Override
+    public void writeToFile(File file, List<String> data) {
+        try (FileWriter writer = new FileWriter(file)) {
+            for (String value : data) {
+                writer.write(value);
+            }
+            System.out.println("The report was successfully written to file " + file);
```
sarakhmen, line 16: Seems like a debug line. So let's remove it

src/main/java/core/basesyntax/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,27 @@
...
+public class OperationStrategyImpl implements OperationStrategy {
+    private static final Map<FruitTransaction.Operation,
+            OperationHandler> operations = new HashMap<>();
+
+    public OperationStrategyImpl() {
+        operations.put(FruitTransaction.Operation.BALANCE, new BalanceOperationHandler());
+        operations.put(FruitTransaction.Operation.SUPPLY, new SupplyOperationHandler());
+        operations.put(FruitTransaction.Operation.PURCHASE, new PurchaseOperationHandler());
+        operations.put(FruitTransaction.Operation.RETURN, new ReturnOperationHandler());
```
sarakhmen, line 20: I'd also recommend passing this map as constructor argument

sarakhmen (approved):
Good job :)

PR #709, opened 2023-03-14
first version of FruitShop

oleksiistepaniak:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,52 @@
...
+                new ReturnOperationHandler());
+
+        ReaderService readerService = new ReaderServiceImpl();
+        ParserService parserService = new ParserServiceImpl();
+        OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlerMap);
+        TransactionHandler transactionHandler = new TransactionHandlerImpl(operationStrategy);
+        ReportService reportService = new ReportServiceImpl();
+        WriterService writerService = new WriterServiceImpl();
+
+        List<String> strings = readerService.readDataFromFile("src/main/resources/input.csv");
```
vika-trach, line 45: "src/main/resources/input.csv" it might be better to make it a constant

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,52 @@
...
+        List<FruitTransaction> fruitTransactions = parserService.parseData(strings);
+        transactionHandler.processData(fruitTransactions);
+        String reportFromData = reportService.makeReport();
+        writerService.writeReportToFile("src/main/resources/output.csv", reportFromData);
```
vika-trach, line 49: and here too

vika-trach (changes requested):
Good job, fix a little mistakes

vika-trach (approved):

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public enum Operation {
+        BALANCE("b"),
+        SUPPLY("s"),
+        PURCHASE("p"),
+        RETURN("r");
+
+        private String code;
+
+        Operation(String code) {
```
nacenik, line 53: access modifier

src/main/java/core/basesyntax/strategy/operations/PurchaseOperationHandler.java
```diff
@@ -0,0 +1,20 @@
...
+import core.basesyntax.exceptions.NotEnoughFruitException;
+import core.basesyntax.model.FruitTransaction;
+
+public class PurchaseOperationHandler implements OperationHandler {
+    @Override
+    public void handleOperation(FruitTransaction fruitTransaction) {
+        ifPresent(fruitTransaction.getFruit(), fruitTransaction.getQuantity());
+    }
+
+    private void ifPresent(String name, int quantity) {
```
nacenik, line 13: Name `ifPresent` means that you need to check on existing but not means that the program will add some data anywhere. Rename to `processIfPresent` or sth like this

src/main/java/core/basesyntax/exceptions/NotEnoughFruitException.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.exceptions;
+
+public class NotEnoughFruitException extends RuntimeException {
```
nacenik, line 3: [It's a good practice to create a custom exception if it really needs](https://stackoverflow.com/questions/26984/what-is-the-general-rule-of-thumbs-for-creating-an-exception-in-java) (ex: for specific logic). You can use a simple RuntimeException with an understandable message instead of creating a specific exception that you will use only once in your program.

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class Storage {
+    public static final Map<String, Integer> fruits = new HashMap<>();
+
+    public static void add(String name, int quantity) {
+        if (Storage.fruits.containsKey(name)) {
+            quantity += Storage.fruits.get(name);
+        }
+        Storage.fruits.put(name, quantity);
+    }
```
nacenik, line 14: better to use exact logic in classes. because the process of adding for return and supply will be different in the future, and you have to delete this method, cause it will be unnecessary

nacenik (changes requested):
Good solution! Let’s improve it 😉

PR #725, opened 2023-03-15
added solution hw-jv-fruit-shop

HelenMatus:


src/main/java/core/basesyntax/input.csv
```diff
@@ -0,0 +1,8 @@
+b,banana,20
```
Andriy-dz, line 1: all files must be in resources

src/main/java/core/basesyntax/exceptions/NotEnoughFruitsInStorageException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.exceptions;
+
+public class NotEnoughFruitsInStorageException extends RuntimeException {
```
Andriy-dz, line 3: [It's a good practice to create a custom exception if it really needs](https://stackoverflow.com/questions/26984/what-is-the-general-rule-of-thumbs-for-creating-an-exception-in-java) (ex: for specific logic). You can use a simple RuntimeException with an understandable message instead of creating a specific exception that you will use only once in your program.

src/main/java/core/basesyntax/handlers/BalanceOperationHandler.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.handlers;
```
Andriy-dz, line 1: I would recommend moving the handler package to the strategy

src/main/java/core/basesyntax/impl/GenerateReportImpl.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax.impl;
```
Andriy-dz, line 1: I would recommend moving the impl package to services

src/main/java/core/basesyntax/impl/ReadDataFromFileImpl.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.impl;
+
+import core.basesyntax.service.ReadDataFromFile;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class ReadDataFromFileImpl implements ReadDataFromFile {
+    private static final String FILE_PATH = "src/main/java/core/basesyntax/input.csv";
```
Andriy-dz, line 10: it is better to make the path an input parameter, because in this case this method will read only from one file, without the possibility to change it

src/main/java/core/basesyntax/impl/WriteReportToFileImpl.java
```diff
@@ -0,0 +1,22 @@
...
+import core.basesyntax.service.GenerateReport;
+import core.basesyntax.service.WriteReportToFile;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Map;
+
+public class WriteReportToFileImpl implements WriteReportToFile {
+    private static final String FILE_PATH = "src/main/java/core/basesyntax/report.csv";
+    private GenerateReport generateReport = new GenerateReportImpl();
```
Andriy-dz, line 12: same with file path and this class has a dependency on another class (GenerateReport), let's fix it

src/main/java/core/basesyntax/impl/GenerateReportImpl.java
```diff
@@ -0,0 +1,37 @@
...
+        return reportBuilder.toString();
+    }
+
+    @Override
+    public void generateReport(List<String[]> data) {
+        for (String[] item : data) {
+            operationStrategy.get(Operation.getByCode(item[OPERATION_INDEX]))
+                    .apply(item[FRUIT_INDEX], Integer.parseInt(item[QUANTITY_INDEX]));
+        }
+    }
```
Andriy-dz, line 36: here you are not generating a report, you are using a strategy. It is better to make a separate class out of it

src/main/java/core/basesyntax/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class OperationStrategyImpl implements OperationStrategy {
+    private Map<Operation, OperationHandler> operationHandlerMap;
+
+    public OperationStrategyImpl() {
+        this.operationHandlerMap = new HashMap<>();
+        this.operationHandlerMap.put(Operation.BALANCE, new BalanceOperationHandler());
+        this.operationHandlerMap.put(Operation.SUPPLY, new SupplyOperationHandler());
+        this.operationHandlerMap.put(Operation.PURCHASE, new PurchaseOperationHandler());
+        this.operationHandlerMap.put(Operation.RETURN, new ReturnOperationHandler());
```
Andriy-dz, line 20: it is better to accept this data in the constructor, because you limit the number of possible strategies. It is better to create them in the main method

Andriy-dz (changes requested):
Good Job! Let`s improve your solution

src/main/java/core/basesyntax/service/GenerateStringToWrite.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+import java.util.Map;
+
+public interface GenerateStringToWrite {
```
nacenik, line 5: All services should be nouns and contain the word "Service" in their names 
Also, naming should be understandable for someone who could see your code for the first time. What does that mean `GenerateStringToWrite`? Better to name sth like "ReportMakerService", "ReportGeneratorService" etc.

Fix naming in all services and their implementations according to the naming convention.

src/main/java/core/basesyntax/service/GenerateStringToWrite.java
```diff
@@ -0,0 +1,7 @@
...
+    String stringToWrite(Map<String, Integer> fruits);
```
nacenik, line 6: The method name should be a verb, understandable, and briefly describe what It does. 
FIx method name here

src/main/java/core/basesyntax/resources/input.csv
```diff
@@ -0,0 +1,8 @@
+b,banana,20
```
nacenik, line 1: [Place the input and output files into the `src/main/resources` folder.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/solid/fruit-shop#place-the-input-and-output-files-into-the-srcmainresources-folder)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,51 @@
...
+        ReadDataFromFile readDataFromFile = new ReadDataFromFileImpl();
+        final List<String> readData = readDataFromFile.readData(FILE_PATH_TO_READ);
+
+        ProcessDataFromFile processDataFromFile = new ProcessDataFromFileImpl();
+        final List<String[]> processData = processDataFromFile.processData(readData);
+
+        for (String[] item : processData) {
+            operationStrategy.get(Operation.getByCode(item[OPERATION_INDEX]))
+                    .apply(item[FRUIT_INDEX], Integer.parseInt(item[QUANTITY_INDEX]));
+        }
```
nacenik, line 46: About this block of code: this is business logic, cause without this code your application will work incorrectly. I'd recommend you create an additional service and its implementation and put this block of code there

nacenik (changes requested):
Good solution! Let’s improve it 😉

PR #730, opened 2023-03-18
Added solution for jv-fruit-shop hw

ArinaSaara:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,47 @@
...
+        final WriterService writerService = new WriterServiceImpl();
+        FruitDao fruitDao = new FruitDaoImpl();
+        List<FruitTransaction> fruitTransactions =
+                readerService.readFromFile("src/main/resources/file.csv");
+        System.out.println(fruitTransactions);
+        FruitService fruitService = new FruitServiceImpl(fruitDao);
+        fruitService.recordFruits(fruitTransactions);
+        System.out.println(fruitService.getAllFruits());
+        FruitShopService fruitShopService = new FruitShopServiceImpl(operationsStrategy, fruitDao);
+        fruitShopService.calculateFruits(fruitTransactions);
```
andriy-shulepa, line 43: you're processing transactions in two services - FruitService and FruitShopService. it's unclear which service is responsible for which operation. Only one service should process transactions

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.db;
+
+import core.basesyntax.model.Fruit;
+import java.util.ArrayList;
+import java.util.List;
+
+public class Storage {
+    public static final List<Fruit> fruits = new ArrayList<>();
```
andriy-shulepa, line 8: it's a working solution but consider have Map where the key is the fruit name and the value is the fruit quantity. updating such storage would be easier

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+
+public class ReaderServiceImpl implements ReaderService {
+    @Override
+    public List<FruitTransaction> readFromFile(String filePath) {
```
andriy-shulepa, line 16: split the logic of this method into two services. one will read strings from the file and the other will parse these strings into FruitTransaction. with such services will follow SRP and can be used independently

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class WriterServiceImpl implements WriterService {
+    private final String fileName = "src/main/resources/report.csv";
+
+    @Override
+    public void writeToFile(List<Fruit> fruits) {
```
andriy-shulepa, line 15: also split this logic into two services: generating report to string and writing this string to file

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+                .map(this::getFromCsvRow)
+                .collect(Collectors.toList());
+    }
+
+    private FruitTransaction getFromCsvRow(String line) {
+        String[] fields = line.split(",");
+        FruitTransaction fruitTransaction = new FruitTransaction();
+        fruitTransaction.setOperation(parseOperation(fields[0]));
+        fruitTransaction.setFruit(fields[1]);
+        fruitTransaction.setQuantity(Integer.parseInt(fields[2]));
```
andriy-shulepa, line 34: magic numbers should be extracted to constants

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        return fruitTransaction;
+    }
+
+    private FruitTransaction.Operation parseOperation(String operation) {
```
andriy-shulepa, line 38: this method can be moved to Operation enum and renamed to getByCode

andriy-shulepa (changes requested):

PR #742, opened 2023-04-09
Hw fruit shop x750l

iamchgi:


olekskov (changes requested):
It's very hard make full review at the moment due to major problems.
Please fix main issues to proceed further:
1) Storage is something where we store some amount of some goods. You have List of transactions there. It's not a storage, it's some kind of journal on accountant table
2) Due to this point you have absolutely unintuitive interfaces that doing something and all have just apply() method. Dao should perform understandable operations with our database (get amount of product, change amount of product etc. applicable for us)

So please restructure you storage to store different products and their quantity.
Rewrite dao layer with 1 ProductDao with some methods that will change quantity for proper product in our storage

src/main/java/core/basesyntax/function/impl/StringToFruitTransaction.java
```diff
@@ -0,0 +1,20 @@
...
+import core.basesyntax.model.FruitTransaction;
+import java.util.function.Function;
+
+public class StringToFruitTransaction implements Function<String,FruitTransaction> {
+
+    public static final String TRANSACTION_STRING_SEPARATOR = ",";
+    public static final int TRANSACTION = 0;
+    public static final int FRUIT = 1;
+    public static final int QUANTITY = 2;
+
```
okuzan, line 12: add _INDEX suffixes

src/main/java/core/basesyntax/function/impl/StringToFruitTransaction.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax.function.impl;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.function.Function;
+
+public class StringToFruitTransaction implements Function<String,FruitTransaction> {
+
+    public static final String TRANSACTION_STRING_SEPARATOR = ",";
```
okuzan, line 8:
```suggestion
    public static final String TRANSACTION_SEPARATOR = ",";
```

src/main/java/core/basesyntax/function/impl/StringToFruitTransaction.java
```diff
@@ -0,0 +1,20 @@
...
+    public static final int TRANSACTION = 0;
+    public static final int FRUIT = 1;
+    public static final int QUANTITY = 2;
+
+    @Override
+    public FruitTransaction apply(String stringFruitTransaction) {
+        String[] strings = stringFruitTransaction.split(TRANSACTION_STRING_SEPARATOR);
+        FruitTransaction fruitTransaction =
+                new FruitTransaction(strings[TRANSACTION],strings[FRUIT],strings[QUANTITY]);
```
okuzan, line 17: use Ctrl Alt L for automatic formatting before committing
here should be spaces after each comma

src/main/java/core/basesyntax/model/FruitQuantity.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax.model;
+
+public class FruitQuantity {
```
okuzan, line 3: Why do you need this model? Your storage isn't List of FruitQuantity, it's hashMap of Fruits and their respective quantities. IMO you won't need this class when you change the logic according to other comments

src/main/java/core/basesyntax/service/CalculateBalanceService.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+import java.util.Map;
+
+public interface CalculateBalanceService {
```
okuzan, line 7: You should have just one report generator, and it should deal with Storage/Dao directly, why would there be arguments?
review all your methods that deal with Map<String,Integer> and check that you really need them or is it over-engineering

okuzan (changes requested):
Non-idiomatic naming makes it hard to follow the logic you put behind your app. Remember that method names should be verbs, class names should be nouns, and interfaces - nouns/adjectives. First of all, focus your attention on dealing with Storage according to the comments, and also rewrite Handlers. Make sure your code is clean (no unused variables and classes)

okuzan (changes requested):
Non-idiomatic naming makes it hard to follow the logic you put behind your app. Remember that method names should be verbs, class names should be nouns, and interfaces - nouns/adjectives. First of all, focus your attention on dealing with Storage according to the comments, and also rewrite Handlers. Make sure your code is clean (no unused variables and classes)

src/main/java/core/basesyntax/function/impl/StringToTransactionConverter.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax.function.impl;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.function.Function;
+
+public class StringToTransactionConverter implements Function<String, FruitTransaction> {
```
okuzan, line 6:
```suggestion
public class TransactionParser implements Function<String, FruitTransaction> {
```

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,63 @@
...
+        this.quantity = Integer.parseInt(quantity);
+    }
+
+    private Operation getTransactionName(String code) {
+        for (Operation operation : Operation.values()) {
+            if (operation.code.equals(code)) {
+                return operation;
+            }
+        }
+        return null;
```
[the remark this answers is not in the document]
okuzan: not fixed

src/main/java/core/basesyntax/service/impl/FileWriterServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+import core.basesyntax.service.FileWriterService;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+
+public class FileWriterServiceImpl implements FileWriterService {
+
```
okuzan, line 11: empty line

src/main/java/core/basesyntax/service/impl/ReportCreatorServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+        Map<String, Integer> reportMap = storageRecordsGetter.get();
+        List<String> dailyReportList = reportMap
+                .entrySet()
+                .stream()
+                .map(entry -> entry.getKey()
+                        + FRUIT_TRANSACTION_SEPARATOR
+                        + entry.getValue())
+                .collect(Collectors.toList());
+        dailyReportList.add(TITLE_INDEX, CSV_REPORT_TITLE);
+        writerService.writeToFile(dailyReportList, fileName);
```
okuzan, line 30: One responsibility principle - report generator should not deal with writing to the file

src/main/java/core/basesyntax/service/impl/BalanceCalculatorServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+                         ));
+        for (Map.Entry<String, Integer> entry : balance.entrySet()) {
+            String key = entry.getKey();
+            Integer value = entry.getValue();
+            if (value < 0) {
+                throw new RuntimeException("Something wrong with balance for " + key
+                        + "\n It is negative.");
+            }
+        }
+        storageRecordsSetter.set(balance);
```
okuzan, line 46: don't operate with storage in bulk (delete the storageRecordsSetter class, you don't need it), parse transactions one by one. When IllgaleState would occur as a result of it (like negative balance, not enough to sell something), throw an exception then

okuzan (changes requested):
Good job renaming some of the classes, but please make sure to go through all the comments before requesting a review :)
Rename your handlers. Don't deal with storage contents in bulk, there should be only one Dao (with methods that returns all content for report generator. and the one that sets/puts quantity for specific fruit), or you could just make that map public and don't create dao at all, it's an abstraction of real DB one way or another.

PR #754, opened 2023-05-09
home-work/fruit-shop

andrii-subash:


src/main/java/core/basesyntax/service/impl/FruitTransactionServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class FruitTransactionServiceImpl implements FruitTransactionService {
+    private static final String TITLE = "type";
+    private static final String SEPARATOR = ",";
+    private static final int FRUIT_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+    private static final int OPERATION_INDEX = 0;
+    private static final int TRANSACTION_DATA_QUANTITY = 3;
+
+    private final ReaderService readerService = new ReaderServiceImpl();
```
olekskov, line 17: Consider using Dependency Injection for the ReaderService instead of directly initializing it. This will make the code more modular and easier to test.

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import java.io.IOException;
+
+public class WriterServiceImpl implements WriterService {
+    @Override
+    public void createReportFile(String fruitBalanceReport) {
+        File file = new File("src/main/resources/report.txt");
+        try (FileWriter fileWriter = new FileWriter(file)) {
+            fileWriter.write(fruitBalanceReport);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
olekskov, line 15: Provide a more specific error message when throwing a RuntimeException, e.g., 'Failed to write report to the file: ' + e.getMessage()'.

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,18 @@
...
+
+import core.basesyntax.service.WriterService;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class WriterServiceImpl implements WriterService {
+    @Override
+    public void createReportFile(String fruitBalanceReport) {
+        File file = new File("src/main/resources/report.txt");
```
olekskov, line 11: Replace the hardcoded file path with the output file name parameter passed to the 'createReportFile' method.

src/main/java/core/basesyntax/service/impl/FruitTransactionServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+        for (String information : fruitInfoFromFile) {
+            String[] infoAboutFruitTransaction = information.split(SEPARATOR);
+            if (infoAboutFruitTransaction[0].equals(TITLE)) {
+                continue;
+            }
+            validateTransactionInfo(information);
+            FruitTransaction fruitTransaction = new FruitTransaction(getOperation(
+                    infoAboutFruitTransaction[OPERATION_INDEX]),
+                            infoAboutFruitTransaction[FRUIT_INDEX],
+                            Integer.parseInt(infoAboutFruitTransaction[QUANTITY_INDEX]));
```
olekskov, line 32: Such structure is not readable. Please prepare some variable before with part of parsing or extract to separate method to improve readability

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+    private static final String SEPARATOR = ",";
+
+    @Override
+    public String createFruitBalanceReport(Map<String, Integer> fruitBalance) {
+        StringBuilder reportBuilder = new StringBuilder("fruit")
+                .append(SEPARATOR).append("quantity").append(System.lineSeparator());
+
+        for (Map.Entry<String, Integer> balance : fruitBalance.entrySet()) {
+            if (balance.getValue() < 0) {
+                throw new RuntimeException("Fruit quantity can`t be less than 0");
```
olekskov, line 16: IllegalAurgumentException is more precise and suitable

olekskov (changes requested):

src/main/java/core/basesyntax/service/impl/FruitTransactionServiceImpl.java
andrii-subash: I haven't studied Dependency Injection so I don't know how to use it properly. Please don't take this as a mistake.

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
andrii-subash: I сan`t replace RuntimeException with IllegalAccessException, because I need to write an error in the method signature through "throws" and make changes to the interface, which is not allowed by Dependency Inversion

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+import core.basesyntax.service.WriterService;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class WriterServiceImpl implements WriterService {
+    public static final String FILE_NAME = "src/main/resources/report.txt";
+
+    @Override
+    public void createReportFile(String fruitBalanceReport) {
```
kozhukhovsky, line 12:
```suggestion
    public void createReportFile(String fruitBalanceReport, String filePath) {

```

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+        File file = new File(FILE_NAME);
```
kozhukhovsky, line 13:
```suggestion
        File file = new File(filePath);

```

kozhukhovsky, line 13:
```suggestion
        File file = new File(filePath);

```

It meant that it was possible to pass the path to the method in which we use it, and it turns out that this class can only be used to write to one specific file.

kozhukhovsky (approved):

src/main/java/core/basesyntax/service/impl/FruitTransactionServiceImpl.java
kozhukhovsky: It meant that it is better not to initialize directly in the class. It is better to transfer all internal dependencies from the outside.

PR #755, opened 2023-05-11
first version

sZlishchev:


OlehPryshchepa (approved):
good job

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,35 @@
...
+import core.basesyntax.service.impl.DataProcessorImpl;
+import java.util.List;
+
+public class Application {
+    private static final String READ_FROM_FILE =
+            "src\\main\\java\\core\\basesyntax\\resourses\\ValidReadFromFile.csv";
+    private static final String NEGATIVE_QUANTITY_READ_FROM_FILE =
+            "src\\main\\java\\core\\basesyntax\\resourses\\NegativeQuantityReadFromFile.csv";
+    private static final String WRITE_TO_FILE =
+            "src\\main\\java\\core\\basesyntax\\resourses\\WriteToFile.csv";
```
sarakhmen, line 22: as described in common mistakes, please put input and output files in src/main/resources. Also consider making names more clear

src/main/java/core/basesyntax/db/DaoService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.db;
+
+import java.util.List;
+
+public interface DaoService {
```
sarakhmen, line 5:
```suggestion
public interface FruitDao {
```
I'd rather recommend naming it this way. The same for implementation

src/main/java/core/basesyntax/service/Calculator.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface Calculator {
```
sarakhmen, line 6: let's make the name of this interface and related impl more clear

src/main/java/core/basesyntax/service/DataProcessor.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface DataProcessor {
```
sarakhmen, line 6: the same issue here

src/main/java/core/basesyntax/strategy/OperationServiceStrategy.java
```diff
@@ -0,0 +1,18 @@
...
+public class OperationServiceStrategy {
+    public OperationService getOperationService(String operation) {
+        switch (operation) {
+            case "b" :
+            case "s" :
+            case "r" :
+                return new AddOperationService();
+            default:
+                return new SubtractOperationService();
+        }
```
sarakhmen, line 16: as described in common mistakes, let's avoid switch cases and use map with strategies in such cases

src/main/java/core/basesyntax/service/impl/SubtractOperationService.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.OperationService;
+
+public class SubtractOperationService implements OperationService {
```
sarakhmen, line 5: As for this homework we recommend creating a separate handler for each operation (4 in total) even though they are pretty much the same. These handlers should be named this way: PurchaseOperationHandler and so on. And implement OperationHandler interface. You can use different naming convention, but please don't name they as 'service'

src/main/java/core/basesyntax/service/impl/CsvFileWriterServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import core.basesyntax.service.CsvFileWriterService;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class CsvFileWriterServiceImpl implements CsvFileWriterService {
+    private static final String FRUIT = "fruit";
+    private static final String QUANTITY = "quantity";
+    private static final String COMA_SEPARATOR = ",";
+    private StringBuilder stringBuilder;
```
sarakhmen, line 13: redundant field

src/main/java/core/basesyntax/service/impl/CalculatorImpl.java
```diff
@@ -0,0 +1,34 @@
...
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.Calculator;
+import core.basesyntax.service.OperationService;
+import core.basesyntax.strategy.OperationServiceStrategy;
+import java.util.List;
+
+public class CalculatorImpl implements Calculator {
+    private static final String BANANA = "banana";
+    private static final String APPLE = "apple";
+    private OperationServiceStrategy operationServiceStrategy = new OperationServiceStrategy();
```
sarakhmen, line 12: let's get the instance of this class using a constructor instead of field-initialization

src/main/java/core/basesyntax/strategy/OperationServiceStrategy.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.service.OperationService;
+import core.basesyntax.service.impl.AddOperationService;
+import core.basesyntax.service.impl.SubtractOperationService;
+
+public class OperationServiceStrategy {
```
sarakhmen, line 7: let's avoid using 'service' here as well. Consider also creating an interface for this class

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,35 @@
...
+import core.basesyntax.service.impl.CalculatorImpl;
+import core.basesyntax.service.impl.CsvFileReaderServiceImpl;
+import core.basesyntax.service.impl.CsvFileWriterServiceImpl;
+import core.basesyntax.service.impl.DataProcessorImpl;
+import java.util.List;
+
+public class Application {
+    private static final String READ_FROM_FILE =
+            "src\\main\\java\\core\\basesyntax\\resourses\\ValidReadFromFile.csv";
+    private static final String NEGATIVE_QUANTITY_READ_FROM_FILE =
```
sarakhmen, line 19: NEGATIVE_QUANTITY_READ_FROM_FILE seems to be redundant field

sarakhmen (changes requested):
Let's improve your solution :)

sarakhmen (commented):
Also pay attention that some of your services have state and are not reusable

sarakhmen (changes requested):
Let's slightly improve your solution one more time ;)

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,33 @@
...
+    public static void main(String[] args) {
+        QuantityCalculatorService quantityCalculator = new QuantityCalculatorServiceImpl(
+                new OperationHandlerStrategyImpl());
+        FruitDao fruitDao = new FruitDaoImpl();
+        CsvFileWriterService csvFileWriterService = new CsvFileWriterServiceImpl();
+        List<FruitTransaction> fruitTransactions =
+                new TransactionParserServiceImpl()
+                        .getTransactions(new CsvFileReaderServiceImpl()
+                                .readFormFile(VALID_READ_FROM_FILE));
+        fruitDao.addDataToStorage(quantityCalculator.calculate(fruitTransactions));
```
maxline, line 30: `fruitDao.addDataToStorage()` is redundant. all data should be added to storage inside `quantityCalculator.calculate()`

src/main/java/core/basesyntax/Application.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.List;
+
+public class Application {
+    private static final String VALID_READ_FROM_FILE =
+            "src/main/resources/ValidReadFromFile.csv";
+    private static final String WRITE_TO_FILE =
+            "src/main/resources/WriteToFile.csv";
+
+    public static void main(String[] args) {
+        QuantityCalculatorService quantityCalculator = new QuantityCalculatorServiceImpl(
```
maxline, line 22: let's try to introduce variable before the place where it used and try to write the code that can be readed easily line by line like a book. See the example:
```
List<String> linesFormFile = new CsvFileReaderServiceImpl()
                .readFormFile(VALID_READ_FROM_FILE);

List<FruitTransaction> fruitTransactions = new TransactionParserServiceImpl()
                .getTransactions(linesFormFile);

new QuantityCalculatorServiceImpl(new OperationHandlerStrategyImpl())
                .calculate(fruitTransactions);
        
CsvFileWriterService csvFileWriterService = new CsvFileWriterServiceImpl();
csvFileWriterService.writeToFile(WRITE_TO_FILE);
```

src/main/java/core/basesyntax/service/impl/QuantityCalculatorServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private static final String BANANA = "banana";
+    private static final String APPLE = "apple";
+    private OperationHandlerStrategy operationServiceStrategy;
+    private OperationHandler operationService;
+
+    public QuantityCalculatorServiceImpl(OperationHandlerStrategy operationServiceStrategy) {
+        this.operationServiceStrategy = operationServiceStrategy;
+    }
+
+    public List<String> calculate(List<FruitTransaction> fruitTransactions) {
```
maxline, line 19: I would suggest refactor this method totaly. The idea of using strategy pattern is get the needed service from the `strategy map` without using if else. Repeat the topic of Strategy pattern please. 
The implementation shoul looks like:

```
public void calculate(List<FruitTransaction> fruitTransactions) {
    for (FruitTransaction fruitTransaction : fruitTransactions) {
            OperationHandler handler = operationServiceStrategy.getOperationService(fruitTransaction.getOperation());
            handler.doCalculation(fruitTransaction);
        }
```
Please pay attention the method should return nothing and method `doCalculation()` should store the result of calculations to the Storage.

src/main/java/core/basesyntax/strategy/OperationHandlerStrategy.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.strategy;
+
+public interface OperationHandlerStrategy {
+    public OperationHandler getOperationService(String operation);
```
maxline, line 4: Use specific object for operation instead of `String`
```suggestion
    OperationHandler getOperationService(FruitTransaction.Operation type);

```
remove public modifier from the interface.

src/main/java/core/basesyntax/strategy/impl/BalanceOperationHandler.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.strategy.impl;
+
+import core.basesyntax.strategy.OperationHandler;
+
+public class BalanceOperationHandler implements OperationHandler {
+    @Override
+    public int doCalculation(int totalQuantity, int quantity) {
+        return totalQuantity += quantity;
```
maxline, line 8: - This method should return nothing and should store the result of the operation to the `Storage` map.
- Please fix this suggestion in all  `*Handler` classes.
- Do not forget check the balance (e.g. negative balance) when operation processed and throw the runtime exceptions inside of `doCalculation()` methods

sarakhmen (approved):
Good job ;)

PR #766, opened 2023-05-22
Solid hm

MykytaKat:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,33 @@
...
+    private static final String INPUT_FILE
+            = "src/main/resources/input_file.csv";
+    private static final String REPORT_FILE
+            = "src/main/resources/report_file.csv";
+
+    public static void main(String[] args) {
+        final List<String> linesFromFile = new ReaderServiceImpl().readFromFile(INPUT_FILE);
+        final List<FruitTransaction> fruitTransactions =
+                new TransactionServiceImpl().parseTransactions(linesFromFile);
+        final Map<FruitTransaction.Operation, OperationHandler> operationHandlerMap
```
maxline, line 26: Please move this map declaration and initialisation inside of class OperationStrategy.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,33 @@
...
+                = new HashMap<>();
+        final OperationStrategy strategy = new OperationStrategy(operationHandlerMap);
```
maxline, line 28: why do we put an empty map via constructor?

maxline (approved):
Well done! Approved with one small comment. Please take a look for the future.

PR #767, opened 2023-05-22
Jv fruit shop solution

Ivan-Yuriv:


src/main/java/strategy/PurchaseHandler.java
```diff
@@ -0,0 +1,15 @@
...
+import db.Storage;
+import model.FruitTransaction;
+import service.OperationHandler;
+
+public class PurchaseHandler implements OperationHandler {
+    @Override
+    public void handle(FruitTransaction fruitTransaction) {
+        String fruitName = fruitTransaction.getFruit();
+        int purchasedQuantity = fruitTransaction.getQuantity();
+        int oldQuantity = Storage.fruits.get(fruitName);
```
aaananas, line 12: lets throw an exception if there is not enough fruits in the storage

src/main/java/service/impl/TransactionParserImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        fruitTransaction.setFruit(fields[FRUIT_NAME]);
+        fruitTransaction.setQuantity(Integer.parseInt(fields[QUANTITY]));
+        return fruitTransaction;
+    }
+
+    private Map<String, FruitTransaction.Operation> createMapOperations() {
+        return Arrays.stream(FruitTransaction.Operation.values())
+                .map(o -> new AbstractMap.SimpleEntry<>(o.getOperation(), o))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
```
aaananas, line 47: lets add a static method to Operation enum, ```getByCode(code)```

aaananas (changes requested):

src/main/java/service/impl/TransactionParserImpl.java
Ivan-Yuriv: Is it really needed? It will crash my code, because of my previous logic from FruitTransction class. I changed field "code" to "operation" there. My method getOperation() has the same logic.

src/main/java/model/FruitTransaction.java
```diff
@@ -0,0 +1,47 @@
...
+    public void setQuantity(int quantity) {
+        this.quantity = quantity;
+    }
+
+    public enum Operation {
+        BALANCE("b"),
+        SUPPLY("s"),
+        PURCHASE("p"),
+        RETURN("r");
+        private final String operation;
```
Rommelua, line 37: **operation** is an enum instance. **code** is a corresponding string.
```suggestion
        private final String code;
```

src/main/java/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+public class ReaderServiceImpl implements ReaderService {
+    @Override
+    public List<String> readDataFromFile(String filePath) {
+        List<String> dataFromFile = new ArrayList<>();
+        try {
+            dataFromFile = Files.readAllLines(Path.of(filePath));
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t read data from file" + filePath);
+        }
+        return dataFromFile;
```
Rommelua, line 19: image

```suggestion
        try {
            return Files.readAllLines(Path.of(filePath));
        } catch (IOException e) {
            throw new RuntimeException("Can`t read data from file" + filePath, e);
        }
```

src/main/java/service/impl/TransactionParserImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import model.FruitTransaction;
+import service.TransactionParser;
+
+public class TransactionParserImpl implements TransactionParser {
+    private static final int TYPE = 0;
```
Rommelua, line 13:
```suggestion
    private static final int OPERATION_INDEX = 0;
```

src/main/java/service/impl/TransactionParserImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final int FRUIT_NAME = 1;
```
Rommelua, line 14:
```suggestion
    private static final int FRUIT_NAME_INDEX = 1;
```

Rommelua (changes requested):

PR #776, opened 2023-05-29
Hw jv

marimarig:


src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Map;
+
+public interface Storage {
+    Map<String, Integer> storage = new HashMap<>();
+    void update(FruitTransaction fruitTransaction, int count);
+
+    int calculateAmount(FruitTransaction fruitTransaction);
+
+    List<String> getAll();
+}
```
chrisnyrka, line 15: According to the principles of SOLID (namely Single Responsibility Principle) -  class or interface must be responsible for one function. I think, such decision is not ok. It's better to implement the map in a class and not an interface, and transfer the methods to separate interfaces.

src/main/java/core/basesyntax/service/CsvFileWriterService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+import java.util.List;
+
+public interface CsvFileWriterService {
+    public void writeFile(List<String> lines, String fileName);
```
chrisnyrka, line 6: redundant access modifier

src/main/java/core/basesyntax/service/FormatTransformerService.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface FormatTransformerService {
+    List<FruitTransaction> formatData(List<String> data);
+
```
chrisnyrka, line 8: redundant line

chrisnyrka (approved):
Good job, just a few recommendations

src/main/java/core/basesyntax/db/Storage.java
marimarig: In principle, I agree with you, but I think that having too many different interfaces for such a task might lead to difficulties in test coverage. Let's wait for the mentor's opinion.

src/main/java/core/basesyntax/strategy/impl/Balance.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax.strategy.impl;
+
+import core.basesyntax.db.StorageImpl;
+import core.basesyntax.exception.InvalidValueExeption;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.strategy.OperationAnalysis;
+
+public class Balance implements OperationAnalysis {
```
maxline, line 8:
```suggestion
public class BalanceOperation implements OperationAnalysis {
```
please rename all other operations like in example above

src/main/java/core/basesyntax/db/StorageImpl.java
```diff
@@ -0,0 +1,25 @@
...
+        return Storage.storage.getOrDefault(fruitTransaction.getFruit(), 0);
+    }
+
+    @Override
+    public List<String> getAll() {
+        String joiningSymbol = ",";
+        return Storage.storage.entrySet().stream()
+                .map(entry -> entry.getKey() + joiningSymbol + entry.getValue())
+                .collect(Collectors.toList());
+    }
```
maxline, line 24: This code is about formatting. So I would recommend move it to the more appropriate place according Single Responsibility principle - to `ReportService`.
The class `ReportService` can take a `storage Map` as parameter and format data in needed way for reporting purposes.

maxline (changes requested):
Very good job! I left a several suggestions to make the solution even better.

maxline (approved):
Well done! Approved

PR #782, opened 2023-05-30
implemented all needed classes and interfaces

ykhmaruk:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,45 @@
...
+        ParserService parserService = new ParserServiceImpl();
+        OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlerMap);
+        ReportService reportService = new ReportServiceImpl();
+        WriterService writerService = new WriterServiceImpl();
+        List<String> fruits = readerService.readFromFile(READ_FROM_FILE);
+        List<FruitTransaction> parsedFruits = parserService.parseReadedData(fruits);
+        for (FruitTransaction fruitTransaction : parsedFruits) {
+            operationStrategy.getOperationStrategy(fruitTransaction.getOperation())
+                    .handle(fruitTransaction);
+        }
```
Tupadzaky, line 42: You should create separate service for this operation

Tupadzaky (changes requested):
looks good to me, except one thing. Check it out in comments

README.md
```diff
@@ -152,7 +152,7 @@ public class UserServiceImpl implements UserService {
 Don't forget that your code will need to be tested, so try to anticipate all invalid input data and handle it in advance.
 For example:
 1. Incorrect file path
-2. Incorrect data in the input file, for example, quantity less than zero or incorrect strategy
```
Tupadzaky, line 155: It's not a problem, but it seems that your IDE changed the text in Readme for some reason)

src/main/java/core/basesyntax/service/DataProcess.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.strategy.OperationStrategy;
+import java.util.List;
+
+public interface DataProcess {
+    void addDataToDB(List<FruitTransaction> fruitTransactions, OperationStrategy operationStrategy);
```
Tupadzaky, line 8: It's not a mistake, but I'd better move operationStrategy from the method parameter and put as a field on the class level. Since operationStrategy probably doesn't need to be changed on a specific instance it's better to just pass it into constructor, or inject it via Injector.
The problem that your approach may cause is that if addDataToDB method is used in a lot of places, you would need to pass operationStrategy every time you call it.

Tupadzaky (approved):
Looks good, but check out the comments

src/main/java/core/basesyntax/service/ParserService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface ParserService {
+    List<FruitTransaction> parseReadedData(List<String> infoFromFile);
```
rovein, line 7:
```suggestion
    List<FruitTransaction> parseData(List<String> rawData);
```

src/main/java/core/basesyntax/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ParserServiceImpl implements ParserService {
+    private static final String COMA_SEPARATOR = ",";
+    private static final int INDEX_OF_OPERATION = 0;
+    private static final int INDEX_OF_FRUIT = 1;
+    private static final int INDEX_OF_QUANTITY = 2;
```
rovein, line 14:
```suggestion
    private static final int OPERATION_INDEX = 0;
    private static final int FRUIT_INDEX = 1;
    private static final int QUANTITY_INDEX = 2;
```

src/main/java/core/basesyntax/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public List<FruitTransaction> parseReadedData(List<String> infoFromFile) {
+        return infoFromFile.stream()
+                .skip(1)
+                .map(this::getFromCsvRow)
```
rovein, line 20:
```suggestion
                .map(this::parseCsvRow)
```

src/main/java/core/basesyntax/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+        String fruit = fields[INDEX_OF_FRUIT];
+        int quantity = Integer.parseInt(fields[INDEX_OF_QUANTITY]);
+        return new FruitTransaction(operation, fruit, quantity);
+    }
+
+    private FruitTransaction.Operation getOperationFromString(String operationAbr) {
+        return Stream.of(FruitTransaction.Operation.values())
+                .filter(o -> o.getCode().equals(operationAbr))
+                .findAny()
+                .orElseThrow(() -> new NoSuchElementException("Invalid operation " + operationAbr));
```
rovein, line 36:
```suggestion
                .orElseThrow(() -> new IllegalArgumentException("Invalid operation " + operationAbr));
```

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+    @Override
+    public List<String> readFromFile(String path) {
+        List<String> infoFromFile;
+        try {
+            infoFromFile = Files.readAllLines(Path.of(path));
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t get data from file" + path, e);
+        }
+        return infoFromFile;
+    }
```
rovein, line 19:
```suggestion
    public List<String> readFromFile(String path) {
        try {
            return Files.readAllLines(Path.of(path));
        } catch (IOException e) {
            throw new RuntimeException("Can`t get data from file" + path, e);
        }
    }
```

You don't need to return anything if there was an exception while reading.

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.List;
+
+public class ReaderServiceImpl implements ReaderService {
+    @Override
+    public List<String> readFromFile(String path) {
+        List<String> infoFromFile;
+        try {
+            infoFromFile = Files.readAllLines(Path.of(path));
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t get data from file" + path, e);
```
rovein, line 16:
```suggestion
            throw new RuntimeException("Can`t get data from file " + path, e);
```

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.ReportService;
+import core.basesyntax.storage.Storage;
+import java.util.Map;
+
+public class ReportServiceImpl implements ReportService {
+    public static final String FIRST_LINE_OF_REPORT = "fruit,quantity";
```
rovein, line 8:
```suggestion
    public static final String REPORT_HEADER = "fruit,quantity";
```

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,22 @@
...
+    public static final String COMMA_SEPARATOR = ",";
+
+    @Override
+    public String makeReport() {
+        StringBuilder builder = new StringBuilder(FIRST_LINE_OF_REPORT + System.lineSeparator());
+        for (Map.Entry<String, Integer> data : Storage.fruitStorage.entrySet()) {
+            builder.append(data.getKey());
+            builder.append(COMMA_SEPARATOR);
+            builder.append(data.getValue());
+            builder.append(System.lineSeparator());
```
rovein, line 18: Alternatively, you can use chain calls:
```suggestion
builder.append()
    .append()...
```

src/main/java/core/basesyntax/storage/Storage.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.storage;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    public static final Map<String, Integer> fruitStorage = new HashMap<>();
```
rovein, line 7:
```suggestion
    public static final Map<String, Integer> FRUIT_STORAGE = new HashMap<>();
```

src/main/java/core/basesyntax/strategy/impl/BalanceOperationHandler.java
```diff
@@ -0,0 +1,18 @@
...
+public class BalanceOperationHandler implements OperationHandler {
+    @Override
+    public void handle(FruitTransaction fruitTransaction) {
+        int quantity = fruitTransaction.getQuantity();
+        String fruit = fruitTransaction.getFruit();
+        if (quantity > 0) {
+            Storage.fruitStorage.put(fruit, quantity);
+        } else {
+            throw new RuntimeException("Invalid quantity of fruit" + quantity);
+        }
```
rovein, line 16: Let's create less `if/else`:
```suggestion
        if (quantity <= 0) {
            throw new RuntimeException("Invalid quantity of fruit" + quantity);
        } 
        Stroage....
```

Applicable for all other places.

rovein (changes requested):
Good job! But let's improve your solution ;)

PR #803, opened 2023-05-31
Hw jv fruit shop

yuriy-kononchuk:


yuriy-kononchuk:
Thanks for review, that is an interesting approach! Perhaps, I overengineered the solution. It's because I wanted to make my solution more useful and practice, including chosen structure and model of storage. But it seems the implementation and understanding of SOLID principles from my side is correct.
So could you please approve PR, so that I will move to the next task and use all these recommendations for the next challenge!

PR #811, opened 2023-06-09
added solution

Serhii-Nosulia:


src/main/java/core/basesyntax/service/impl/ReadServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+    @Override
+    public List<String> readInputData(String inputFilePath) {
+        List<String> inputData;
+        try {
+            inputData = Files.readAllLines(Path.of(inputFilePath));
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t read file: " + inputFilePath);
+        }
+        return inputData;
+    }
```
Ivan95kos, line 19:
```suggestion
    @Override
    public List<String> readInputData(String inputFilePath) {
        try {
            return Files.readAllLines(Path.of(inputFilePath));
        } catch (IOException e) {
            throw new RuntimeException("Can`t read file: " + inputFilePath);
        }
    }
```

src/main/java/core/basesyntax/service/impl/WriteServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+            "src/main/java/core/basesyntax/resources/report.scv";
+
+    @Override
+    public void writeReport(String report) {
+        try {
+            Files.writeString(Path.of(REPORT_FILE_PATH), report);
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t write to file: " + REPORT_FILE_PATH);
+        }
+    }
```
Ivan95kos, line 19: It is better to get the path as parameter
```suggestion
    @Override
    public void writeReport(String filePath, String report) {
        try {
            Files.writeString(Path.of(filePath), report);
        } catch (IOException e) {
            throw new RuntimeException("Can`t write to file: " + filePath);
        }
    }
```

Ivan95kos (changes requested):
Great job overall!

aaananas (approved):

PR #812, opened 2023-06-11
Implemented bench of classes with appropriate methods in them, so it'…

vaniii1:
…s possible to count the quantity of each fruit and write report to another file


src/main/java/core/basesyntax/model/OperationFinderImpl.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax.model;
```
olekskov, line 1: Model is class that holds some properties.
This class looks more like util or service.
If you make method name as parseOperationFromInput - then it will perfectly fit your ParsingService and may be moved there

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    public void editMapReportDependingOnOperation(List<FruitTransaction> fruits) {
+        Map<FruitTransaction.Operation, OperationHandler> operationHandlerMap = new HashMap<>();
+        operationHandlerMap.put(FruitTransaction.Operation.BALANCE,
+                new BalanceOperationHandler());
+        operationHandlerMap.put(FruitTransaction.Operation.PURCHASE,
+                new PurchaseOperationHandler());
+        operationHandlerMap.put(FruitTransaction.Operation.RETURN,
+                new ReturnOperationHandler());
+        operationHandlerMap.put(FruitTransaction.Operation.SUPPLY,
+                new SupplyOperationHandler());
```
olekskov, line 43: Please remove hardcoded map upgrade, move to class field and accept map as argument in constructor.
Create proper map in Main class and pass to constructor of service.
With current implementation you recreate map each time method is called

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        operationStrategy = new OperationStrategyImpl(operationHandlerMap);
```
olekskov, line 44: Please don't use initialization of services we depend on with new.
You will not be able to test and maintain your application properly
All services we need in our class should be class fields and for the moment should be passed via constructor

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        for (FruitTransaction fruit : fruits) {
+            operationStrategy.get(fruit.getOperation()).getNewValueDependingOnOperation(fruit);
+        }
```
olekskov, line 47: This part of code has nothing connected with name of method which it's placed in.
Please remember about single responsibility principle and make your method do only one thing it supposed to do

olekskov (changes requested):
Good job overall, but require some improvements to proceed further

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+            builder.append(map.getKey());
+            builder.append(SEPARATOR);
+            builder.append(map.getValue());
+            builder.append(System.lineSeparator());
+        }
+        return builder.toString();
+    }
+
+    @Override
+    public void loopThroughOperationsOfList(List<FruitTransaction> fruitTransactionList) {
```
olekskov, line 32:
```suggestion
    public void applyTransactionsToStorage(List<FruitTransaction> fruitTransactionList) {
```

olekskov (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,51 @@
...
+        List<String> transactions = csvFileReaderService.readDataFromTheFile(FROM_FILE_NAME);
+        ParsingService parsingService = new ParcingCsvServiceImpl();
+        List<FruitTransaction> fruits = parsingService.parsingDataFromListOfStrings(transactions);
+
+        ReportService reportService = new ReportServiceImpl(
+                new OperationStrategyImpl(operationHandlerMap));
+        reportService.applyTransactionsToStorage(fruits);
+        String report = reportService.createFinalReport(Storage.fruits);
+        WriterService writerService = new WriterServiceImpl();
+        writerService.writingDataToFile(TO_FILE_NAME, report);
```
olekskov, line 49: Usually we use present simple in method naming
```suggestion
        writerService.writeDataToFile(TO_FILE_NAME, report);
```

src/main/java/core/basesyntax/operation/OperationHandler.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.operation;
+
+import core.basesyntax.model.FruitTransaction;
+
+public interface OperationHandler {
+    void getNewValueDependingOnOperation(FruitTransaction fruitTransaction);
```
olekskov, line 6: void method shouldn't contain 'get' in naming as it returns nothing
```suggestion
    void applyTransactionToStorage(FruitTransaction fruitTransaction);
```

olekskov (approved):
Good job, approved.
Please apply changes in comments below to code before proceeding with testing task for this project

PR #813, opened 2023-06-13
solution

VictoriaMrkvch:


src/main/java/core/basesyntax/dao/FruitDaoImpl.java
```diff
@@ -0,0 +1,34 @@
+package core.basesyntax.dao;
+
+import core.basesyntax.db.Storage;
+import java.util.Map;
+
+public class FruitDaoImpl implements FruitDao {
```
olekskov, line 6: In order to simplify solution and further testing I'd recommend to remove DAO layer from solution and allow handlers to have direct access to storage and actually change quantity, not just calculate it

src/main/java/core/basesyntax/service/impl/FruitShopServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+        this.fileService = fileService;
+        this.fruitTransactionParseService = fruitTransactionParseService;
+        this.operationStrategy = operationStrategy;
+    }
+
+    @Override
+    public void generateRreport(String pathFrom, String pathTo) {
+        List<String> records = fileService.readFromFile(pathFrom);
+        List<FruitTransaction> transactions =
+                fruitTransactionParseService.parseFruitTransaction(records);
```
olekskov, line 31: This part is better to be moved to main method

src/main/java/core/basesyntax/service/impl/FruitShopServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+            int newQuantity = operationStrategy.get(transaction.getOperation())
+                    .execudeOperation(currentQuantity, transaction.getQuantity());
+            if (newQuantity < 0) {
+                throw new RuntimeException("Quantity can't be less than zero");
+            } else {
+                fruitDao.add(transaction.getFruit(), newQuantity);
+            }
+        }
+
+        fileService.writeToFile(pathTo, fruitDao.getDataForReportFile());
```
olekskov, line 47: This part also will be fine in main method

src/main/java/core/basesyntax/service/impl/FruitShopServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+
+    }
```
olekskov, line 49: Please move here last method from dao and it will be actually generateReport()

olekskov (changes requested):
Hi, Victoria. Really nice job overall.
My requests below are just aimed to simplify solution and further testing (next task)

kozhukhovsky (approved):

PR #819, opened 2023-06-25
Hometask JV Fruit SHOP completed

MarynaKushniruk:


src/main/java/core/basesyntax/service/impl/FilesParserImpl.java
```diff
@@ -0,0 +1,28 @@
...
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.FilesParser;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FilesParserImpl implements FilesParser {
+    private static final String SEPARATOR = ",";
+    private static final int INDEX_ZERO = 0;
+    private static final int INDEX_ONE = 1;
+    private static final int INDEX_TWO = 2;
```
olekskov, line 12: Constants names should describe meaning but not value itself
```suggestion
    private static final int OPERATION_TYPE_INDEX = 0;
    private static final int FRUIT_NAME_INDEX = 1;
    private static final int FRUIT_QUANTITY_INDEX = 2;
```

src/main/java/core/basesyntax/strategy/OperationHandler.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.strategy;
+
+public interface OperationHandler {
+    int executeOperation(int currentQuantity, int quantityFromTransaction);
```
olekskov, line 4: Overall such approach is ok, I cannot say that this is mistake. 
But for current task and in scope of further testing I'd recommend to rework handlers to put new result directly to storage (and obtain current quantity from there), not just calculate result.
With such approach you will be able to remove FruitsDao totally and make logic more simple in TransactionHandlerImpl.

```suggestion
    void executeOperation(int quantityFromTransaction);
```

olekskov (changes requested):
Good job overall, but I'd recommend to simplify solution according to comment below:

kozhukhovsky (approved):
Good job!

PR #829, opened 2023-07-05
added fruit shop

krismva:


src/main/java/core/shop/model/ActivityType.java
```diff
@@ -0,0 +1,27 @@
+package core.shop.model;
+
+public enum ActivityType {
```
kozhukhovsky, line 3:
```suggestion
public enum OperationType {
```

src/main/java/core/shop/model/FruitRecord.java
```diff
@@ -0,0 +1,50 @@
+package core.shop.model;
+
+public class FruitRecord {
```
kozhukhovsky, line 3:
```suggestion
public class FruitTransaction {
```

src/main/java/core/shop/service/impl/FileServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+import core.shop.service.FileService;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class FileServiceImpl implements FileService {
```
kozhukhovsky, line 11: It needs to be divided into two interfaces.
A separate class with reading from a file, a separate class with writing to a file.

src/main/java/core/shop/service/impl/ParseListServiceImpl.java
```diff
@@ -0,0 +1,31 @@
+package core.shop.service.impl;
+
+import core.shop.model.ActivityType;
+import core.shop.model.FruitRecord;
+import core.shop.service.ParseListService;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ParseListServiceImpl implements ParseListService {
```
kozhukhovsky, line 9:
```suggestion
public class ParseFruitTransactionServiceImpl implements ParseListService {
```

src/main/java/core/shop/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+        Set<String> fruitsSet = getFruits(fruitRecordList);
+
+        for (String fruit : fruitsSet) {
+            int totalQuantity = calculateQuantity.calculateQuantity(fruitRecordList, fruit);
+            reportBuilder.append(fruit)
+                    .append(COMA_SEPARATOR)
+                    .append(totalQuantity)
+                    .append(System.lineSeparator());
+        }
+        fileService.write(REPORT_FILE, reportBuilder.toString());
```
kozhukhovsky, line 34: You just need to create a report. And return a line with a report. You need to write it to a file in another place, for example, in Main in this case. You don't need to hard-wire a class to write to a file. We may need a report without writing to a file.

src/main/java/core/shop/service/impl/CalculateQuantityServiceImpl.java
```diff
@@ -0,0 +1,26 @@
+package core.shop.service.impl;
+
+import core.shop.model.ActivityType;
+import core.shop.model.FruitRecord;
+import core.shop.service.CalculateQuantityService;
+import java.util.List;
+
+public class CalculateQuantityServiceImpl implements CalculateQuantityService {
+    @Override
+    public int calculateQuantity(List<FruitRecord> fruits, String fruitName) {
```
kozhukhovsky, line 10: It is necessary to divide this logic into several classes. Create your own handler for each operation, for example BalanceHandler. It will have one handle(Operation) method and will do the necessary actions.

kozhukhovsky: It will also be necessary to create an OperationStrategy class, which, depending on the type of operation, will return the necessary handler for us to process

kozhukhovsky: It is worth creating a Storage class, where the Map will be stored, to which fruits will first be added in accordance with the operations. And then a report is generated from it.

kozhukhovsky (changes requested):

src/main/java/core/shop/service/impl/FileWriteServiceImpl.java
```diff
@@ -0,0 +1,17 @@
...
+
+import core.shop.service.FileWriteService;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class FileWriteServiceImpl implements FileWriteService {
+    @Override
+    public void write(String fileName, String info) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName, true))) {
```
okuzan, line 11: why true here?, no need to append

okuzan (changes requested):

src/main/java/core/shop/service/impl/ParseFruitTransactionServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class ParseFruitTransactionServiceImpl implements ParseFruitTransactionService {
+    private static final int ACTIVITY_TYPE_INDEX = 0;
+    private static final int FRUIT_NAME_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+    private static final String COMA_SPLITERATOR = ",";
+
+    @Override
+    public List<FruitTransaction> getFruitOperations(List<String> list) {
+        List<String> withoutFirstLine = list.subList(1, list.size());
```
Ivan95kos, line 17: use constants instead of numbers

Ivan95kos (approved):

PR #830, opened 2023-07-08
hm done

NSashchenko:


src/main/java/core/basesyntax/test/Test.java
```diff
@@ -0,0 +1,48 @@
+package core.basesyntax.test;
```
olekskov, line 1: It's better to apply below change before proceeding to next task in order not to make confusion with tests that will be added
```suggestion
package core.basesyntax;

```

src/main/java/core/basesyntax/test/Test.java
```diff
@@ -0,0 +1,48 @@
...
+import core.basesyntax.service.impl.ParserServiceImpl;
+import core.basesyntax.service.impl.ReaderServiceImpl;
+import core.basesyntax.service.impl.ReportGeneratorServiceImpl;
+import core.basesyntax.service.impl.WriterServiceImpl;
+import core.basesyntax.service.strategy.OperationStrategyImpl;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Test {
```
olekskov, line 24:
```suggestion
public class Main {

```

olekskov (approved):
Nice solution, approved

src/main/java/core/basesyntax/service/FruitShopService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface FruitShopService {
+    void processOfOperations(List<FruitTransaction> fruitTransactions);
```
nowickimj, line 7: minor: `processOfOperations` sounds more like a noun than a verb to me. `processTransactions` sounds more descriptive.

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    public static final Map<String, Integer> fruits = new HashMap<>();
```
nowickimj, line 7: I see that you wanted to create a globally accessible storage service, but there are few things wrong about this approach. In case of dynamically changing internal state I'd recommend to follow the Singleton pattern.
1. Make constructors private - no one should be able to instantiate the singleton.
2. Store the only class instance internally as the private static field (`T instance`). You should store the state as a non-static private field - provide necessary accessors and mutators (e.g. T.addFruit(...). 
3. Create `public static T getInstance()` method that provides he internally stored instance:
```java
public static T getInstance() {
  if(instance == null) {
    instance = new T();
  }
  return instance;
}
```

src/main/java/core/basesyntax/service/handler/PurchaseHandler.java
```diff
@@ -0,0 +1,16 @@
...
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.FruitTransaction;
+
+public class PurchaseHandler implements OperationHandler {
+    @Override
+    public void handleTransaction(FruitTransaction transaction) {
+        int result = Storage.fruits.get(transaction.getFruit()) - transaction.getQuantity();
+        if (result >= 0) {
+            Storage.fruits.put(transaction.getFruit(), result);
```
nowickimj, line 11: This seems like you're trying to alter the constant which is illegal. After applying the singleton pattern to `Storage` class, you should be able to access the collection with: `Storage.getInstance().addFruit(...)`

nowickimj (changes requested):

PR #838, opened 2023-07-20
java-fruit-shop-solution

knyazkyylubomir:
created FruitTransaction model for containing converted data from file;
created CsvFileReaderService which has readData() method for reading data from file; 
created ConvertService which has convertData() method for converting list of strings to list of FruitTransaction object; 
created StoreService which has getOperationsAmount() for counting sum of each operation of a fruit and getFruitList() for containing unique list of fruits; 
created ReportService which has getReport() for getting string of inputs lists of fruits and their amounts; 
created CsvFileWriterService which has writeData() for writing input string to file

### DIFF
--- a/src/main/java/core/basesyntax/model/Operation.java
+++ b/src/main/java/core/basesyntax/model/Operation.java
@@ -1,8 +1,25 @@
 package core.basesyntax.model;
 
+import java.util.Arrays;
+
 public enum Operation {
-        BALANCE,
-        SUPPLY,
-        PURCHASE,
-        RETURN;
+    BALANCE("b"),
+    SUPPLY("s"),
+    PURCHASE("p"),
+    RETURN("r");
+
+    private static final String INCORRECT_DATA_EXCEPTION
+            = "Incorrect operation type! Insert correct type to input file";
+    private String code;
+
+    Operation(String code) {
+        this.code = code;
+    }
+
+    public static Operation getOperationType(String type) {
+        return Arrays.stream(values())
+                .filter(operation -> operation.code.equals(type))
+                .findFirst()
+                .orElseThrow(() -> new RuntimeException(INCORRECT_DATA_EXCEPTION));
+    }
 }
--- a/src/main/java/core/basesyntax/service/impl/ConvertServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/impl/ConvertServiceImpl.java
@@ -3,14 +3,10 @@ package core.basesyntax.service.impl;
 import core.basesyntax.model.FruitTransaction;
 import core.basesyntax.model.Operation;
 import core.basesyntax.service.impl.service.ConvertService;
-import core.basesyntax.strategy.OperationStrategy;
 import java.util.List;
 import java.util.stream.Collectors;
 
 public class ConvertServiceImpl implements ConvertService {
-    private static final String INCORRECT_DATA_EXCEPTION
-            = "Incorrect operation type or file contains empty line!"
-            + " Insert correct type to input file or delete empty line";
     private static final String LESS_THAN_ZERO_EXCEPTION
             = "Quantity cannot be negative! Insert correct values to input file";
     private static final String COMA_DELIMITER = ",";
@@ -20,11 +16,6 @@ public class ConvertServiceImpl implements ConvertService {
     private static final String COLUMN_NAME = "type";
     private static final String INVALID_INPUT_PARAMETER
             = "Invalid input parameter in convertData()";
-    private OperationStrategy operationStrategy;
-
-    public ConvertServiceImpl(OperationStrategy operationStrategy) {
-        this.operationStrategy = operationStrategy;
-    }
 
     @Override
     public List<FruitTransaction> convertData(List<String> strings) {
@@ -32,20 +23,17 @@ public class ConvertServiceImpl implements ConvertService {
             throw new RuntimeException(INVALID_INPUT_PARAMETER);
         }
         return strings.stream()
-                .filter(line -> !line.startsWith(COLUMN_NAME))
+                .filter(line -> !line.startsWith(COLUMN_NAME) && !line.isEmpty())
                 .map(this::getFruitTransactionObject)
                 .collect(Collectors.toList());
     }
 
     private FruitTransaction getFruitTransactionObject(String line) {
         String[] row = line.split(COMA_DELIMITER);
-        if (operationStrategy.get(row[TYPE_INDEX]) == null) {
-            throw new RuntimeException(INCORRECT_DATA_EXCEPTION);
-        }
         if (Integer.parseInt(row[QUANTITY_INDEX]) < 0) {
             throw new RuntimeException(LESS_THAN_ZERO_EXCEPTION);
         }
-        Operation operation = operationStrategy.get(row[TYPE_INDEX]).getOperationType();
+        Operation operation = Operation.getOperationType(row[TYPE_INDEX]);
         String fruit = row[FRUIT_INDEX];
         int quantity = Integer.parseInt(row[QUANTITY_INDEX]);
         return new FruitTransaction(operation, fruit, quantity);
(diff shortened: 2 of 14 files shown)


src/main/java/core/basesyntax/service/impl/CsvFileWriterServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.IOException;
+
+public class CsvFileWriterServiceImpl implements CsvFileWriterService {
+    private static final String FILE_PATH = "src/main/resources/report.csv";
+    private static final String EXCEPTION_INFO = "Cannot write file";
+    private static final String FIRST_FILE_ROW = "fruit,quantity";
+    private static final String INVALID_INPUT_PARAMETER = "Invalid input parameter in writeData()";
+
+    @Override
+    public void writeData(String report) {
```
Ivan95kos, line 15: Better to get the path as a parameter
```suggestion
    public void writeData(String report, String outputPath) {
```

src/main/java/core/basesyntax/service/impl/service/CsvFileWriterService.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.service.impl.service;
+
+public interface CsvFileWriterService {
```
Ivan95kos, line 3:
```suggestion
public interface WriterService {
```

src/main/java/core/basesyntax/strategy/BalanceActivityTypeHandler.java
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    @Override
+    public int getSumOfOperation(int quantity) {
+        if (quantity < 0) {
+            throw new RuntimeException("Fruit quantity cannot be negative");
+        }
+        return quantity;
+    }
+}
```
Ivan95kos, line 18: You should not return the value, do the operation in this method
```suggestion
public class BalanceOperationHandlerImpl implements OperationHandler {
    @Override
    public void handle(FruitTransaction fruitTransaction) {
        FruitStorage.FRUITS.put(fruitTransaction.getFruit(), fruitTransaction.getQuantity());
    }
}
```

src/main/java/core/basesyntax/strategy/ActivityTypeHandler.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.model.FruitTransaction;
+
+public interface ActivityTypeHandler {
+    FruitTransaction.Operation getActivityType();
+
+    int getSumOfOperation(int quantity);
+}
```
Ivan95kos, line 9: try to rewrite with these interfaces:
```
public interface OperationStrategy {
    OperationHandler get(FruitTransaction.Operation operation);
}
```
```
    void handle(FruitTransaction transaction);
}
```

Ivan95kos (changes requested):
The project should have packages (service, db, model, strategy...)
package `service` should have: FruitShopService, ParserService, ReaderService, ReportService, WriterService 
package `db` should have: Storage 
package `model` should have: FruitTransaction
package `strategy` should have: OperationStrategy , OperationHandler
all implementation should be to move to package impl
this will be a good code decomposition.

src/main/java/core/basesyntax/service/impl/ConvertServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+                .filter(line -> !line.startsWith(COLUMN_NAME))
+                .map(this::getFruitTransactionObject)
+                .collect(Collectors.toList());
+    }
+
+    private FruitTransaction getFruitTransactionObject(String line) {
+        String[] row = line.split(COMA_DELIMITER);
+        if (operationStrategy.get(row[TYPE_INDEX]) == null) {
+            throw new RuntimeException(INCORRECT_DATA_EXCEPTION);
+        }
```
olekskov, line 44: I don't like idea to create operation handler in order to check if operation is valid.
Please move this logic to ENUM where it will be natural

src/main/java/core/basesyntax/service/impl/ConvertServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+        if (Integer.parseInt(row[QUANTITY_INDEX]) < 0) {
+            throw new RuntimeException(LESS_THAN_ZERO_EXCEPTION);
+        }
+        Operation operation = operationStrategy.get(row[TYPE_INDEX]).getOperationType();
```
olekskov, line 48: same here

src/main/java/core/basesyntax/strategy/handlers/OperationHandler.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.strategy.handlers;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Operation;
+
+public interface OperationHandler {
+    void handle(FruitTransaction transaction);
+
+    Operation getOperationType();
```
olekskov, line 9: this method is not suitable for handlers and should be moved to enum

src/main/java/core/basesyntax/strategy/OperationStrategy.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.model.Operation;
+import core.basesyntax.strategy.handlers.OperationHandler;
+
+public interface OperationStrategy {
+    OperationHandler get(Operation operation);
+
+    OperationHandler get(String type);
```
olekskov, line 9: this method is not required here

src/main/java/core/basesyntax/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    {
+        operationHandlersMap.put(Operation.BALANCE, new BalanceOperationHandler());
+        operationHandlersMap.put(Operation.SUPPLY, new SupplyOperationHandler());
+        operationHandlersMap.put(Operation.PURCHASE, new PurchaseOperationHandler());
+        operationHandlersMap.put(Operation.RETURN, new ReturnOperationHandler());
+        typeHandlersMap.put("b", new BalanceOperationHandler());
+        typeHandlersMap.put("s", new SupplyOperationHandler());
+        typeHandlersMap.put("p", new PurchaseOperationHandler());
+        typeHandlersMap.put("r", new ReturnOperationHandler());
+    }
```
olekskov, line 25: Please initialize map in Main method and pass as argument
```suggestion
    private final Map<Operation, OperationHandler> operationHandlersMap;

    public OperationStrategyImpl(Map<Operation, OperationHandler> operationHandlersMap) {
        this. operationHandlersMap = operationHandlersMap;
    }
```

src/main/java/core/basesyntax/model/Operation.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.model;
+
+public enum Operation {
+        BALANCE,
+        SUPPLY,
+        PURCHASE,
+        RETURN;
```
olekskov, line 7: Should be something like this
```suggestion
    BALANCE("b"),
    SUPPLY("s"),
    PURCHASE("p"),
    RETURN("r");

    public final String shortValue;

    Operation(String shortValue) {
        this.shortValue = shortValue;
    }

    public static Operation byShortValue(String shortValue) {
        for (Operation operation : values()) {
            if (operation.shortValue.equals(shortValue)) {
                return operation;
            }
        }
        return null;
    }
```

olekskov (changes requested):
Good job in general but some logic should be moved to avoid inappropriate usage of strategy and handlers

src/main/java/core/basesyntax/service/impl/ConvertServiceImpl.java
knyazkyylubomir: I thought if we'll need to create a new operation so a handler for this is suitable

kozhukhovsky (approved):
Good job!

PR #850, opened 2023-07-29
created FruitTransaction model, created file data_shop; created servi…

misha-levko:
…ce classes for Balance, Purchase, Return, Supply impl; created classes for reading, writing and report data for FruitShop


src/main/java/core/basesyntax/model/FruitsShopServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    @Override
+    public Map<String, Integer> finalReport(List<FruitTransaction> parsed) {
+        if (parsed == null) {
+            throw new RuntimeException("Value can't be null");
+        }
+        parsed.forEach(t -> operationStrategy.get(t.getOperation()).operate(t));
+        return Storage.getMap();
+    }
```
aaananas, line 22:
```suggestion
    @Override
    public void process(List<FruitTransaction> parsed) {
        if (parsed == null) {
            throw new RuntimeException("Value can't be null");
        }
        parsed.forEach(t -> operationStrategy.get(t.getOperation()).operate(t));
    }
```
in report service just access storage

aaananas (changes requested):

aaananas (approved):

PR #853, opened 2023-07-31
doned solution - my variant. by my tests in main - all work correctly

DaniilNepelev:


src/main/java/core/service/impl/Strategy.java
```diff
@@ -0,0 +1,5 @@
+package core.service.impl;
+
+public interface Strategy {
```
satori11111, line 3: Strategy used not correct in this case, Strategy can take parameter Operation and return Handler

src/main/java/core/service/OperationData.java
```diff
@@ -0,0 +1,34 @@
+package core.service;
+
+public class OperationData {
+    private String operationType;
+    private String product;
+    private int number;
```
satori11111, line 6: better name quantity

src/main/java/core/service/OperationData.java
```diff
@@ -0,0 +1,34 @@
+package core.service;
+
+public class OperationData {
+    private String operationType;
```
satori11111, line 4: Must be Enum

src/main/java/core/service/FruitStore.java
```diff
@@ -0,0 +1,62 @@
...
+                case "s":
+                case "r":
+                    newQuantity = currentQuantity + number;
+                    break;
+                case "p":
+                    newQuantity = currentQuantity - number;
+                    break;
+                default:
+                    break;
+            }
```
satori11111, line 34: Do not use switch it is violation of Open Closed principle, use enum give them code value and check in for loop their code like this: 
for (Operation operation : Operation.values()) {
                if (operation.code.equals(operationLetter)) {
                    return operation;
                }
            }

src/main/java/core/service/DataSpliterator.java
```diff
@@ -0,0 +1,30 @@
...
+        for (String line : lines) {
+            String[] parts = line.split(splitRegex);
+            if (parts.length != 3) {
+                continue;
+            }
+            try {
+                int quantity = Integer.parseInt(parts[2].trim());
+                result.add(new OperationData(parts[0], parts[1], quantity));
+            } catch (NumberFormatException e) {
+                System.err.println("Invalid quantity value: " + parts[2]);
```
satori11111, line 24: Add constants to 0,1,2,3 not clear why you use this numbers

src/main/java/core/service/DataSpliterator.java
```diff
@@ -0,0 +1,30 @@
+package core.service;
+
+import core.service.impl.Spliterator;
+import java.util.ArrayList;
+import java.util.List;
+
+public class DataSpliterator implements Spliterator<OperationData> {
+    private static final String splitRegex = ",";
```
satori11111, line 8: better name COMMA ir just uppercase your name

satori11111 (changes requested):

satori11111:
also create some storage like in task registration validation(you can use HashMap or other structures) remember your solution need to have this architecture
image

src/main/java/core/service/FruitStore.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class FruitStore {
+    private Map<OperationType, OperationHandler> operationHandlers;
+
+    public FruitStore() {
+        operationHandlers = new HashMap<>();
+        operationHandlers.put(OperationType.B, new BalanceOperationHandler());
+        operationHandlers.put(OperationType.S, new SupplyOperationHandler());
+        operationHandlers.put(OperationType.R, new ReturnOperationHandler());
+        operationHandlers.put(OperationType.P, new PurchaseOperationHandler());
```
satori11111, line 21: move this hashMap to main method and give it to constructor of your strategy

src/main/java/core/strategy/ReportFactory.java
```diff
@@ -0,0 +1,24 @@
...
+    private static final FruitStore fruitStore = new FruitStore();
+    private static final ReportWriter reportWriter = new ReportWriter();
+
+    @Override
+    public void start(String fromPath, String toPath) {
+        String data = readerFile.read(fromPath);
+        List<OperationData> splitData = dataSpliterator.splitData(data);
+        List<OperationData> report = fruitStore.processOperations(splitData);
+        String reportInString = fruitStore.convertListToString(report);
+        reportWriter.createReport(toPath, reportInString);
```
satori11111, line 22: i think you can do it in main method, class which implements strategy should have method that return OperationHandler, method start can be in main

satori11111 (changes requested):
Good job!! But you still did not create Storage and save report data to storage

satori11111 (approved):
Great! But please move Storage.java to db folder

AntonZhdanov (approved):

src/main/java/core/Main.java
```diff
@@ -0,0 +1,13 @@
+package core;
+
+import core.strategy.ReportFactory;
+
+public class Main {
+
+    public static void main(String[] args) {
+        ReportFactory reportFactory = new ReportFactory();
+        String filePath = "src/main/java/core/db/dataBase.csv";
+        String pathTo = "test333";
```
nacenik, line 10: bad value for files 
is it database or input?? 
is it some unknown file "test333" or report?? 
image

src/main/java/core/dao/StorageDao.java
```diff
@@ -0,0 +1,7 @@
+package core.dao;
+
+import java.util.Map;
+
+public interface StorageDao {
+    Map<String,Integer> getAll();
+}
```
nacenik, line 7: if you don't know what dao do, please, don't use it. Make your storage `public static final` and use it everywhere

src/main/java/core/db/dailyReport/test33
```diff
@@ -0,0 +1,3 @@
+fruit,quantity
+banana,152
+apple,90
```
nacenik, line 3: delete all this files and use just one input and one output files with appropriate path

src/main/java/core/exception/ReadException.java
```diff
@@ -0,0 +1,7 @@
+package core.exception;
+
+public class ReadException extends RuntimeException {
+    public ReadException(String message) {
+        super(message);
```
nacenik, line 5: why do you create only a read exception? Read what? Rename it 
```suggestion
public class FileReadException extends RuntimeException {
    public FileReadException(String message) {
        super(message);
```

src/main/java/core/exception/FruitException.java
```diff
@@ -0,0 +1,7 @@
+package core.exception;
+
+public class FruitException extends RuntimeException {
```
nacenik, line 3: Bad exception name 
```suggestion
public class OperationHandlerException extends RuntimeException {
```

src/main/java/core/service/impl/Readable.java
```diff
@@ -0,0 +1,5 @@
+package core.service.impl;
+
+public interface Readable {
```
nacenik, line 3: bad name. ReadService or ReaderService is much better.

src/main/java/core/service/impl/Writable.java
```diff
@@ -0,0 +1,5 @@
+package core.service.impl;
+
+public interface Writable {
```
nacenik, line 3: Bad name

src/main/java/core/service/FruitStore.java
nacenik: not fixed

src/main/java/core/Main.java
```diff
@@ -0,0 +1,13 @@
+package core;
+
+import core.strategy.ReportFactory;
+
+public class Main {
+
+    public static void main(String[] args) {
```
nacenik, line 7: Your program should have a good structure and services name 
here is a diagram that should help you 
image

you don't use SOLID in your solution. Fix it. Every service should have an interface and implementation and should be called from the main (see common mistakes carefully)


Be careful with PR names 
image
"doned"  is the wrong word.

nacenik (changes requested):
let's improve your solution

src/main/java/core/service/impl/ReadService.java
```diff
@@ -1,5 +1,5 @@
 package core.service.impl;
```
Vasul-Pankiv, line 1: interface should be in service package and its implementation - in imp packege. You did the opposite

src/main/java/core/strategy/DataOperationStrategyImpl.java
```diff
@@ -0,0 +1,22 @@
...
+public class DataOperationStrategyImpl implements DataOperationStrategy {
+    @Override
+    public Map<OperationType, OperationHandler> strategy() {
+        Map<OperationType, OperationHandler> operationHandlers = new HashMap<>();
+        operationHandlers.put(OperationType.B, new BalanceOperationHandler());
+        operationHandlers.put(OperationType.S, new SupplyOperationHandler());
+        operationHandlers.put(OperationType.R, new ReturnOperationHandler());
+        operationHandlers.put(OperationType.P, new PurchaseOperationHandler());
+        return operationHandlers;
+    }
```
Vasul-Pankiv, line 21: Why you ignore comment about your strategy second time? We won't review your work until you fix it
"move this hashMap to main method and give it to constructor of your strategy"

Vasul-Pankiv (changes requested):

src/main/java/core/strategy/DataOperationStrategyImpl.java
DaniilNepelev: I didn't ignore. I thought that it will be better..

Vasul-Pankiv: If you did not fix the code and did not react to the comment in any way - it is called ignored
So if you think your solution is better - please, provide some arguments

Vasul-Pankiv (changes requested):

src/main/java/core/strategy/DataOperationStrategyImpl.java
Vasul-Pankiv: Please, read the discussion in slack before review

andrii-hoienko (approved):
Good job!

PR #855, opened 2023-08-01
Jv fruit shop hw

ihorvasylenko23:


satori11111 (approved):
Awesome solution!!! You can create your own exception, but it is maybe Optional

keda-vlad (approved):

src/main/java/core/basesyntax/service/strategy/Balance.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.service.strategy;
+
+import core.basesyntax.service.ActivitiesService;
+
+public class Balance implements ActivitiesService {
```
nacenik, line 5: Follow name convention. Fix it for all operation 
```suggestion
public class BalanceActivityHandler implements ActivitiesService {
```

src/main/java/core/basesyntax/service/strategy/Balance.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.service.strategy;
+
+import core.basesyntax.service.ActivitiesService;
+
+public class Balance implements ActivitiesService {
+    @Override
+    public int quantityModify(int adderBefore) {
+        return adderBefore;
+    }
```
nacenik, line 9: what is the reason for this strategy? Just return the quantity that you put as a parameter? 
I'd recommend you to make an exact operation in every single operation, not return just an the same quantity

nacenik (changes requested):
Let's improve your solution

src/main/java/core/basesyntax/Main.java
```diff
@@ -3,49 +3,56 @@
...
+        activitiesServiceMap.put(Fruit.Operation.SUPPLY, new SupplyActivityHandler());
+        activitiesServiceMap.put(Fruit.Operation.RETURN, new ReturnActivityHandler());
+
         //convert data from file to Java object
-        FruitDao newDao = new FruitDaoImpl();
-        DaoService daoService = new DaoServiceImpl(newDao);
-        daoService.convertData(input);
+        FruitDao fruitDao = new FruitDaoImpl();
+        DataConvertor dataConvertor = new DataConvertorImpl();
+        List<Fruit> fruits1 = dataConvertor.convertData(input);
```
Vasul-Pankiv, line 43: Bad naming

Vasul-Pankiv (approved):

PR #856, opened 2023-08-01
implemented task from README.md

Denis-Balako:


keda-vlad (approved):

ivan0dyatlyukkk (approved):

src/main/java/core/basesyntax/service/operation/OperationStrategyImpl.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax.service.operation;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.Map;
+
+public class OperationStrategyImpl implements OperationStrategy {
+    private static final Map<FruitTransaction.OperationType, OperationHandler> map =
+            Map.of(FruitTransaction.OperationType.BALANCE, new BalanceSetOperationHandler(),
+                    FruitTransaction.OperationType.SUPPLY, new AdditionOperationHandler(),
+                    FruitTransaction.OperationType.RETURN, new AdditionOperationHandler(),
```
nacenik, line 10: Map with operation should be created in main and set here via constructor 
`supply` and `return` should be separate operations. 
Rename your operation according to their business logic (e.g.  `BalanceOperationHandler`, `ReturnOperationHandler` and etc.)

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,59 @@
...
+    public enum OperationType {
+        BALANCE("b"),
+        SUPPLY("s"),
+        PURCHASE("p"),
+        RETURN("r");
+
+        private static final Map<String, OperationType> ENUM_MAP;
+        private final String code;
+
+        OperationType(String code) {
```
nacenik, line 39: access modifier

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,59 @@
...
+            return code;
+        }
+
+        static {
+            Map<String, OperationType> map = new ConcurrentHashMap<>();
+            for (OperationType instance : OperationType.values()) {
+                map.put(instance.getCode().toLowerCase(), instance);
+            }
+            ENUM_MAP = Collections.unmodifiableMap(map);
+        }
```
nacenik, line 53: 1. Why do you need to use ConcurrentHashMap? 
2. that's enough to create a method called `findOperation` or `fromString` that will check your enum and if an operation won't be found - throw an exception.

src/main/java/core/basesyntax/service/FruitService.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+import java.util.Map;
+
+public interface FruitService {
```
nacenik, line 7:
```suggestion
public interface FruitShopService {
```

src/main/java/core/basesyntax/service/FruitService.java
```diff
@@ -0,0 +1,13 @@
...
+    void add(FruitTransaction transaction);
+
+    void addAll(List<FruitTransaction> transactions);
+
+    Map<String, Integer> getAll();
```
nacenik, line 12: it should contain 1 method `processData(List<FruitTransaction> transaction)` that should do the same as your addAll + add. Because you don't add data, you are processing data. It could be a return or supply

nacenik (changes requested):
Good but let's improve your solution :)

src/main/java/core/basesyntax/service/impl/FruitShopServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+        return getDataFromDatabase();
+    }
+
+    private void addTransactionsToDatabase(List<FruitTransaction> transactions) {
+        transactions.forEach(transaction -> operationStrategy
+                .getOperation(transaction.getOperation())
+                .executeOperation(transaction));
+    }
+
+    private Map<String, Integer> getDataFromDatabase() {
```
andrii-hoienko, line 29: I don`t think you need this method, you can directly obtain storage

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Map;
+
+public class ReportServiceImpl implements ReportService {
+    private static final String COMMA_SEPARATOR = ",";
+    private static final String LINE_SEPARATOR = System.lineSeparator();
+
+    @Override
+    public String createReport(Map<String, Integer> data) {
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("fruit")
```
andrii-hoienko, line 13: You can create constant with name HEADER, wich will contain header string and append this constant to stringBuilder, it will be more clear in my mind  because now "fruit" and "quantity" kind of magic strings

andrii-hoienko (approved):
Very good!

PR #857, opened 2023-08-01
Jv fruit shop solution

keda-vlad:


OlegTuryk (approved):

nacenik (changes requested):
Good solution, let's improve it :)

vovasalyha (approved):
Nice, approving your PR, but try to mark fields which you are not intended to change with `final` keyword.

PR #868, opened 2023-08-01
implemented supply, purchase, return order opeartion for fruit shop

DmytroV95:


IvanFadieiev (approved):
Looks okay, nothing to add!

README.md
```diff
@@ -11,10 +11,10 @@ Your tasks are:
```
Vasul-Pankiv, line 11: Do not edit readme

src/main/java/core/basesyntax/db/FruitsStorage.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.db;
+
+import java.math.BigDecimal;
+import java.util.HashMap;
+import java.util.Map;
+
+public class FruitsStorage {
+    private final Map<String, BigDecimal> fruitsStorage = new HashMap<>();
```
Vasul-Pankiv, line 8: make it public, static and use it directly where you need it

src/main/java/Main.java
```diff
@@ -0,0 +1,64 @@
...
+        PurchaseHandler purchaseHandler = new PurchaseHandlerImpl(fruitsStorage);
+        PurchaseService purchaseService = new PurchaseServiceImpl(
+                purchaseHandler, dataParserService);
+        purchaseService.purchase(dataFromFile);
+
+        // 4. Return order
+        OrderReturnHandler orderReturnHandler = new OrderReturnHandlerImpl(fruitsStorage);
+        OrderReturnService orderReturnService = new OrderReturnServiceImpl(
+                orderReturnHandler, dataParserService);
+        orderReturnService.returnOrder(dataFromFile);
```
Vasul-Pankiv, line 56: you have to implement strategy pattern to handle all transaction (see requirements)

src/main/java/core/basesyntax/handler/impl/OrderReturnHandlerImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import core.basesyntax.db.FruitsStorage;
+import core.basesyntax.handler.OrderReturnHandler;
+import java.math.BigDecimal;
+
+public class OrderReturnHandlerImpl implements OrderReturnHandler {
+    private final FruitsStorage storage;
+
+    public OrderReturnHandlerImpl(FruitsStorage storage) {
+        this.storage = storage;
+    }
```
Vasul-Pankiv, line 12: storage will be sttic, so you dont need it

src/main/java/core/basesyntax/service/OperationService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.Operation;
+
+public interface OperationService {
+    Operation getOperation(String operationType);
```
Vasul-Pankiv, line 6: add an applyOperation() method and all your operation have to implement this interface and override applyOperation() method or better to create OperationHandler interface with this method

src/main/java/core/basesyntax/service/impl/OperationServiceImpl.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.Operation;
+import core.basesyntax.service.OperationService;
+
+public class OperationServiceImpl implements OperationService {
+    @Override
+    public Operation getOperation(String typeFromFile) {
```
Vasul-Pankiv, line 8: bad name for param, we could pass data not only from file

src/main/java/core/basesyntax/service/impl/OrderReturnServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+import static core.basesyntax.Constants.FRUIT_NAME_ARRAY_INDEX;
+
+import core.basesyntax.handler.OrderReturnHandler;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Operation;
+import core.basesyntax.service.DataParserService;
+import core.basesyntax.service.OrderReturnService;
+import java.util.List;
+
+public class OrderReturnServiceImpl implements OrderReturnService {
```
Vasul-Pankiv, line 12: Or better to create 
```suggestion
public class ReturnOperationHandler implements OperationHandler {
```

src/main/java/core/basesyntax/service/impl/ReadFromFileServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ReadFromFileServiceImpl implements ReadFromFileService {
+    @Override
+    public List<String> redData(String fromFileName) {
```
Vasul-Pankiv, line 13:
```suggestion
    public List<String> readData(String filePath) {
```

src/main/java/Main.java
```diff
@@ -0,0 +1,64 @@
...
+    public static void main(String[] args) {
+        FruitsStorage fruitsStorage = new FruitsStorage();
+        OperationService operationService = new OperationServiceImpl();
+        DataParserService dataParserService = new DataParserServiceImpl(operationService);
+
+        // 1. Read data from file
+        ReadFromFileService fileReader = new ReadFromFileServiceImpl();
+        List<String> dataFromFile = fileReader.redData("src"
+                        + File.separator + "main" + File.separator + "resources"
+                        + File.separator + INPUT_DATA_FILE_NAME);
```
Vasul-Pankiv, line 38: make input and output files path a constant

src/main/java/core/basesyntax/service/impl/ReadFromFileServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+        List<String> dataFromFileList = new ArrayList<>();
+        File file = new File(fromFileName);
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
```
Vasul-Pankiv, line 16:
```suggestion
        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(filePath))) {
```

Vasul-Pankiv (changes requested):
You need to implement strategy and refactor your operationServices

src/main/java/core/basesyntax/service/impl/OrderReturnServiceImpl.java
DmytroV95: What about ReturnOperationService? Becouse i have handler already

src/main/java/core/basesyntax/service/impl/ReadFromFileServiceImpl.java
DmytroV95: I use referense "file" in exception message. So i need it. To get full information about the problem.

README.md
DmytroV95: Hmm.. did not do anything

src/main/java/core/basesyntax/service/impl/SupplierServiceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    @Override
+    public boolean applyOperation(FruitTransaction fruitTransaction) {
+        FruitTransaction newFruit = new FruitTransaction.FruitBuilder()
+                .setOperationType(fruitTransaction.getOperationType())
+                .setFruitName(fruitTransaction.getFruitName())
+                .setFruitQuantity(fruitTransaction.getFruitQuantity())
+                .setOperationType(fruitTransaction.getOperationType())
+                .build();
```
nacenik, line 21: why do you need this block of code?

src/main/java/core/basesyntax/service/impl/SupplierServiceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+        return supplierHandler.add(newFruit);
```
nacenik, line 22: do you really need so many redundant class? 
Here is diagram that should help you to improve your architecture 
image

nacenik (changes requested):
let's improve your solution

src/main/java/core/basesyntax/service/impl/OperationHandlerImpl.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.Operation;
+import core.basesyntax.service.OperationService;
+
+public class OperationHandlerImpl implements OperationService {
+    @Override
+    public Operation getOperation(String operationType) {
```
vovasalyha, line 8: It's not very flexible solution to use switch case here. What if you would need to support 100 operations? Simply iterate through all operation values and return if values match.

Also, it's redundant to create separate service to get Operation by it's value, you can do it in Operation itself.

src/main/java/core/basesyntax/service/impl/PurchaseHandlerImpl.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.db.FruitsStorage;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.OperationHandler;
+
+public class PurchaseHandlerImpl implements OperationHandler {
+    @Override
+    public boolean applyOperation(FruitTransaction fruitTransaction) {
+        boolean isSubstracted = false;
```
vovasalyha, line 10: Why do you need to return boolean value if operation applied? The only way it can be not applied is when some exception appear, but in this case you would not reach return statement anyway.

vovasalyha (approved):

PR #879, opened 2023-08-03
I need a help

Stipaxa-hub:


src/main/java/core/service/impl/TransactionParserImpl.java
```diff
@@ -0,0 +1,23 @@
...
+    private static final int INDEX_CODE_OPERATION = 0;
+    private static final int INDEX_FRUIT_NAME = 1;
+    private static final int INDEX_FRUIT_QUANTITY = 2;
+
+    @Override
+    public List<FruitTransaction> parseData(List<String> data) {
+        return data.stream()
+                .map(element -> element.split(SEPARATOR))
+                .map(element -> new FruitTransaction(FruitTransaction.Operation.getOperationByCode(element[INDEX_CODE_OPERATION]),
+                        element[INDEX_FRUIT_NAME], Integer.parseInt(element[INDEX_FRUIT_QUANTITY])))
```
okuzan, line 20: getOperationByCode throws IllegalArgumentException, parseInt throws NumberFormatException. Later on, you will have to write tests for fruitshop, so it's probably a good idea to think about handling cases of wrong input format now. Let's catch them and throw an exception informing others that the data has a wrong format

src/main/java/core/service/impl/ReaderFromFileImp.java
```diff
@@ -0,0 +1,22 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ReaderFromFileImp implements ReaderFromFile {
+    @Override
+    public List<String> readFromFileData(String filePath) {
+        try(BufferedReader reader = new BufferedReader(new FileReader(filePath))) {
+            return reader.lines().skip(1).collect(Collectors.toList());
```
okuzan, line 17:
```suggestion
            return reader.lines().skip(OFFSET).collect(Collectors.toList());
```
let's make it more visible, maybe at some point the header will have different number of lines

src/main/java/core/service/operation/OperationHandler.java
```diff
@@ -0,0 +1,7 @@
+package core.service.operation;
+
+import core.model.FruitTransaction;
+
+public interface OperationHandler {
+    void operation(FruitTransaction transaction);
```
okuzan, line 6:
```suggestion
    void handle(FruitTransaction transaction);
```

src/main/java/core/service/OperationStrategy.java
```diff
@@ -0,0 +1,4 @@
+package core.service;
+
+public interface OperationStrategy {
```
okuzan, line 3: strategy is a simple class. The idea behind it - is the ability to provide a respective handler to each operation type. Which data structure sounds suitable for this task? Create a constructor, where you will pass the map, and set it to a field. Than create a method, that would provide handler for an operation type

okuzan (changes requested):
Good job overall! You have a solid base. Major points left - operation strategy and report service. You can create the Map in Main and pass it to the strategy. Also, as of now you don't use Strategy anywhere, you have to pass  it as a parameter somewhere to include it in your program

olekskov (changes requested):
Seems fine overall but build is failed due to code-style errors, please fix them

kozhukhovsky (approved):
Good job!

PR #880, opened 2023-08-03
create my own fruit shop

karvatskyi:


pom.xml
```diff
@@ -17,8 +17,12 @@
             https://raw.githubusercontent.com/mate-academy/style-guides/master/java/checkstyle.xml
         </maven.checkstyle.plugin.configLocation>
     </properties>
-
     <dependencies>
+        <dependency>
+            <groupId>com.opencsv</groupId>
+            <artifactId>opencsv</artifactId>
+            <version>5.8</version>
+        </dependency>
```
olekskov, line 25: cannot see where this dependency is required

src/main/java/core/basesyntax/service/operation/BalanceOperation.java
```diff
@@ -0,0 +1,15 @@
...
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.FruitShopService;
+import core.basesyntax.strategy.OperationHandler;
+
+public class BalanceOperation implements OperationHandler {
+
+    @Override
+    public void processWithTransaction(FruitTransaction transaction) {
+        FruitShopService fruitShopService = new FruitShopServiceImpl();
+        fruitShopService.balanceOfFruit(transaction);
```
olekskov, line 13: Don't see sense to create additional service to just call it's method from handler.
Just move logic to handlers

src/main/java/core/basesyntax/strategy/OperationStrategy.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public interface OperationStrategy {
+    Map<FruitTransaction.Operation, OperationHandler> operationHashMap = new HashMap<>();
+    FruitTransaction.Operation getOperation(FruitTransaction transaction,
+                                            Map<FruitTransaction.Operation,
+                                                    OperationHandler> operationHashMap);
+
+}
```
olekskov, line 13: 1. I cannot see any usage of this interface.
2. Please create main method where you will use your components so mentor can see how you intend to use them from reading file till writing report because as is it seems for me that some required logic in the middle is missing
3. Map<FruitTransaction.Operation, OperationHandler> operationHashMap should be argument in method, it should be class variable and passed to constructor

olekskov (changes requested):

src/main/java/core/basesyntax/service/ParserService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+import java.util.List;
+
+public interface ParserService<T> {
+    List<T> parser(List<String> inputData);
```
kozhukhovsky, line 6:
```suggestion
    List<T> parse(List<String> inputData);
```

Methods are usually named a verb.

src/main/java/core/basesyntax/service/ReportService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import java.util.List;
+import java.util.Map;
+
+public interface ReportService<T> {
+    List<String> report(Map<String, Integer> map);
```
kozhukhovsky, line 7:
```suggestion
    List<String> createReport(Map<String, Integer> map);
```

src/main/java/core/basesyntax/service/operation/BalanceOperation.java
kozhukhovsky: not fixed

kozhukhovsky (changes requested):

src/main/java/Main.java
```diff
@@ -0,0 +1,41 @@
+import core.basesyntax.db.Storage;
```
olekskov, line 1: this class should be in core.basesyntax package

src/main/java/core/basesyntax/service/operation/SupplyOperation.java
```diff
@@ -0,0 +1,18 @@
...
+    private final FruitShopService fruitShopService;
+
+    public SupplyOperation(FruitShopService fruitShopService) {
+        this.fruitShopService = fruitShopService;
+    }
+
+    @Override
+    public void processWithTransaction(FruitTransaction transaction) {
+        fruitShopService.supplyFruit(transaction);
+    }
```
olekskov, line 17: Please move logic from fruitShopService to handlers like below, there is no need for separate service with such usage
```suggestion
    @Override
    public void processWithTransaction(FruitTransaction transaction) {
        if (Storage.getFruits().containsKey(transaction.getFruit())) {
            int quantity = Storage.getFruits().get(transaction.getFruit());
            Storage.getFruits().put(transaction.getFruit(),transaction.getQuantity() + quantity);
        }
        Storage.getFruits().put(transaction.getFruit(), transaction.getQuantity());
    }
```

olekskov (changes requested):

olekskov (approved):

PR #895, opened 2023-08-21
Added Interfase: FruitDao, OperationHandler, ParserService, ReaderSer…

hanshyn:
…vice, GeheratopReport, WriterService, FruitService, OperationStrategy and added implaments for this interfaces.  Also added class: Storege, FruitTransaction, Main

### DIFF
--- a/src/main/java/core/basesyntax/service/operation/OperationHandler.java
+++ b/src/main/java/core/basesyntax/service/operation/OperationHandler.java
@@ -1,0 +1,9 @@
+package core.basesyntax.service.operation;
+
+import core.basesyntax.dao.FruitDao;
+import core.basesyntax.dao.FruitDaoImpl;
+
+public interface OperationHandler {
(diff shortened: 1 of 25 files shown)


src/main/java/core/basesyntax/service/operation/OperationHandler.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.service.operation;
+
+import core.basesyntax.dao.FruitDao;
+import core.basesyntax.dao.FruitDaoImpl;
+
+public interface OperationHandler {
+    FruitDao fruitDao = new FruitDaoImpl();
```
olekskov, line 7: Please avoid hardcoding dependencies with new. It often makes code not testable and not maintainable.
Use constructor in implementations instead

olekskov (changes requested):
Very nice job for 1st try!
Left some comments to correct before proceeding to next task

kozhukhovsky (approved):

PR #896, opened 2023-08-21
Homework jv-fruit-shop

zkolesnyk:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,61 @@
...
+        ParserService parserService = new ParserServiceImpl();
+        List<FruitTransaction> fruitTransactions = parserService.parse(dataFromFile);
+
+        for (FruitTransaction fruitTransaction : fruitTransactions) {
+            OperationHandler operationHandler =
+                    operationStrategy.get(fruitTransaction.getOperation());
+            Fruit fruit = new Fruit(fruitTransaction.getName(),
+                    fruitTransaction.getQuantity());
+            operationHandler.operate(fruit);
+        }
```
olekskov, line 54: Would be nice to move this logic to some new service class

src/main/java/core/basesyntax/service/impl/ParserServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+    private static final int QUANTITY_POSITION = 2;
+
+    @Override
+    public List<FruitTransaction> parse(List<String> lines) {
+        List<FruitTransaction> fruitTransactions = new ArrayList<>();
+        if (lines.isEmpty()) {
+            throw new RuntimeException("Data is empty.");
+        }
+        for (int i = 1; i < lines.size(); i++) {
+            String[] parsedLine = lines.get(i).split(SEPARATOR);
```
olekskov, line 21:
```suggestion
        for (line : lines) {
            String[] parsedLine = line.split(SEPARATOR);
```

olekskov (changes requested):
Very nice job for 1st try.
Only some minor style comments from my side

src/main/java/core/basesyntax/service/impl/ParserServiceImpl.java
zkolesnyk: but i need to start with index = 1

kozhukhovsky (approved):

PR #897, opened 2023-08-22
jv-fruit-shop

Ka1zerov:


src/main/java/service/impl/ActivityReaderServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+                );
+            }
+            reader.close();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't find or read file", e);
+        }
+
+        return activityList;
+    }
+}
```
olekskov, line 38: This class doesn't follow single responsibility principle. It's doing both reading from file and parsing content.
Also it performs additional actions during keeping opened connection to file what may be not the best solution.
Please split this for 2 classes, first will read file content as String or String[] or List<String>, second will parse this result to List<Activity>

And also use try-with-resources in reworked reader

src/main/java/Main.java
```diff
@@ -0,0 +1,54 @@
...
+        ActivityReaderService activityReaderService = new ActivityReaderServiceImpl();
+        final List<Activity> activityList = activityReaderService.readFromFile(FROM_FILE_NAME);
+
+        Map<String, Integer> fruitBox = new HashMap<>();
+        fruitBox.put("apple", 0);
+        fruitBox.put("banana", 0);
+        Storage storage = new Storage(fruitBox);
+
+        Map<ActivityType, ActivityTypeHandler> activityTypeHandlerMap = new HashMap<>();
+        activityTypeHandlerMap.put(ActivityType.BALANCE, new AdditionTypeHandler());
```
olekskov, line 33: Balance operation should replace amount of fruit, not to be added

src/main/java/Main.java
```diff
@@ -0,0 +1,54 @@
...
+
+        ActivityTypeStrategy activityTypeStrategy = new ActivityTypeStrategyImpl(
+                activityTypeHandlerMap);
+
+        for (Activity activity : activityList) {
+            storage.getFruitBox().put(activity.getFruit(), activityTypeStrategy
+                    .get(activity.getType())
+                    .getNewQuantity(storage.getFruitBox()
+                            .get(activity.getFruit()), activity.getQuantity()));
+        }
```
olekskov, line 46: 1) Please extract this logic to separate service class
2) This code will not work with any other fruits except added to Map on lines 28-29
Please rework solution so input file may contain any amount of unpredictable fruits

src/main/java/Main.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final String FROM_FILE_NAME = "src/files/fruit.csv";
+    private static final String TO_FILE_NAME = "src/files/report.csv";
+
+    public static void main(String[] args) {
+        ActivityReaderService activityReaderService = new ActivityReaderServiceImpl();
+        final List<Activity> activityList = activityReaderService.readFromFile(FROM_FILE_NAME);
+
+        Map<String, Integer> fruitBox = new HashMap<>();
+        fruitBox.put("apple", 0);
+        fruitBox.put("banana", 0);
```
olekskov, line 29: Solution should work with any fruits, we do not know content of file we will get as input
```suggestion
```

olekskov (changes requested):
Good job overall but solution require improvements

src/main/java/service/impl/CsvReaderServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+            while ((line = reader.readLine()) != null) {
+                dataFromFile.append(line).append(System.lineSeparator());
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't find or read file", e);
+        }
+        return dataFromFile.toString().split(System.lineSeparator());
+    }
+}
+
```
kozhukhovsky, line 26:
```suggestion
```

src/main/java/service/impl/ActivityParserServiceImpl.java
```diff
@@ -0,0 +1,24 @@
+package service.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+import model.Activity;
+import model.ActivityType;
+import service.ActivityParserService;
+
+public class ActivityParserServiceImpl implements ActivityParserService {
+
```
kozhukhovsky, line 10:
```suggestion
```

kozhukhovsky (approved):

PR #899, opened 2023-08-25
Hw fruit shop

kaptetos:


kozhukhovsky (changes requested):
used In this task it is important to create a proper architecture. Your main task is to divide logic, and create separate independent services for each step of the whole process.

- FileReader  to read data about transactions from file

- TransactionParser  parses list of rows from file into list of FruitTransaction class

- OperationHandler  and an implementation for each operation type, each OperationHandler should make changes to the storage

- Storage class - that should have Map<String, Integer>. This class is a simplier DB. Key of the map is a fruit name, and value is amount of that fruit that is currently in our storage

- OperationStrategy that should have a map of handlers, and should return a handler by operation type

- TransactionProcessor is a service that should accept list of transactions and process them all

- ReportCreator should create a report from values that are in storage

- FileWriter writes a string value to a file



This task is to practice SOLID principles:

You should create separate services that have its own responsibilities

You are creating interfaces - defining functionality, and creating an implementation.

For example: TransactionParser is an abstraction, you should create an implementation CsvTransactionParser which parses transactions from SCV format. In the future, for other formats such as pdf, xlsx, etc. you would need to create a new implementation XlsxTransactionParser. There would be no need to modify existing code.

When your service needs some dependency you should pass from the outside, and be dependant on abstractions.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,50 @@
...
+                transactionData = new ArrayList<>();
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    transactionData.add(line);
+                }
+            }
+        } catch (IOException e) {
+            System.err.println("Error opening/reading file: " + e.getMessage());
+            return;
+        }
```
olekskov, line 28: I can't understand why you FileReader service isn't used here

src/main/java/core/basesyntax/CsvReportCreator.java
```diff
@@ -0,0 +1,34 @@
...
+public class CsvReportCreator implements ReportCreator {
+    private String filePath;
+
+    public CsvReportCreator(String filePath) {
+        this.filePath = filePath;
+    }
+
+    @Override
+    public void createReport(Map<String, Integer> fruitInventory,
+                             String filePath) throws IOException {
```
olekskov, line 16: You have exception handling in catch block, there is no sense to also have this exception in method header
```suggestion
    public void createReport(Map<String, Integer> fruitInventory,
                             String filePath) {
```

src/main/java/core/basesyntax/CsvReportCreator.java
```diff
@@ -0,0 +1,34 @@
...
+                int quantity = entry.getValue();
+
+                String line = String.format("%s,%d\n", fruit, quantity);
+                writer.write(line);
+
+            }
+
+            System.out.println("Report created successfully" + filePath);
+        } catch (IOException e) {
+            System.err.println("Error creating report" + e.getMessage());
```
olekskov, line 31: We're stopping application logic by throwing RuntimeException with proper message and cause of failure
```suggestion
            throw new RuntimeException("Error creating report", e);
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,50 @@
...
+        transactionProcessor.processTransactions(transactions);
+
+        ReportCreator reportCreator = new CsvReportCreator(outputFilePath);
+
+        Map<String, Integer> fruitInventory = storage.getFruitInventory();
+        try {
+            reportCreator.createReport(fruitInventory, outputFilePath);
+        } catch (IOException e) {
+            System.err.println("Error creating report: " + e.getMessage());
+        }
```
olekskov, line 48: after proper exception handling in method there is no need to have try/catch here
```suggestion
        reportCreator.createReport(fruitInventory, outputFilePath);
```

src/main/java/core/basesyntax/CsvFileReader.java
```diff
@@ -0,0 +1,21 @@
...
+    public List<String> readData(String filePath) throws IOException {
+        List<String> data = new ArrayList<>();
+        try (BufferedReader reader = new BufferedReader(new FileReader(filePath))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                data.add(line);
+            }
+        }
+        return data;
+    }
```
olekskov, line 20: 1. I can't see any logic specific for csv. Please remove csv from naming. FileReaderImpl would be fine
2. Please add catch block and throw RuntimeException there as shown in comment for FileWriter

src/main/java/core/basesyntax/DefaultOperationStrategy.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class DefaultOperationStrategy implements OperationStrategy {
+    private Map<FruitTransaction.Operation, OperationHandler> handlers = new HashMap<>();
+
+    public DefaultOperationStrategy() {
+        handlers.put(FruitTransaction.Operation.BALANCE, new BalanceHandler());
+        handlers.put(FruitTransaction.Operation.SUPPLY, new SupplyHandler());
+        handlers.put(FruitTransaction.Operation.PURCHASE, new PurchaseHandler());
+        handlers.put(FruitTransaction.Operation.RETURN, new ReturnHandler());
+    }
```
olekskov, line 14: Please initialize map in Main method and pass through constructor. This will make code more maintainable and testable

```suggestion
    private Map<FruitTransaction.Operation, OperationHandler> handlers;

    public DefaultOperationStrategy(Map<FruitTransaction.Operation, OperationHandler> handlers) {
        this.handlers = handlers;
    }
```

olekskov (changes requested):
Good job overall but require some improvements.
Please check comments and also please make some packages and split classes among them.
It's already hard to go through when such amount of classes are just placed in base package of project

src/main/java/core/basesyntax/FileReaderImpl.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FileReaderImpl implements FileReaderService {
```
Ivan95kos, line 9: This `Service` keyword indicates that there is business logic here
```suggestion
public class FileReaderServiceImpl implements FileReaderService {
```

Alternatively, you can rewrite as in the following logic
```
    @Override
    public List<String> readFromFile(String fromFilePath) {
        try {
            return Files.readAllLines(Path.of(fromFilePath));
        } catch (IOException e) {
            throw new RuntimeException("Can't read data from file "
            + fromFilePath + " !", e);
        }
    }
```

src/main/java/core/basesyntax/FruitTransaction.java
```diff
@@ -0,0 +1,54 @@
...
+
+        private String code;
+
+        Operation(String code) {
+            this.code = code;
+        }
+
+        public String getCode() {
+            return code;
+        }
```
Ivan95kos, line 52: add the next method:
```suggestion
        }
        
        public static Operation getByCode(String character) {
            for (Operation operation : values()) {
                if (operation.getOperation().equals(character)) {
                    return operation;
                }
            }
            throw new RuntimeException("Unknown operation");
        }
```

src/main/java/core/basesyntax/Main.java
Ivan95kos: not fixed

src/main/java/core/basesyntax/parser/CsvTransactionParser.java
```diff
@@ -0,0 +1,26 @@
...
+import core.basesyntax.FruitTransaction;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CsvTransactionParser implements TransactionParser {
+    @Override
+    public List<FruitTransaction> parseTransactions(List<String> data) {
+        List<FruitTransaction> transactions = new ArrayList<>();
+        for (String line : data) {
+            String[] parts = line.split(",");
```
Ivan95kos, line 12:
```suggestion
            String[] parts = line.split(SEPARATOR);
```

src/main/java/core/basesyntax/parser/CsvTransactionParser.java
```diff
@@ -0,0 +1,26 @@
...
+            if (parts.length == 3) {
+                String operationCode = parts[0];
+                String fruit = parts[1];
```
Ivan95kos, line 15: use constants instead of numbers

src/main/java/core/basesyntax/report/CsvReportCreator.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax.report;
+
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Map;
+
+public class CsvReportCreator implements ReportCreator {
```
Ivan95kos, line 7: Here you have two logics
Divide it into:

```
    String report();
}
```

```
    void writeToFile(String data, String outputPath);
}
```

src/main/java/core/basesyntax/report/CsvReportCreator.java
```diff
@@ -0,0 +1,33 @@
...
+    private String filePath;
+
+    public CsvReportCreator(String filePath) {
+        this.filePath = filePath;
+    }
+
+    @Override
+    public void createReport(Map<String, Integer> fruitInventory, String filePath) {
+        try (FileWriter writer = new FileWriter(filePath)) {
+            writer.write("fruit,quantity\n");
```
Ivan95kos, line 17: All magic numbers, strings, and symbols in your code should be constants.
[this article](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)
```suggestion
            writer.write(HEADER);
```

src/main/java/core/basesyntax/report/CsvReportCreator.java
```diff
@@ -0,0 +1,33 @@
...
+            for (Map.Entry<String, Integer> entry : fruitInventory.entrySet()) {
+                String fruit = entry.getKey();
+                int quantity = entry.getValue();
+
+                String line = String.format("%s,%d\n", fruit, quantity);
+                writer.write(line);
+
+            }
+
+            System.out.println("Report created successfully" + filePath);
```
Ivan95kos, line 28: avoid `System.out.println` in code, use `Logger` for that
```suggestion
```

Ivan95kos (changes requested):
The project should have packages (service, db, model, strategy...)
package `service` should have: FruitShopService, ParserService, ReaderService, ReportService, WriterService 
package `db` should have: Storage 
package `model` should have: FruitTransaction
package `strategy` should have: OperationStrategy , OperationHandler
All implementation should be to move to package `impl`
This will be a good code decomposition.

kozhukhovsky (changes requested):
No changes since the last review.

olekskov (approved):

PR #903, opened 2023-08-30
version1-stas

stas83ggg:


src/main/java/Main.java
```diff
@@ -0,0 +1,6 @@
+
+public class Main {
+    public static void main(String[] args) {
+
```
kozhukhovsky, line 4: image

src/main/java/service/impl/FileWriterServiceImpl.java
```diff
@@ -0,0 +1,20 @@
+package service.impl;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import service.FileWriterService;
+
+public class FileWriterServiceImpl implements FileWriterService {
+
```
kozhukhovsky, line 10:
```suggestion
```

kozhukhovsky (changes requested):

src/main/java/service/TransactionParse.java
```diff
@@ -0,0 +1,8 @@
+package service;
+
+import java.util.List;
+import model.FruitTransaction;
+
+public interface TransactionParse {
+    List<FruitTransaction> transactionParse(List<String> data);
```
olekskov, line 7:
```suggestion
public interface TransactionParser {
    List<FruitTransaction> parseTransactions(List<String> data);
```

src/main/java/service/operation/OperationTransaction.java
```diff
@@ -0,0 +1,7 @@
+package service.operation;
+
+import model.FruitTransaction;
+
+public interface OperationTransaction {
+    void fruitOperation(FruitTransaction fruitTransaction);
```
olekskov, line 6:
```suggestion
public interface TransactionHadler {
    void handleTransaction(FruitTransaction fruitTransaction);
```

src/main/java/strategy/OperationStrategy.java
```diff
@@ -0,0 +1,8 @@
+package strategy;
+
+import model.FruitTransaction;
+import service.operation.OperationTransaction;
+
+public interface OperationStrategy {
+    OperationTransaction getOperation(FruitTransaction.Operation operation);
```
olekskov, line 7:
```suggestion
    OperationTransaction getHandler(FruitTransaction.Operation operation);
```

src/main/java/service/impl/FileWriterServiceImpl.java
```diff
@@ -0,0 +1,19 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import service.FileWriterService;
+
+public class FileWriterServiceImpl implements FileWriterService {
+    @Override
+    public void fileWriter(String report, String writeFile) {
```
olekskov, line 11:
```suggestion
    public void writeTofile(String report, String writeFile) {
```

olekskov (changes requested):
Good job overall.
Need to fix classes/methods naming because it's not clear sometimes

stas83ggg:
Hi! Could You check my code ? I don't understand why nothing added to report file.
Снимок экрана 2023-09-03 в 21 12 06

kozhukhovsky (changes requested):
Hello, I need to debug the code, somewhere there is an error in parsing a file or transactions. Visually, it is very difficult to detect, try to set breakpoints and understand what happens at what stage.
I will reject the PR because it does not work as it should.
You can contact slack, we will help you to debug.

src/main/java/service/operation/TransactionHandler.java
```diff
@@ -0,0 +1,7 @@
+package service.operation;
```
Ivan95kos, line 1: It is better to move the package `operation` to the same level as the `service`
Then create a package `impl` inside, and to move there all handlers

Ivan95kos (approved):

PR #906, opened 2023-09-08
implemented fruit storage

RedArtelerist:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,48 @@
...
+    static {
+        OPERATION_HANDLER_MAP.put(Operation.BALANCE, new BalanceOperationHandler());
+        OPERATION_HANDLER_MAP.put(Operation.PURCHASE, new PurchaseOperationHandler());
+        OPERATION_HANDLER_MAP.put(Operation.SUPPLY, new SupplyOperationHandler());
+        OPERATION_HANDLER_MAP.put(Operation.RETURN, new ReturnOperationHandler());
+    }
+
+    public static void main(String[] args) {
+        String dataFilePath = "src/main/resources/monday.csv";
+        String reportFilePath = "src/main/resources/report.csv";
```
teract10s, line 34: Maybe it will be better to create constant here

teract10s (approved):
Good solution! Honestly i don't know what i can add

src/main/java/core/basesyntax/Main.java
liliia-ponomarenko: agree

src/main/java/core/basesyntax/handler/BalanceOperationHandler.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.handler;
+
+import core.basesyntax.db.FruitStorage;
+import core.basesyntax.model.FruitTransaction;
+
+public class BalanceOperationHandler implements OperationHandler {
+    @Override
+    public void process(FruitTransaction transaction) {
+        FruitStorage.STORAGE.put(transaction.getFruit(), transaction.getQuantity());
```
liliia-ponomarenko, line 9: You should add a DAO layer that is responsible for interaction with DB and use it in your handlers

src/main/java/core/basesyntax/model/Operation.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+public enum Operation {
+    BALANCE("b"),
+    SUPPLY("s"),
+    PURCHASE("p"),
+    RETURN("r");
+
+    private static final Map<String, Operation> MAP = new HashMap<>();
```
liliia-ponomarenko, line 14: Looks  like MAP is redundant  variable

src/main/java/core/basesyntax/service/impl/ReaderServiceCsvImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final int FRUIT_NAME_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public List<FruitTransaction> readFromFile(String filePath) {
+        try (Stream<String> lines = Files.lines(Paths.get(filePath))) {
+            return lines
+                    .skip(1)
+                    .map(this::getFruitTransaction)
+                    .collect(Collectors.toList());
```
liliia-ponomarenko, line 26:
```suggestion
                    ..toList();
```

src/main/java/core/basesyntax/service/impl/ReaderServiceCsvImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        return new FruitTransaction(operation, transaction[FRUIT_NAME_INDEX], quantity);
+    }
+
+    private Operation getOperation(String record, String operationType) {
+        return Operation
+                .getByCode(operationType)
+                .orElseThrow(() -> new InvalidDataException(
+                        "File has invalid operation type: " + record)
+                );
+    }
```
liliia-ponomarenko, line 48: move this logic to Operation

src/main/java/core/basesyntax/service/impl/ReaderServiceCsvImpl.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private FruitTransaction getFruitTransaction(String record) {
+        String[] transaction = record.split(SPLIT_DELIMITER);
+        Operation operation = getOperation(record, transaction[OPERATION_INDEX]);
+        int quantity = Integer.parseInt(transaction[QUANTITY_INDEX]);
+        if (quantity < 0) {
+            throw new InvalidDataException("Quantity can't be negative value: " + record);
+        }
+        return new FruitTransaction(operation, transaction[FRUIT_NAME_INDEX], quantity);
+    }
```
liliia-ponomarenko, line 40: This logic is not appropriate for this service. ReaderCsvService just reads data from a file. Let’s  move it to other or separate service

src/main/resources/monday.csv
```diff
@@ -0,0 +1,9 @@
+    type,fruit,quantity
```
liliia-ponomarenko, line 1:
```suggestion
type,fruit,quantity
```

liliia-ponomarenko (changes requested):
Let’s add dao layer  and fix several mistakes  ;)

liliia-ponomarenko (approved):
Great 👍😊

PR #909, opened 2023-09-09
Fruit shop impl

Oleksandr-Tymoshenko:


src/main/java/core/basesyntax/db/StorageDaoImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class StorageDaoImpl implements StorageDao {
+    @Override
+    public void add(String product, Integer value) {
+        if (value >= 0) {
+            Storage.STOCK.put(product, value);
+            return;
+        }
+        throw new RuntimeException("Amount of products '" + product + "' can`t be less than 0");
+    }
```
RedArtelerist, line 13: You can simplify this, just at the beginning check negative value, so you don't need to use return operator.
```suggestion
    public void add(String product, Integer value) {
        if (value < 0) {
        throw new RuntimeException("Amount of products '" + product + "' can`t be less than 0");
        }
        Storage.STOCK.put(product, value);
        
    }

```

src/main/java/core/basesyntax/file/writer/CsvFileWriterImpl.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.List;
+
+public class CsvFileWriterImpl implements CsvFileWriter {
+    @Override
+    public void writeDataToFile(File file, List<String> report) {
+        try (FileWriter fileWriter = new FileWriter(file)) {
+            for (String s : report) {
+                fileWriter.write(s);
+                fileWriter.write(System.lineSeparator());
+            }
```
RedArtelerist, line 15: you can use java.nio.Files class here

src/main/java/core/basesyntax/operationstrategy/operation/PurchaseOperationHandler.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.operationstrategy.operation;
+
+import core.basesyntax.db.StorageDao;
+import core.basesyntax.db.StorageDaoImpl;
+
+public class PurchaseOperationHandler implements OperationHandler {
+    private final StorageDao storageDao = new StorageDaoImpl();
```
RedArtelerist, line 7: I don't know that you need ti provide Dao, because in each handler you create a separate entity of StorageDao, so it maybe more optimal to access the Storage directly. But up to you

src/main/java/core/basesyntax/processdata/convertdata/ConvertDataImpl.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.List;
+
+public class ConvertDataImpl implements ConvertData {
+    @Override
+    public List<List<String>> convertData(List<String> data) {
+        return data.stream()
+                .filter(s -> s.charAt(0) == ' ')
+                .map(String::trim)
+                .map(s -> List.of(s.split(",")))
+                .toList();
```
RedArtelerist, line 12: Sorry, but I can't understand this code. Why do we need this filter here, why do we filter only those lines that begin with an empty character. This code is not very intuitive. And you return a list of lists of strings. Why not create a separate entity for the transaction and parse it here and return a list of transactions, as shown in the example for the task.

src/main/java/core/basesyntax/db/StorageDaoImpl.java
```diff
@@ -0,0 +1,33 @@
...
+    public void addAmount(String product, Integer amountToAdd) {
+        Integer value = Storage.STOCK.get(product);
+        value += amountToAdd;
+        add(product, value);
+    }
+
+    @Override
+    public Map<String, Integer> getStock() {
+        return Storage.STOCK;
+    }
```
RedArtelerist, line 32: It a little strange to return all Storage in one method, you can just parse storage in ReportService and return a String as report or list of strings. But I don't sure about this

src/main/java/core/basesyntax/processdata/ProcessDataImpl.java
```diff
@@ -0,0 +1,37 @@
...
+        storageDao = new StorageDaoImpl();
+    }
+
+    @Override
+    public Map<String, Integer> process(List<List<String>> processedData) {
+        for (List<String> dataLine : processedData) {
+            Character operationType = dataLine.get(OPERATION_TYPE_POSITION).charAt(0);
+            String itemName = dataLine.get(ITEM_NAME_POSITION);
+            Integer value = Integer.parseInt(dataLine.get(ITEM_VALUE_POSITION));
+            OperationHandler operationHandler = operationStrategy.get(operationType);
```
RedArtelerist, line 29: Also, it's a little complicated. You can check it right in Reader Service

src/main/java/core/basesyntax/processdata/ProcessDataImpl.java
```diff
@@ -0,0 +1,37 @@
...
+            if (operationHandler == null) {
+                throw new RuntimeException("Incorrrect operation type: " + operationType);
+            }
+            operationHandler.processOperation(itemName, value);
```
RedArtelerist, line 33: Can we you use Optional to get rid of the null check?

RedArtelerist (changes requested):
Good job. I have some advices and thoughts about it, so try improve your solution. In some cases, you can simply think about my suggestions.

src/main/java/core/basesyntax/operationstrategy/operation/PurchaseOperationHandler.java
Oleksandr-Tymoshenko: we use class Storage here as a fake DB, but if it was a real DB, we could do that, so from my point, we need DAO

src/main/java/core/basesyntax/processdata/convertdata/ConvertDataImpl.java
Oleksandr-Tymoshenko: Cause i receive data from file like 

```
columnName, columnName
      data,value
      data,value
```

I need to delete only first line so i found that it hasn`t space at the beginning, but i agree that it is not readable.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,63 @@
...
+        List<String> data = FILE_READER.readDataFromFile(FILE_PATH);
+        List<List<String>> convertData = CONVERT_DATA.convertData(data);
+        Map<String, Integer> stock = PROCESS_DATA.process(convertData);
+        List<String> report = CREATE_REPORT.createReport(stock);
+        File todayReport = new File(TODAY_REPORT_PATH);
+        try {
+            todayReport.createNewFile();
+        } catch (IOException e) {
+            throw new RuntimeException("Cannot create file " + TODAY_REPORT_PATH, e);
+        }
```
liliia-ponomarenko, line 60: Let’s  move this logic to the FileWriter service

src/main/java/core/basesyntax/db/StorageDaoImpl.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax.db;
+
+import java.util.Map;
+
+public class StorageDaoImpl implements StorageDao {
```
liliia-ponomarenko, line 5: Move dao classes to separate package

src/main/java/core/basesyntax/processdata/ProcessDataImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public ProcessDataImpl(Map<Character, OperationHandler> operationHandlerMap) {
+        operationStrategy = new OperationStrategyImpl(operationHandlerMap);
+        storageDao = new StorageDaoImpl();
+    }
+
+    @Override
+    public Map<String, Integer> process(List<List<String>> processedData) {
+        for (List<String> dataLine : processedData) {
+            Character operationType = dataLine.get(OPERATION_TYPE_POSITION).charAt(0);
```
liliia-ponomarenko, line 26:
```suggestion
            Character operationType = dataLine.get(OPERATION_TYPE_POSITION);
```
It’s hardcode think how to cast it in another way. Also,  I recommend you create enum Operation (see example  in Readme)

src/main/java/core/basesyntax/processdata/ProcessDataImpl.java
```diff
@@ -0,0 +1,33 @@
...
+    private final OperationStrategy operationStrategy;
+    private final StorageDao storageDao;
+
+    public ProcessDataImpl(Map<Character, OperationHandler> operationHandlerMap) {
+        operationStrategy = new OperationStrategyImpl(operationHandlerMap);
+        storageDao = new StorageDaoImpl();
+    }
+
+    @Override
+    public Map<String, Integer> process(List<List<String>> processedData) {
```
liliia-ponomarenko, line 24: Looks like this method should be void

src/main/java/core/basesyntax/processdata/createreport/ReportImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Map;
+
+public class ReportImpl implements Report {
+    private final String fruitColumnName = "fruit";
+    private final String fruitAmountColumnName = "quantity";
+    private final String columnSeparator = ",";
+
+    @Override
+    public List<String> createReport(Map<String, Integer> stock) {
+        List<String> report = new ArrayList<>();
```
liliia-ponomarenko, line 14: Let’s  return a simple string and simplify  this method

liliia-ponomarenko (changes requested):
Good start! About project structure. Let’s read common mistakes and try to follow the recommended structure and package names. If you have questions let’s discuss them in Slack.

Elena-Bruyako (approved):

PR #914, opened 2023-09-12
HW Fruit Shop

vrudkovsky:


src/main/java/core/basesyntax/services/impl/FruitReportServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("fruit,quantity").append(System.lineSeparator());
+        for (Map.Entry<String, Integer> entry : Storage.fruitStorage.entrySet()) {
+            stringBuilder.append(entry.getKey())
+                    .append(COMMA)
+                    .append(entry.getValue())
+                    .append(System.lineSeparator());
+        }
+        return stringBuilder.toString();
+    }
```
olekskov, line 39: This is rather different logic and including it into single service and reusing one method in another one violates single responsibility principle and makes code harder to maintain and test.

Please leave getFormattedStringReport in this service under name createReport.
But method that uses strategy should be extracted to separate service and be something like
public void processInputData()
So it's work is only update Storage with input data

src/main/java/core/basesyntax/services/impl/FruitReportServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Map;
+
+public class FruitReportServiceImpl implements FruitReportService {
+    public static final String COMMA = ",";
+
+    @Override
+    public String createReport(List<String> listData,
+                               Map<Transaction.Operation, Operation> operationMap) {
+        Validator validator = new ValidatorImpl();
+        TransactionParser parserService = new TransactionParserImpl();
```
olekskov, line 19: Services we rely on shouldn't be initialized with 'new' in method.
With such practice we create new object each time method is called and making code very hard to maintain and test
Please make services class variable and initialize them via constructor

src/main/java/core/basesyntax/FruitShop.java
```diff
@@ -0,0 +1,41 @@
...
+    public static void main(String[] args) {
+        Map<Transaction.Operation, Operation> operationMap = new HashMap<>();
+        operationMap.put(Transaction.Operation.RETURN, new SupplyOperation());
+        operationMap.put(Transaction.Operation.SUPPLY, new SupplyOperation());
+        operationMap.put(Transaction.Operation.BALANCE, new BalanceOperation());
+        operationMap.put(Transaction.Operation.PURCHASE, new PurchaseOperation());
+
+        FileReader fileReader = new FileReaderImpl();
+        List<String> listData = fileReader.readFromFile(INPUT_FILEPATH);
+        listData.remove(removeIndex);
```
olekskov, line 35: Please move skipping header to method where you work with your listData

olekskov (changes requested):

Ivan95kos (approved):

src/main/java/core/basesyntax/services/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+    @Override
+    public boolean writeToFile(String fileName, String report) {
+        Path path = Paths.get(fileName);
+        try {
+            Files.writeString(path, report);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to file: " + fileName, e);
+        }
+        return true;
+    }
```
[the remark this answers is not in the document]
vrudkovsky: Thanks you!

PR #921, opened 2023-09-14
FruitShop_Implementation

SalnikovMarkiian:


src/main/java/core/basesyntax/strategy/handler/impl/OperationStrategyImpl.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public OperationStrategyImpl(Map<FruitTransaction.Operation,
+            OperationHandler> operationHandlerMap) {
+        this.operationHandlerMap = operationHandlerMap;
+    }
+
+    @Override
+    public OperationHandler getHandler(FruitTransaction.Operation operation) {
+        return operationHandlerMap.get(operation);
+    }
```
Vt0rtikV, line 19: I think you don't need this class at all, but it's up to you

Vt0rtikV (approved):
Good solution. I <3 names of classes, but left some minor comments

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public String getFruit() {
+        return fruit;
+    }
+
+    public int getQuantity() {
+        return quantity;
+    }
+
+    public enum Operation {
```
Andriy-dz, line 28: move it to a separate class

src/main/java/core/basesyntax/strategy/handler/impl/OperationStrategyImpl.java
Andriy-dz: agree with @Vt0rtikV

Andriy-dz (approved):
Very Good Solution!!

PR #922, opened 2023-09-15
Fruit shop realisation

Vt0rtikV:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,45 @@
...
+    public static void main(String[] args) {
+        ReaderService readerService = new CsvReaderServiceImpl();
+        FormaterService formaterService = new FormaterServiceImpl();
+        ProcessServiceImpl processService = new ProcessServiceImpl(HANDLER_MAP);
+        ReportService reportService = new ReportServiceImpl();
+        WriterService writerService = new WriterServiceImpl();
+        processService.manageTransactions(
+                formaterService.format(
+                        readerService.read(PATH_TO_INPUT_FILE)));
+        writerService.writeToFile(reportService.generateReport(), PATH_TO_REPORT_FILE);
```
RedArtelerist, line 43: I think it's better call it separately, receive the result in a separate variable and pass it to the next service method

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    public static final Map<String, Integer> DATA_BASE = new HashMap<>();
```
RedArtelerist, line 7: This is one word
```suggestion
    public static final Map<String, Integer> DATABASE = new HashMap<>();

```

src/main/java/core/basesyntax/model/Operation.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public String getCode() {
+        return code;
+    }
+
+    public static Operation getOperationByCode(String code) {
+        return Arrays.stream(Operation.values())
+                .filter(n -> n.getCode().equals(code))
+                .findFirst()
+                .orElseThrow(NoSuchElementException::new);
```
RedArtelerist, line 26: Maybe better to add explanatory error message

src/main/java/core/basesyntax/service/impl/CsvReaderServiceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+    @Override
+    public List<String> read(String fileName) {
+        List<String> listOfLines = new ArrayList<>();
+        Path path = Paths.get(fileName);
+        try (Stream<String> stream = Files.lines(path).skip(1)) {
+            stream.forEach(listOfLines::add);
+        } catch (IOException e) {
+            throw new RuntimeException("Some problems with file reading: " + e);
+        }
+        return listOfLines;
```
RedArtelerist, line 22: I think this looks better. Also so you can add to exception message a fileName
```suggestion
        try (Stream<String> stream = Files.lines(Paths.get(fileName))) {
            return stream.skip(1).toList();
        } catch (IOException e) {
            throw new RuntimeException("Some problems with file reading: " + e);
        }
```

src/main/java/core/basesyntax/service/impl/FormaterServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+        for (String line: stringList) {
+            String[] splitedLine = line.split(SPLITERATOR);
+            transactionsList.add(
+                    new FruitTransaction(
+                            Operation.getOperationByCode(splitedLine[OPERATION_INDEX]),
+                            splitedLine[FRUIT_NAME_INDEX],
+                            Integer.parseInt(splitedLine[QUANTITY_INDEX])));
+        }
+        return transactionsList;
+    }
```
RedArtelerist, line 32: Maybe we can separate the conversion of a string into a transaction into a method. And already in this method use something like
`return stringList.stream().map(str -> parseMethod(s)); `
And I also don't see some validation:
- if we have negative amount
- if we have non-numeric amount
- if we have invalid csv format (for example "b, banana, 100, afaf" or "b, 20")

src/main/java/core/basesyntax/strategy/impl/ReturnOperationHandler.java
```diff
@@ -0,0 +1,18 @@
...
+import core.basesyntax.strategy.OperationHandler;
+
+public class ReturnOperationHandler implements OperationHandler {
+    @Override
+    public void handleTransaction(FruitTransaction transaction) {
+        String key = transaction.getFruit();
+        if (!Storage.DATA_BASE.containsKey(key)) {
+            //just some logic
+            throw new RuntimeException("No such fruit in shop");
+        }
```
RedArtelerist, line 14: i don't sure that we need this check, because we can just use Storage.DATA_BASE.getOrDefault(key, 0). But up to you

RedArtelerist (changes requested):
Good solutions, let's improve it :)

src/main/java/core/basesyntax/service/impl/FruitStorageServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Operation;
+import core.basesyntax.service.FruitStorageService;
+import core.basesyntax.strategy.OperationHandler;
+import java.util.List;
+import java.util.Map;
+
+public class FruitStorageServiceImpl implements FruitStorageService {
+
```
Andriy-dz, line 11: empty line

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.ReportService;
+
+public class ReportServiceImpl implements ReportService {
+
```
Andriy-dz, line 7: empty line

src/main/java/core/basesyntax/strategy/impl/ReturnOperationHandler.java
```diff
@@ -0,0 +1,19 @@
...
+import core.basesyntax.exception.InvalidDataException;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.strategy.OperationHandler;
+
+public class ReturnOperationHandler implements OperationHandler {
+    @Override
+    public void handleTransaction(FruitTransaction transaction) {
+        String key = transaction.getFruit();
+        if (!Storage.DATABASE.containsKey(key)) {
+            //just some logic
```
Andriy-dz, line 13:
```suggestion

```

src/main/java/core/basesyntax/service/FruitStorageService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface FruitStorageService {
```
Andriy-dz, line 6:
```suggestion
public interface FruitService {

```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,49 @@
...
+    public static final String PATH_TO_REPORT_FILE = "src/main/java/resources/report.csv";
+
+    public static final Map<Operation, OperationHandler> HANDLER_MAP = new HashMap<>();
+
+    static {
+        HANDLER_MAP.put(Operation.BALANCE, new BalanceOperationHandler());
+        HANDLER_MAP.put(Operation.PURCHASE, new PurchaseOperationHandler());
+        HANDLER_MAP.put(Operation.RETURN, new ReturnOperationHandler());
+        HANDLER_MAP.put(Operation.SUPPLY, new SupplyOperationHandler());
+    }
```
Andriy-dz, line 34: I think you don't need this class field. It is better to create a private method that will return such a Map when needed

Andriy-dz (approved):

PR #937, opened 2023-09-19
create fruit shop

DenisGorda:


RomanYuzviak (changes requested):
Great solution, but you can fix a few things

PR #943, opened 2023-10-02
add interfaces & implementations for Fruit shop project

VoitseshukV:


src/main/java/core/basesyntax/service/impl/DataParserImpl.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.errors.DataParserError;
+import core.basesyntax.model.Operation;
+import core.basesyntax.model.OperationType;
+import core.basesyntax.service.DataParser;
+
+public class DataParserImpl implements DataParser {
+    private static final String SEPARATOR = ",";
```
okuzan, line 9: extract indexes to constants with meaningful names

src/main/java/core/basesyntax/model/OperationType.java
```diff
@@ -0,0 +1,27 @@
...
+        return code;
+    }
+
+    public static OperationType fromString(String text) {
+        for (OperationType type : OperationType.values()) {
+            if (type.code.equalsIgnoreCase(text)) {
+                return type;
+            }
+        }
+        return null;
```
okuzan, line 25: it's generally preferred to throw an exception (e.g. IllegalArgumentException) than return a null silently if that's now an expected behavior

src/main/java/core/basesyntax/db/FruitDaoStorageInMap.java
```diff
@@ -0,0 +1,50 @@
+package core.basesyntax.db;
+
+import core.basesyntax.errors.DaoError;
+import java.util.HashMap;
+import java.util.Map;
+
+public class FruitDaoStorageInMap implements FruitDao {
```
okuzan, line 7:
```suggestion
public class FruitStorageDaoImpl implements FruitStorageDao {
```

src/main/java/core/basesyntax/FruitShop.java
```diff
@@ -0,0 +1,60 @@
...
+        OperationStrategy operationStrategy = new OperationStrategyImpl(strategies);
+        FruitDao fruitDb = new FruitDaoStorageInMap();
+
+        // Read data
+        DataReader dataReader = new DataReaderCsv();
+        String[] data = dataReader.read(DATA_FILE_NAME);
+
+        // Parse & put in DB
+        DataParser dataParser = new DataParserImpl();
+        for (int i = 1; i < data.length; i++) {
```
okuzan, line 45:
```suggestion
        for (int i = OFFSET; i < data.length; i++) {
```

okuzan (changes requested):
Clean code and a good understanding of the strategy pattern! Good job 👏

PR #956, opened 2023-10-17
Jv fruit shop solution

gordgelin:
Initial implementation for review


okuzan (changes requested):
Overall, you've got a cool and very (maybe even too much) OOP implementation with creative solutions! But code is sometimes hard to navigate and make sense of - for example, having VirtualStorage ,VirtualStorageDao, Storage, StorageDao, StorageService, 
VirtualStorageService is a big overengineering and makes your solution hard to read. You basically can leave just one of them - some storageDao with hashmap constant as a field and accessor methods. Review your naming of classes and methods - it should be clear and concise. For example, in this application Dto and model are basically the same thing. Put test files inside of src/main/resources.

gordgelin:
cool, thnx )) overengineering was the underlying idea ) it is to create more like universal solution with possibilities to implement different extensions (f.e. more complex strategies, services, data sources and storages) in future if needed and not just like mock DB MVP to pass review. So for now i would like to stick with Service -> DAO -> Storage structure and leave Operation as it is if possible.

src/main/java/core/basesyntax/db/dto/StorageOperationDto.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax.db.dto;
+
+public class StorageOperationDto {
+    private final String type;
+    private final StorageItemDto storageItem;
```
olekskov, line 5: I'd remove Dto from naming.
DTO is used when we accept objects from outside or passing objects to external user in order to not to be tight to internal implementation and possibly hide some aspects of it.
Here we reading file by ourselves and there is no reason to wrap anything and by fact it's just a model

olekskov (approved):
Good job, approved.
Minor comment left for consideration

PR #969, opened 2023-10-31
realised main logic of all classes

fedorovychh:
1. Storage classes ad DB and their logic
2. Dao for DB
3. Models
4. Services for program logic
5. Strategy pattern using
6. Custom exceptions for handling problems with returning products and purchasing it


src/main/java/core/basesyntax/service/FileReaderService.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.service;
+
+public interface FileReaderService {
+    String readInputData();
```
Sofasmile, line 4: pass fileName to the methid

src/main/java/core/basesyntax/service/OperationService.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.service;
+
+public interface OperationService {
+    void manageStorageCells(String operationCode, String fruitName, int amount);
```
Sofasmile, line 4: use FruitTransaction or FruitData in the method parameters
```suggestion
    void manageStorageCells(FruitData fruitData);

```

src/main/java/core/basesyntax/service/ReportGeneratorService.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.service;
+
+public interface ReportGeneratorService {
+    public void generateReport(String report);
+}
```
Sofasmile, line 5:
```suggestion
public interface FileWriterService {
    public void write(String report, String fileName);
}

```

src/main/java/core/basesyntax/service/impl/DataHandlerServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+
+public class DataHandlerServiceImpl implements DataHandlerService {
+    private static final String REPORT_COLUMNS = "fruit,quantity";
+    private static final int INDEX_OF_REPORT_COLUMNS = 1;
+    private static final String SEPARATE_SYMBOL = ",";
+    private final List<StorageUpdateStrategy> storageUpdateStrategies = createServicesList();
+    private final OperationService operationService =
+            new OperationServiceImpl(storageUpdateStrategies);
+    private final FruitDao fruitDao = new FruitDaoImpl();
```
Sofasmile, line 23: initialize fields in constructor

src/main/java/core/basesyntax/service/impl/FileReaderServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public String readInputData() {
+        StringBuilder fileInputBuilder = new StringBuilder();
+        try (FileInputStream dataStream = new FileInputStream(INPUT_FILE_NAME)) {
+            int charInAscii;
+            while ((charInAscii = dataStream.read()) != READ_END_INDEX) {
+                fileInputBuilder.append((char) charInAscii);
+            }
+        } catch (IOException ex) {
+            throw new RuntimeException(String.format(READER_FAILURE_MESSAGE, INPUT_FILE_NAME));
```
Sofasmile, line 21:
```suggestion
            throw new RuntimeException(String.format(READER_FAILURE_MESSAGE, INPUT_FILE_NAME), ex);

```

src/main/java/core/basesyntax/service/impl/OperationServiceImpl.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.OperationService;
+import core.basesyntax.strategy.StorageUpdateStrategy;
+import java.util.List;
+
+public class OperationServiceImpl implements OperationService {
+    private final List<StorageUpdateStrategy> servicesList;
```
Sofasmile, line 8:
```suggestion
public class OperationStrategyImpl implements OperationStrategy {
    private final List<StorageUpdateHandler> handlers;

```

src/main/java/core/basesyntax/strategy/impl/FruitBalanceStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Override
+    public void updateStorage(String fruitName, int amount) {
+        fruitDao.addFirst(fruitName, amount);
+        balanceDao.add(fruitName, amount);
+    }
+
+    @Override
+    public boolean isServiceApplicable(String operationCode) {
+        return operationCode.equals(Operation.BALANCE.getCode());
```
Sofasmile, line 22:
```suggestion
        return operationCode == Operation.BALANCE.getCode();

```

Sofasmile (changes requested):

mnyshenko (changes requested):
A few points to consider

proxychang:
> A few points to consider

mnyshenko (approved):
Approved, but I want to pay your attention to two moments:
1) You should pass instances to the constructor and then assign them to fields. For example, with the current implementation, you've created a few instances of dao instead one instance, that you can pass everywhere and reuse it.
2) I would recommend you reconsider your fruitStorage implementation. Because of the structure you chose, you cannot get all the benefits of the map structure to quickly search for data inside it. For example, when you search the fruit in the db - you iterate by for each through all the entries in the map, which has O(n) time complexity.

Possible solution to the second problem would be the following:
1) Store fruit number in the fruit class
2) Map<String, Fruit>, as a key in this case you should use fruit name which you'll get from the fruit.

PR #976, opened 2023-11-01
fruit-shop solution

OlexVoit:


src/main/java/core/basesyntax/FruitShopApp.java
```diff
@@ -0,0 +1,45 @@
+package core.basesyntax;
```
BolsunovaNataliia, line 1: Interesting job. But I have a couple of questions for the author.
dao - data access object. What object are you looking at in the reader and writer?

src/main/java/core/basesyntax/FruitShopApp.java
```diff
@@ -0,0 +1,45 @@
...
+import core.basesyntax.transactions.impl.PurchaseOperationHandler;
+import core.basesyntax.transactions.impl.ReturnOperationHandler;
+import core.basesyntax.transactions.impl.SupplyOperationHandler;
+import java.util.List;
+import java.util.Map;
+
+public class FruitShopApp {
+    public static void main(String[] args) {
+        CsvFileReader readFile = new CsvFileReaderImpl();
+        List<String> readAllLines = readFile.read("input");
```
BolsunovaNataliia, line 24: "input", "output" should have a more real meaning. And must be replaced by constants.

src/main/java/core/basesyntax/FruitShopApp.java
```diff
@@ -0,0 +1,45 @@
...
+        FruitMap fruitMap = new FruitMap();
+        List<FruitTransaction> fruitTransactions = fruitMap.mapData(readAllLines);
+
+        Map<Operation, OperationHandler> transactionStrategies = Map.of(
```
BolsunovaNataliia, line 28: It's only my opinion, but why strategy element present in main method?

src/main/java/core/basesyntax/FruitShopApp.java
```diff
@@ -0,0 +1,45 @@
...
+                Operation.PURCHASE, new PurchaseOperationHandler(),
+                Operation.RETURN, new ReturnOperationHandler()
+        );
+
+        OperationStrategy operationStrategy = new OperationStrategy(transactionStrategies);
+        for (FruitTransaction fruitTransaction : fruitTransactions) {
+            operationStrategy.operation(fruitTransaction);
+        }
+        ReportService report = new ReportService();
+        String writePath = "output";
```
BolsunovaNataliia, line 40: It's the same ("output")

src/main/java/core/basesyntax/dao/reads/CsvFileReaderImpl.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final String EXCEPTION_TEXT = "Error while reading a file";
+
+    @Override
+    public List<String> read(String filePath) {
+        List<String> transactions = new ArrayList<>();
+        try (BufferedReader br = new BufferedReader(new FileReader(filePath))) {
+            String line;
+            br.readLine();
+            while ((line = br.readLine()) != null) {
+                String trimmedLine = line.trim().replaceAll("\\s+", "");
```
BolsunovaNataliia, line 19: Let's make the regex a constant.

src/main/java/core/basesyntax/dao/writes/CsvFileWriterOrder.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.dao.writes;
+
+public interface CsvFileWriterOrder {
```
BolsunovaNataliia, line 3: As for me, not a very clear name

src/main/java/core/basesyntax/model/Operation.java
```diff
@@ -0,0 +1,27 @@
...
+        return code;
+    }
+
+    public static Operation getOperationByCode(String code) {
+        for (Operation operation : values()) {
+            if (operation.getCode().equals(code)) {
+                return operation;
+            }
+        }
+        throw new IllegalArgumentException("Unable to find an operation for the code: " + code);
```
BolsunovaNataliia, line 25: This message should be replaced by constant too.

src/main/java/core/basesyntax/services/FruitMap.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax.services;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.Operation;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FruitMap {
+    private static final String SYMBOL = ",";
```
BolsunovaNataliia, line 9: If you name this symbol more specifically, the code will be more readable. Something like REGEX_COMMA

src/main/java/core/basesyntax/services/ReportService.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.services;
+
+import java.util.Map;
+
+public class ReportService {
+    private static final String FIRST_LINE = "fruit,quantity";
+    private static final String SYMBOL = ",";
```
BolsunovaNataliia, line 7: The same.

BolsunovaNataliia (approved):

mnyshenko (changes requested):
DAO: It's an object for accessing data from your storage. Your date storage is Storage.java, so any component (handlers/reporter) that wants to interact with that map should do so through this special DAO. You should implement DAO to **abstract invokers from the implementation**. With the current implementation, callers know that you have a map as internal storage. Let's imagine that tommorow you decide to use real database as data storage, so you'll have to go through all the callers (handlers/reporter in your case) and change behaviour there. DAO eliminates this problem, because you'll only need to change DAO class, and nothing will change from callers perspective - they just interact with object **that know how to interact with data storage**.

+ reader and writer should not be stored in this class. Yes, they work with some data, but this is not our data storage

mnyshenko (approved):
Good job

PR #977, opened 2023-11-01
Create fruit shop structure without strategy pattern (not done yet)

MurphyEdward:
This version is not done yet but already works properly. Pushed to register progress on the Mate Academy site.

### DIFF
--- a/src/main/java/main/Main.java
+++ b/src/main/java/main/Main.java
@@ -1,0 +1,13 @@
+package main;
+
+import service.impl.Reader;
+import service.impl.Writer;
+
+public class Main {
+    public static void main(String[] args) {
+        Reader reader = new Reader();
+        reader.read("data.csv");
+        Writer writer = new Writer();
+        writer.write("newData.csv");
+    }
+}
--- a/data.csv
+++ b/data.csv
@@ -1,0 +1,9 @@
+type,fruit,quantity
+b,banana,20
+b,apple,100
+s,banana,100
+p,banana,13
+r,apple,10
+p,apple,20
+p,banana,5
+s,banana,50
--- a/src/main/java/activity/ActivityHandler.java
+++ b/src/main/java/activity/ActivityHandler.java
@@ -1,0 +1,5 @@
+package activity;
+
+public interface ActivityHandler {
+    void accept(String fruit, int number);
+}
--- a/src/main/java/activity/impl/Balance.java
+++ b/src/main/java/activity/impl/Balance.java
@@ -1,0 +1,11 @@
+package activity.impl;
+
+import activity.ActivityHandler;
+import db.Storage;
+
+public class Balance implements ActivityHandler {
+    @Override
+    public void accept(String fruit, int amount) {
+        Storage.setAmount(fruit, amount);
+    }
+}
--- a/src/main/java/activity/impl/Purchase.java
+++ b/src/main/java/activity/impl/Purchase.java
@@ -1,0 +1,12 @@
+package activity.impl;
+
+import activity.ActivityHandler;
+import db.Storage;
+
+public class Purchase implements ActivityHandler {
+
+    @Override
+    public void accept(String fruit, int number) {
+        Storage.addFruit(fruit, -number);
+    }
+}
--- a/src/main/java/activity/impl/Return.java
+++ b/src/main/java/activity/impl/Return.java
@@ -1,0 +1,11 @@
+package activity.impl;
+
+import activity.ActivityHandler;
+import db.Storage;
+
+public class Return implements ActivityHandler {
+    @Override
+    public void accept(String fruit, int number) {
+        Storage.addFruit(fruit, number);
+    }
+}
--- a/src/main/java/activity/impl/Supply.java
+++ b/src/main/java/activity/impl/Supply.java
@@ -1,0 +1,11 @@
+package activity.impl;
+
+import activity.ActivityHandler;
+import db.Storage;
+
+public class Supply implements ActivityHandler {
+    @Override
+    public void accept(String fruit, int number) {
+        Storage.addFruit(fruit, number);
+    }
+}
--- a/src/main/java/core/basesyntax/HelloWorld.java
+++ b/src/main/java/core/basesyntax/HelloWorld.java
@@ -1,8 +1,0 @@
-package core.basesyntax;
-
-/**
- * Feel free to remove this class and create your own.
- */
-public class HelloWorld {
-
-}
--- a/src/main/java/db/Storage.java
+++ b/src/main/java/db/Storage.java
@@ -1,0 +1,29 @@
+package db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    private static Map<String, Integer> storageMap = new HashMap<>();
+
+    public static void addFruit(String fruit, Integer number) {
+        storageMap.put(fruit, number + storageMap.get(fruit));
+    }
+
+    public static Map<String, Integer> getFruitsAndAmount() {
+        Map<String, Integer> newMap = new HashMap<>();
+        for (Map.Entry<String, Integer> entry : storageMap.entrySet()) {
+            newMap.put(entry.getKey(), entry.getValue());
+        }
+
+        return newMap;
+    }
+
+    public static void getAmount(String fruit) {
+        storageMap.get(fruit);
+    }
+
+    public static void setAmount(String fruit, Integer number) {
+        storageMap.put(fruit, number);
+    }
+}
--- a/src/main/java/service/impl/Reader.java
+++ b/src/main/java/service/impl/Reader.java
@@ -1,0 +1,51 @@
+package service.impl;
+
+import activity.impl.Balance;
+import activity.impl.Purchase;
+import activity.impl.Return;
+import activity.impl.Supply;
+import java.io.BufferedReader;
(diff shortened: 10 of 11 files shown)


kozhukhovsky (changes requested):
In this task it is important to create a proper architecture. Your main task is to divide logic, and create separate independent services for each step of the whole process.

- FileReader  to read data about transactions from file

- TransactionParser  parses list of rows from file into list of FruitTransaction class

- OperationHandler  and an implementation for each operation type, each OperationHandler should make changes to the storage

- Storage class - that should have Map<String, Integer>. This class is a simplier DB. Key of the map is a fruit name, and value is amount of that fruit that is currently in our storage

- OperationStrategy that should have a map of handlers, and should return a handler by operation type

- TransactionProcessor is a service that should accept list of transactions and process them all

- ReportCreator should create a report from values that are in storage

- FileWriter writes a string value to a file



This task is to practice SOLID principles:

You should create separate services that have its own responsibilities

You are creating interfaces - defining functionality, and creating an implementation.

For example: TransactionParser is an abstraction, you should create an implementation CsvTransactionParser which parses transactions from SCV format. In the future, for other formats such as pdf, xlsx, etc. you would need to create a new implementation XlsxTransactionParser. There would be no need to modify existing code.

When your service needs some dependency you should pass from the outside, and be dependant on abstractions.

src/main/java/main/Main.java
```diff
@@ -0,0 +1,38 @@
...
+        operationHandlers.put("r", new Return());
+
+        FileReader reader = new FileReader();
+        TransactionParser csvTransactionParser = new CsvTransactionParser(operationStrategy);
+        TransactionProcessor transactionProcessorImpl = new TransactionProcessorImpl();
+        csvTransactionParser.parse(reader.read("data.csv"))
+                .forEach(transactionProcessorImpl::process);
+
+        FileWriter writer = new FileWriter();
+        writer.write(new ReportCreator().create(), "newData.csv");
```
Ivan95kos, line 36: use constants instead of Strings

src/main/java/operation/impl/Purchase.java
```diff
@@ -0,0 +1,13 @@
...
+
+import db.Storage;
+import model.FruitTransaction;
+import operation.OperationHandler;
+
+public class Purchase implements OperationHandler {
+
+    @Override
+    public void accept(FruitTransaction fruitTransaction) {
+        Storage.addFruit(fruitTransaction.fruit(), -fruitTransaction.amount());
```
Ivan95kos, line 11: You should check that you have enough fruits for the Purchase

src/main/java/operation/impl/Purchase.java
```diff
@@ -0,0 +1,13 @@
+package operation.impl;
+
+import db.Storage;
+import model.FruitTransaction;
+import operation.OperationHandler;
+
+public class Purchase implements OperationHandler {
```
Ivan95kos, line 7: Such a name makes it clear what the class is responsible for
refactor it everywhere
```suggestion
public class PurchaseOperationHandlerImpl implements OperationHandler {
```

src/main/java/service/ReportCreator.java
```diff
@@ -0,0 +1,22 @@
+package service;
+
+import db.Storage;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class ReportCreator {
```
Ivan95kos, line 8: You need to set the structure
```suggestion
public class ReportImpl implements Report {
```

src/main/java/service/ReportCreator.java
```diff
@@ -0,0 +1,22 @@
...
+    private static final String HEADER = "fruit,quantity";
+
+    public List<String> create() {
+        List<String> report = new ArrayList<>();
+
+        report.add(HEADER);
+
+        for (Map.Entry<String, Integer> entry : Storage.getFruitsAndAmount().entrySet()) {
+            report.add(entry.getKey() + "," + entry.getValue());
```
Ivan95kos, line 17:
```suggestion
            report.add(entry.getKey() + SEPARATOR + entry.getValue());
```

src/main/java/service/ReportCreator.java
```diff
@@ -0,0 +1,22 @@
...
+
+import db.Storage;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class ReportCreator {
+    private static final String HEADER = "fruit,quantity";
+
+    public List<String> create() {
```
Ivan95kos, line 11: It is better to return the Report as String
```suggestion
    public String prepareReport() {
```

src/main/java/service/TransactionParser.java
```diff
@@ -0,0 +1,8 @@
+package service;
+
+import java.util.List;
+import model.FruitTransaction;
+
+public interface TransactionParser {
```
Ivan95kos, line 6: For structure use more general names
```suggestion
public interface Parser {
```

src/main/java/service/impl/CsvTransactionParser.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public CsvTransactionParser(OperationStrategy operationStrategy) {
+        this.operationStrategy = operationStrategy;
+    }
+
+    @Override
+    public List<FruitTransaction> parse(List<String> rowsList) {
+        List<FruitTransaction> fruitTransactions = new ArrayList<>();
+        for (String row : rowsList) {
+            String[] values = row.split(",");
```
Ivan95kos, line 21:
```suggestion
            String[] values = row.split(SEPARATOR);
```

src/main/java/service/impl/CsvTransactionParser.java
```diff
@@ -0,0 +1,31 @@
...
+            OperationHandler activity = operationStrategy.get(values[0]);
+            String fruit = values[1];
+            int amount = Integer.parseInt(values[2]);
```
Ivan95kos, line 24: use constants instead of numbers

src/main/java/service/impl/FileReader.java
```diff
@@ -0,0 +1,23 @@
+package service.impl;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FileReader {
```
Ivan95kos, line 8: you need have the structure like this:
```
public class ReaderServiceImpl implements ReaderService {
    @Override
    public List<String> readFromFile(String fromFilePath) {
        try {
            return Files.readAllLines(Path.of(fromFilePath));
        } catch (IOException e) {
            throw new RuntimeException("Can't read data from file: " + fromFilePath, e);
        }
    }
}
```

src/main/java/service/impl/FileWriter.java
```diff
@@ -0,0 +1,18 @@
+package service.impl;
+
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.util.List;
+
+public class FileWriter {
```
Ivan95kos, line 7: the same:
```
public class WriterServiceImpl implements WriterService {
    @Override
    public void writeToFile(String data, String outputPath) {
        try {
            Files.writeString(Path.of(outputPath), data);
        } catch (IOException e) {
            throw new RuntimeException("Can't write to file: " + outputPath, e);
        }
    }
}
```

src/main/java/model/FruitTransaction.java
```diff
@@ -0,0 +1,6 @@
+package model;
+
+import operation.OperationHandler;
+
+public record FruitTransaction(OperationHandler operation, String fruit, int amount) {
```
Ivan95kos, line 5: A model should have a state, not a behavior
```suggestion
public record FruitTransaction(Operation operation, String fruit, int amount) {
```


```
    public enum Operation {
        BALANCE("b"),
        SUPPLY("s"),
        PURCHASE("p"),
        RETURN("r");

        private final String operation;

        Operation(String operation) {
            this.operation = operation;
        }

        public String getOperation() {
            return operation;
        }

        public static Operation getByCode(String character) {
                if (operation.getOperation().equals(character)) {
                    return operation;
                }
            }
            throw new RuntimeException("Unknown operation");
        }
    }
```

Ivan95kos (changes requested):

src/main/java/operation/impl/PurchaseOperationHandlerImpl.java
```diff
@@ -0,0 +1,15 @@
...
+import operation.OperationHandler;
+
+public class PurchaseOperationHandlerImpl implements OperationHandler {
+
+    @Override
+    public void accept(FruitTransaction fruitTransaction) {
+        if (Storage.getFruitsAndAmount().get(fruitTransaction.fruit()) >= 0) {
+            Storage.addFruit(fruitTransaction.fruit(), -fruitTransaction.amount());
+        }
+    }
```
Ivan95kos, line 14:
```suggestion
    @Override
    public void accept(FruitTransaction fruitTransaction) {
        if (Storage.getFruitsAndAmount().get(fruitTransaction.fruit()) >= 0) {
            Storage.addFruit(fruitTransaction.fruit(), -fruitTransaction.amount());
        }
        throw new RuntimeException("Not enough " + fruitTransaction.fruit() + " for purchase!");
    }
```

Ivan95kos (approved):
Great job, with a minor comment

PR #980, opened 2023-11-02
implemented main logic to solution

vadymhrnk:


annton-shakur (changes requested):
Well done! I have left a few comments that you may want to look into.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    private static final String FILE_NAME_READ = "src/main/java/input.csv";
+    private static final String FILE_NAME_WRITE = "src/main/java/output.csv";
+
+    public static void main(String[] args) {
+        FileReaderService fileReaderService = new CsvFileReader();
+        List<String> linesInFiles = fileReaderService.readFromFile(FILE_NAME_READ);
```
mnyshenko, line 27: minor: `linesInFiles` -> `linesFromFile` or just `lines`

src/main/java/core/basesyntax/service/ReportService.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.service;
+
+public interface ReportService {
+    String getReport();
```
mnyshenko, line 4: minor: I would say generateReport will be a more proper name because get... name suggests that it may already be created

mnyshenko (approved):
Great job!

PR #999, opened 2023-11-21
jv-fruit-shop_solution

SosunovychB:


src/main/java/model/Operation.java
```diff
@@ -0,0 +1,27 @@
...
+        return code;
+    }
+
+    public static Operation getOperation(String string) {
+        for (Operation operation : Operation.values()) {
+            if (operation.getCode().equals(string)) {
+                return operation;
+            }
+        }
+        return null;
```
Rommelua, line 25: do not return null - throw exception

src/main/java/service/CreateReportService.java
```diff
@@ -0,0 +1,5 @@
+package service;
+
+public interface CreateReportService {
+    String report();
```
Rommelua, line 4:
```suggestion
public interface ReportService {
    String createReport();
```

src/main/java/impl/ReadFileImpl.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.List;
+import service.ReadFileService;
+
+public class ReadFileImpl implements ReadFileService {
+    @Override
+    public List<String> readFile(String filePath) {
+        List<String> file = new ArrayList<>();
+        try {
+            file = Files.readAllLines(Path.of(filePath));
+            return file;
```
kerrrusha, line 16:
```suggestion
            return Files.readAllLines(Path.of(filePath));
```

kerrrusha (approved):

PR #1001, opened 2023-11-24
Made a FruitShop report writer which takes info from one .csv file an…

maxymmusiienko:
…d writing to another .csv file. Created Dao interface implementation for working with file, used pattern strategy for following SOLID principles.


src/main/java/main/FruitStoreMain.java
```diff
@@ -0,0 +1,19 @@
...
+import core.basesyntax.db.CommandDao;
+import core.basesyntax.db.CommandDaoImpl;
+import model.FruitStorage;
+import model.FruitTransactionStorage;
+import service.FruitStoreService;
+
+public class FruitStoreMain {
+    public static void main(String[] args) {
+        CommandDao commandDao = new CommandDaoImpl();
+        FruitTransactionStorage fruitTransactionStorage = commandDao.read();
```
okuzan, line 12: This should just be List<FruitTransaction> . It's not a storage, naming is confusing. You don't need to store, Just parse input in to list, and then pass it to next service

src/main/java/service/FruitStoreService.java
```diff
@@ -0,0 +1,38 @@
...
+import strategy.ReturnStrategy;
+import strategy.SupplyStrategy;
+import strategy.TransactionStrategy;
+
+public class FruitStoreService {
+    private Map<FruitTransaction.Operation, TransactionStrategy> strategyMap;
+
+    public FruitStoreService() {
+        strategyMap = new HashMap<>();
+        strategyMap.put(FruitTransaction.Operation.BALANCE, new BalanceStrategy());
```
okuzan, line 19: use Map.of, pass as arguemnt instead

src/main/java/service/FruitStoreService.java
```diff
@@ -0,0 +1,38 @@
...
+            TransactionStrategy strategy = strategyMap.get(transaction.getOperation());
+            if (strategy != null) {
+                strategy.apply(fruitStorage, transaction);
+            }
+        }
+
+        return fruitStorage;
+    }
+}
+
```
okuzan, line 38:
```suggestion
```

src/main/java/service/FruitStoreService.java
```diff
@@ -0,0 +1,38 @@
...
+        strategyMap.put(FruitTransaction.Operation.PURCHASE, new PurchaseStrategy());
+        strategyMap.put(FruitTransaction.Operation.RETURN, new ReturnStrategy());
+    }
+
+    public FruitStorage processTransactions(FruitTransactionStorage transactions) {
+        FruitStorage fruitStorage = new FruitStorage();
+
+        for (FruitTransaction transaction : transactions.getFruitTransactionList()) {
+            TransactionStrategy strategy = strategyMap.get(transaction.getOperation());
+            if (strategy != null) {
```
okuzan, line 30: better throw exception if strategy not found

storage.csv
```diff
@@ -0,0 +1,9 @@
+    type,fruit,quantity
```
okuzan, line 1: src/main/resources

src/main/java/strategy/PurchaseStrategy.java
```diff
@@ -0,0 +1,12 @@
+package strategy;
+
+import model.FruitStorage;
+import model.FruitTransaction;
+
+public class PurchaseStrategy implements TransactionStrategy {
+    @Override
+    public void apply(FruitStorage fruitStorage, FruitTransaction transaction) {
```
okuzan, line 8: what if we don't have enough to sell

src/main/java/core/basesyntax/db/CommandDaoImpl.java
```diff
@@ -0,0 +1,61 @@
...
+        List<FruitTransaction> transactions = new ArrayList<>();
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(INPUT_FILE_NAME))) {
+            reader.readLine();
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] parts = line.split(SPLIT_REGEX);
+                FruitTransaction transaction = new FruitTransaction();
+                transaction
+                        .setOperation(FruitTransaction
```
okuzan, line 35: separate your concerns. Create FileReader which only reads, and another service to parse the data into models. Think about validation - what if will get NumberFormatException from .parseInt or enum code won't be found?

okuzan (changes requested):
Good job! Read the comments to find out how to improve your code 👇

maxymmusiienko:
> Good job! Read the comments to find out how to improve your code 👇

thank you! Frankly speaking it was a difficult task for me. I will fix all the mistakes

src/main/java/service/ValidationDataException.java
```diff
@@ -0,0 +1,7 @@
+package service;
```
olekskov, line 1: why exception is in service package?

src/main/java/service/StrategyException.java
```diff
@@ -0,0 +1,7 @@
+package service;
```
olekskov, line 1: same

src/main/java/service/InsufficientInventoryException.java
```diff
@@ -0,0 +1,7 @@
+package service;
```
olekskov, line 1: same

src/main/java/service/FruitStoreService.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import java.util.Map;
+import model.FruitTransaction;
+import strategy.BalanceStrategy;
+import strategy.PurchaseStrategy;
+import strategy.ReturnStrategy;
+import strategy.SupplyStrategy;
+import strategy.TransactionStrategy;
+
+public class FruitStoreService {
```
olekskov, line 13: why reader and writer implement interface but this service does not?

src/main/java/service/FruitStoreService.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public FruitStoreService() {
+        strategyMap = Map.of(FruitTransaction.Operation.BALANCE, new BalanceStrategy(),
+                FruitTransaction.Operation.SUPPLY, new SupplyStrategy(),
+                FruitTransaction.Operation.PURCHASE, new PurchaseStrategy(),
+                FruitTransaction.Operation.RETURN, new ReturnStrategy());
+    }
+
+    public FruitStorage processTransactions(List<FruitTransaction> transactions) {
+        FruitStorage fruitStorage = new FruitStorage();
```
olekskov, line 24: In order to have flexible solution you should obtain things you rely on in constructor. So your service can be created to work with any map and any storage that may be created in future:
```suggestion
    public FruitStoreService(Map<FruitTransaction.Operation, TransactionStrategy> strategyMap, FruitStorage fruitStorage ) {
        this.strategyMap = strategyMap;
        this.fruitStorage = fruitStorage;
    }

    public FruitStorage processTransactions(List<FruitTransaction> transactions) {
```
map and fruitStorage should be initialized in main class and passed as arguments

olekskov (changes requested):

src/main/java/db/FruitStorage.java
```diff
@@ -0,0 +1,33 @@
...
+    public FruitStorage() {
+        fruitInventory = new HashMap<>();
+    }
+
+    public FruitStorage(Map<String, Integer> fruitInventory) {
+        this.fruitInventory = fruitInventory;
+    }
+
+    public Set<Map.Entry<String, Integer>> getAllFruitEntries() {
+        return fruitInventory.entrySet();
```
Sanderfuck, line 19: Why do you use this type?

src/main/java/service/FileReaderImpl.java
```diff
@@ -0,0 +1,60 @@
...
+                }
+
+                transactions.add(transaction);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file: " + inputFileName);
+        }
+
+        return transactions;
+    }
```
Sanderfuck, line 59: Single responsibility!!! Method read must only read! Divide this logic

src/main/java/service/FileWriterImpl.java
```diff
@@ -0,0 +1,24 @@
...
+    private static final String REPORT_FIRST_ROW = "fruit,quantity" + System.lineSeparator();
+    private static final String SPLIT_REGEX = ",";
+
+    public void write(FruitStorage fruitInventory, String reportFileName) {
+        try (BufferedWriter writer = new BufferedWriter(new java.io.FileWriter(reportFileName))) {
+            writer.write(REPORT_FIRST_ROW);
+            for (Map.Entry<String, Integer> entry : fruitInventory.getAllFruitEntries()) {
+                writer.write(entry.getKey() + SPLIT_REGEX
+                        + entry.getValue()
+                        + System.lineSeparator());
```
Sanderfuck, line 18: Better use stringBuilder

Sanderfuck (changes requested):

src/main/java/service/FileReaderImpl.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private FruitTransaction extractTransaction(String line) {
```
okuzan, line 34: The logic is fine, but you should place it in a separate service. The reader should return raw content, like List of String, not a list of DTOs. Then you pass it to FruitTransactionParser (you may name it differently) and it returns a list of transactions. That's what we meant by the separation of concerns

src/main/java/db/FruitStorage.java
okuzan: for report generation

okuzan (changes requested):

okuzan (approved):
Good job! Congratulations 🥳

PR #1028, opened 2023-12-28
Implemented Fruit Shop application.

davydov-software:


src/test/java/core/basesyntax/service/impl/CsvFileReportWriterServiceTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class CsvFileReportWriterServiceTest {
```
olekskov, line 18: Please do not be in a hurry with tests.
They will be in the next task. This one doesn't require them.
For the moment we should agree on code structure

src/main/java/core/basesyntax/service/impl/CsvFileTransactionReaderService.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CsvFileTransactionReaderService implements TransactionReaderService {
+    private final String fileName;
+
+    public CsvFileTransactionReaderService(String fileName) {
+        this.fileName = fileName;
+    }
```
olekskov, line 16: It's not good to create service that can read only 1 predefined file.
Please make filename method argument, not constructor

src/main/java/core/basesyntax/service/impl/CsvFileTransactionReaderService.java
```diff
@@ -0,0 +1,37 @@
...
+
+import core.basesyntax.service.TransactionReaderService;
+import java.io.BufferedReader;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CsvFileTransactionReaderService implements TransactionReaderService {
```
olekskov, line 11: This class doesn't have any specific logic to transactions or csv, it just reads lines from file
If you pass to it file with any other information or in different format it would produce result
```suggestion
public class FileReaderImpl implements FileReader {
```

src/main/java/core/basesyntax/service/impl/CsvFileReportWriterService.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.IOException;
+import java.util.List;
+
+public class CsvFileReportWriterService implements ReportWriterService {
+    private static final String HEADER = "fruit,quantity" + System.lineSeparator();
+    private final String filename;
+
+    public CsvFileReportWriterService(String filename) {
+        this.filename = filename;
+    }
```
olekskov, line 16: same as for reader:
1. complicated specific name for class
2. filename should be method argument

src/main/java/core/basesyntax/strategy/TransactionProcessorServiceStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class TransactionProcessorServiceStrategy {
+    private final Map<Operation, TransactionProcessorService> map = new HashMap<>();
+
+    public TransactionProcessorServiceStrategy(FruitDb db) {
+        TransactionProcessorService incoming = new IncomingTransactionProcessorService(db);
+
+        map.put(Operation.BALANCE, incoming);
```
olekskov, line 17: balance is revision operation, it should replace current quantity, not be added to it

src/main/java/core/basesyntax/strategy/TransactionProcessorServiceStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+        map.put(Operation.SUPPLY, incoming);
+        map.put(Operation.RETURN, incoming);
+        map.put(Operation.PURCHASE, new OutgoingTransactionProcessorService(db));
```
olekskov, line 20: Please recheck naming through project, there is no need to add service everywhere. Try to make it understandable but simple
```suggestion
        map.put(Operation.PURCHASE, new OutgoingTransactionProcessor(db));
```

olekskov (changes requested):

PR #1038, opened 2024-01-15
Solution

suprun-anna:
created fruit, storage, dao, fileReader, fileWriter, fruitService, strategies, reportService

### DIFF
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,0 +1,48 @@
+package core.basesyntax;
+
+import core.basesyntax.dao.FruitDao;
+import core.basesyntax.dao.FruitDaoImpl;
+import core.basesyntax.model.Operation;
+import core.basesyntax.service.FileReaderService;
+import core.basesyntax.service.FileWriterService;
+import core.basesyntax.service.FruitService;
+import core.basesyntax.service.ReportService;
+import core.basesyntax.service.impl.FileReaderServiceImpl;
+import core.basesyntax.service.impl.FileWriterServiceImpl;
+import core.basesyntax.service.impl.FruitServiceImpl;
+import core.basesyntax.service.impl.ReportServiceImpl;
+import core.basesyntax.strategy.FruitOperationStrategy;
+import core.basesyntax.strategy.impl.BalanceOperationStrategy;
+import core.basesyntax.strategy.impl.DecreaseQuantityStrategy;
+import core.basesyntax.strategy.impl.IncreaseQuantityStrategy;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    public static void main(String[] args) {
+        final String storeFileName = "src/main/resources/store.csv";
+        final String reportFileName = "src/main/resources/report.csv";
+
+        FileReaderService fileReaderService = new FileReaderServiceImpl();
+        final List<String> fruitData = fileReaderService.readFromFile(storeFileName);
+
+        FruitDao fruitDao = new FruitDaoImpl();
+        
+        Map<Operation, FruitOperationStrategy> strategyMap = new HashMap<>();
+        strategyMap.put(Operation.BALANCE, new BalanceOperationStrategy(fruitDao));
+        strategyMap.put(Operation.SUPPLY, new IncreaseQuantityStrategy(fruitDao));
+        strategyMap.put(Operation.RETURN, new IncreaseQuantityStrategy(fruitDao));
+        strategyMap.put(Operation.PURCHASE, new DecreaseQuantityStrategy(fruitDao));
+
+        FruitService fruitService = new FruitServiceImpl(strategyMap);
+        fruitService.fillStorage(fruitData);
+
+        ReportService reportService = new ReportServiceImpl();
+        String report = reportService.createReport(fruitDao.getAllFruits());
+
+        FileWriterService fileWriterService = new FileWriterServiceImpl();
+        fileWriterService.writeToFile(reportFileName, report);
+    }
+
+}
--- a/src/main/java/core/basesyntax/model/Fruit.java
+++ b/src/main/java/core/basesyntax/model/Fruit.java
@@ -1,0 +1,51 @@
+package core.basesyntax.model;
+
+import java.util.Objects;
+
+public class Fruit {
+    private String fruitName;
+    private int quantity;
+
+    public Fruit(String fruitName, int quantity) {
+        this.fruitName = fruitName;
+        this.quantity = quantity;
+    }
+
+    public String getFruitName() {
+        return fruitName;
+    }
+
+    public void setFruitName(String fruitName) {
+        this.fruitName = fruitName;
+    }
+
+    public int getQuantity() {
+        return quantity;
+    }
+
+    public void setQuantity(int quantity) {
+        this.quantity = quantity;
+    }
+
+    @Override
+    public String toString() {
+        return fruitName + "," + quantity;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Fruit fruit1 = (Fruit) o;
+        return quantity == fruit1.quantity && Objects.equals(fruitName, fruit1.fruitName);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(fruitName);
+    }
+}
--- a/src/main/java/core/basesyntax/service/FruitService.java
+++ b/src/main/java/core/basesyntax/service/FruitService.java
@@ -1,0 +1,7 @@
+package core.basesyntax.service;
+
+import java.util.List;
+
+public interface FruitService {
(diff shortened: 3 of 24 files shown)


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,48 @@
...
+        FruitDao fruitDao = new FruitDaoImpl();
+        
+        Map<Operation, FruitOperationStrategy> strategyMap = new HashMap<>();
+        strategyMap.put(Operation.BALANCE, new BalanceOperationStrategy(fruitDao));
+        strategyMap.put(Operation.SUPPLY, new IncreaseQuantityStrategy(fruitDao));
+        strategyMap.put(Operation.RETURN, new IncreaseQuantityStrategy(fruitDao));
+        strategyMap.put(Operation.PURCHASE, new DecreaseQuantityStrategy(fruitDao));
+
+        FruitService fruitService = new FruitServiceImpl(strategyMap);
+        fruitService.fillStorage(fruitData);
```
Rommelua, line 39: Convert step should be present in main() method
image

src/main/java/core/basesyntax/model/Fruit.java
```diff
@@ -0,0 +1,51 @@
+package core.basesyntax.model;
+
+import java.util.Objects;
+
+public class Fruit {
+    private String fruitName;
+    private int quantity;
```
Rommelua, line 7: quantity is not a Fruit characteristic. Fruit can not know its quantity. Storage knows it.
Actually we do not need Fruit class in this task because String fruitName is the only thing that we know about Fruit.
We can use Map<String, Integer> as Fruit Storage

src/main/java/core/basesyntax/service/FruitService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+import java.util.List;
+
+public interface FruitService {
+    void fillStorage(List<String> allTransactions);
```
Rommelua, line 6:
```suggestion
    void processTransactions(List<FruitTransaction> transactions);
```

src/main/java/core/basesyntax/service/impl/FileReaderServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+        try (BufferedReader reader = new BufferedReader(new FileReader(file))) {
+            String line = reader.readLine();
+            while (line != null) {
+                result.add(line);
+                line = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from file " + fileName);
+        }
+        removeFirstLine(result);
```
Rommelua, line 27: remove First Line in Parser
FileReaderService can be used for reading any files

src/main/java/core/basesyntax/service/impl/FileWriterServiceImpl.java
```diff
@@ -0,0 +1,17 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class FileWriterServiceImpl implements FileWriterService {
+    @Override
+    public void writeToFile(String fileName, String data) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(data);
+        } catch (IOException e) {
+            System.err.println("Can not write to file " + fileName);
```
Rommelua, line 14: throw new RuntimeException

Rommelua (changes requested):

src/main/java/core/basesyntax/strategy/FruitOperationStrategy.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.model.FruitTransaction;
+
+public interface FruitOperationStrategy {
+    void apply(FruitTransaction transaction);
+}
+
```
okuzan, line 8:
```suggestion
```

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.ReportService;
+import java.util.Map;
+
+public class ReportServiceImpl implements ReportService {
+    @Override
+    public String createReport(Map<String, Integer> fruits) {
+        StringBuilder report = new StringBuilder("fruit,quantity" + System.lineSeparator());
```
okuzan, line 9: -> REPORT_HEADER

src/main/java/core/basesyntax/service/impl/ParserImpl.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public FruitTransaction parse(String data) {
+        if (!data.matches("^[brsp],.*")) {
+            return null;
+        }
+        String[] strings = data.split(DELIMITER);
+        String fruitName = strings[NAME_INDEX];
+        int quantity = Integer.parseInt(strings[QUANTITY_INDEX]);
+        Operation operation = Operation
+                .fromCode(strings[OPERATION_INDEX]);
```
okuzan, line 22: Let's wrap this in a try-catch block since we might encounter trouble parsing malformed input. We should throw a runtime exception with a message indicating that this transaction is in the wrong format, along with the transaction details

okuzan: Also, move the logic of iterating the list of transactions from Main to this class, let's keep Main short

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,60 @@
...
+            }
+        }
+
+        FruitDao fruitDao = new FruitDaoImpl();
+        
+        Map<Operation, FruitOperationStrategy> strategyMap = new HashMap<>();
+        strategyMap.put(Operation.BALANCE, new BalanceOperationStrategy(fruitDao));
+        strategyMap.put(Operation.SUPPLY, new IncreaseQuantityStrategy(fruitDao));
+        strategyMap.put(Operation.RETURN, new IncreaseQuantityStrategy(fruitDao));
+        strategyMap.put(Operation.PURCHASE, new DecreaseQuantityStrategy(fruitDao));
```
okuzan, line 49: init using Map.of, make it a class constant

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,60 @@
...
+import core.basesyntax.strategy.impl.IncreaseQuantityStrategy;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    public static void main(String[] args) {
+        final String storeFileName = "src/main/resources/store.csv";
+        final String reportFileName = "src/main/resources/report.csv";
```
okuzan, line 29: make them constants
also, smth like `input` may be more suitable than `store` here

src/main/java/core/basesyntax/service/impl/ParserImpl.java
```diff
@@ -0,0 +1,25 @@
...
+public class ParserImpl implements Parser {
+    private static final String DELIMITER = ",";
+    private static final int OPERATION_INDEX = 0;
+    private static final int NAME_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public FruitTransaction parse(String data) {
+        if (!data.matches("^[brsp],.*")) {
+            return null;
```
okuzan, line 16: throwing an exception is a better practice than swallowing problem and returning null silently

src/main/java/core/basesyntax/service/impl/ParserImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class ParserImpl implements Parser {
+    private static final String DELIMITER = ",";
+    private static final int OPERATION_INDEX = 0;
+    private static final int NAME_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public FruitTransaction parse(String data) {
+        if (!data.matches("^[brsp],.*")) {
```
okuzan, line 15: regex validation is an interesting idea, let's do it fully, and extract PATTERN to a constant
```suggestion
        if (!data.matches("^[brsp],[a-zA-Z]+,[0-9]+$")) {
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,60 @@
...
+        FileReaderService fileReaderService = new FileReaderServiceImpl();
+        final List<String> fruitData = fileReaderService.readFromFile(storeFileName);
+
+        Parser parser = new ParserImpl();
+        List<FruitTransaction> fruitTransactions = new ArrayList<>();
+        for (String data : fruitData) {
+            FruitTransaction fruitTransaction = parser.parse(data);
+            if (fruitTransaction != null) {
+                fruitTransactions.add(fruitTransaction);
+            }
```
okuzan, line 40: remove this wrapper, we will throw exceptions instead

okuzan (changes requested):
Good job overall!

kerrrusha (approved):

PR #1039, opened 2024-01-16
hm for review

vdbond:
I didn't quite understand about storage and in which packages to place classes, so I hope you explain it to me by reviewing this PR


src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax.model;
+
+public record FruitTransaction(Operation operation,
+                               String fruit, int quantity) {
```
okuzan, line 4: add fields of transaction itself - fruit, quantity, constructor, getters setters etc

src/main/java/core/basesyntax/service/impl/CsvWriterServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+        StringBuilder stringBuilder = new StringBuilder("fruit,quantity" + System.lineSeparator());
+        map.forEach((key, value) -> stringBuilder.append(key)
+                .append(",")
+                .append(value)
+                .append(System.lineSeparator()));
+        String report = stringBuilder.toString();
+        try {
+            Files.writeString(toFileName.toPath(), report, StandardOpenOption.APPEND);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to file ", e);
```
okuzan, line 26: always include filename

src/main/java/core/basesyntax/service/impl/CsvWriterServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public void write(File toFileName, Map<String, BigDecimal> map) {
+        StringBuilder stringBuilder = new StringBuilder("fruit,quantity" + System.lineSeparator());
+        map.forEach((key, value) -> stringBuilder.append(key)
+                .append(",")
+                .append(value)
+                .append(System.lineSeparator()));
+        String report = stringBuilder.toString();
+        try {
+            Files.writeString(toFileName.toPath(), report, StandardOpenOption.APPEND);
```
okuzan, line 24: don't append - rewrite

okuzan (changes requested):
Storage should be a separate class. hashMap is a suitable data structure for it - key: fruit name, value: fruit quantity. You can just make it public, or provide Dao-like methods to access it. This class is a mock for DB
You probably need another Service to iterate transactions in a loop/stream and apply them to the storage, check out the diagram in task description for architecture hints
Put resources where they should be - `src/main/resources`

kozhukhovsky (approved):

PR #1042, opened 2024-01-20
created fruitShop.

RCherniavskiy:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,47 @@
...
+
+        System.out.println("\n");
+        fruitTransaction.balance("banana", 20);
+        fruitTransaction.balance("apple", 100);
+        fruitTransaction.supply("banana", 100);
+        fruitTransaction.purchase("banana", 13);
+        fruitTransaction.returnFruits("apple", 10);
+        fruitTransaction.purchase("apple", 20);
+        fruitTransaction.purchase("banana", 5);
+        fruitTransaction.supply("banana", 50);
```
olekskov, line 28: you should obtain data by reading input file, not hardcoding some operations

src/main/java/core/basesyntax/dao/FruitDao.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.List;
+
+public interface FruitDao {
+    void addCsvRow(Fruit fruitTransaction);
+
+    List<String> getCsv();
+
+    void createReportCsv(List<String> transactions);
+
+    public List<String> getReportCsv();
```
olekskov, line 13: Dao should have only methods that are related to the storage of internal data.
Basic setup of Dao is `create`, `read`, `update`, `delete` of records in our storage.
Please extract method that aimed to work with incoming file or report file to separate services like FileReader, FileWriter

src/main/java/core/basesyntax/service/impl/OperationReportImp.java
```diff
@@ -0,0 +1,57 @@
...
+    public void report() {
+        List<String> csvData = fruitDao.getCsv();
+
+        Map<String, Integer> resultMap = csvData.stream()
+                .map(s -> s.split(SEPARATOR))
+                .collect(Collectors.groupingBy(
+                        stringRow -> stringRow[FRUIT],
+                        Collectors.summingInt(stringRow -> {
+                            int quantity = Integer.parseInt(stringRow[QUANTITY]);
+                            return "p".equals(stringRow[OPERATION]) ? -quantity : quantity;
```
olekskov, line 38: Balance operation should override value of particular fruit. It's revision of quantity, not addition

src/main/java/core/basesyntax/service/strategy/PurchaseHandler.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.service.strategy;
+
+import core.basesyntax.dao.FruitDao;
+import core.basesyntax.model.Fruit;
+
+public class PurchaseHandler implements OperationHandler {
+    @Override
+    public void getOperation(FruitDao fruitDao, String fruit, int quantity) {
+        fruitDao.addCsvRow(new Fruit(Fruit.Operation.PURCHASE, fruit, quantity));
+    }
```
olekskov, line 10: There is no point to have different handlers that contain identical logic.
instead of call someHandler.getOperation(fruitDao, fruit, quantity) you can esaily call fruitDao.addCsvRow(...)

please try to delegate some logic to handlers that will be different.
For example calculation of quantity of some fruit depending on operation. As in one case it would be subtraction, in other addition, in another replacement.
subtraction and replacement also could have some validations not to produce negative balance of fruit after operation performed

src/main/java/core/basesyntax/service/strategy/BalanceHandler.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.service.strategy;
+
+import core.basesyntax.dao.FruitDao;
+import core.basesyntax.model.Fruit;
+
+public class BalanceHandler implements OperationHandler {
+    @Override
+    public void getOperation(FruitDao fruitDao, String fruit, int quantity) {
```
olekskov, line 8: 1) Please do not combine `void` and `get` in name. It's impossible to read such code.
If you use getOperation name that you expect that this method returns you some operation but it provides you nothing.
performOperation could be more appropriate but again you don't pass any operation, you pass some service, fruit and quatity, the is no operation anywhere around.
Due to that fact I don't have any idea what is going on here unless I deep-dive into several methods.

2) Do not pass services you rely on as method arguments. They should be class fields and initialized through constructor

olekskov (changes requested):

PR #1047, opened 2024-01-29
task completed

YaroslavBerezhniak:


src/main/java/main/Main.java
```diff
@@ -0,0 +1,57 @@
...
+            }
+            String[] parts = line.split(COMMA);
+            FruitTransaction.Operation operation = FruitTransaction.Operation
+                    .valueOfCode(parts[OPERATION_INDEX].trim());
+            String fruit = parts[FRUIT_INDEX];
+            int quantity = Integer.parseInt(parts[QUANTITY_INDEX]);
+
+            strategyFactory.getStrategy(operation)
+                    .execute(fruit, quantity, storage.getFruitQuantities());
+        }
```
olekskov, line 43: Please move this logic to separate service

src/main/java/main/Main.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private static List<String> generateReport(Map<String, Integer> fruitQuantities) {
+        List<String> report = new ArrayList<>();
+        report.add(REPORT_HEADER);
+        for (Map.Entry<String, Integer> entry : fruitQuantities.entrySet()) {
+            report.add(entry.getKey() + COMMA + entry.getValue());
+        }
+        return report;
+    }
```
olekskov, line 56: and this one also

src/main/java/service/FileWriterService.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.List;
+
+public class FileWriterService {
+    public void writeToFile(List<String> lines, String filePath) {
+        try (FileWriter writer = new FileWriter(filePath)) {
+            for (String line : lines) {
+                writer.write(line + "\n");
```
olekskov, line 11: Please use System.lineSeparator() instead of system dependant hardcoded symbol
```suggestion
                writer.write(line + System.lineSeparator());
```

src/main/java/strategy/OperationStrategyFactory.java
```diff
@@ -0,0 +1,21 @@
+package strategy;
+
+import java.util.HashMap;
+import java.util.Map;
+import model.FruitTransaction;
+
+public class OperationStrategyFactory {
```
olekskov, line 7: Strategy itself is pattern to provide some implementation, no need to add Factory to naming
```suggestion
public class OperationStrategy {
```

src/main/java/strategy/OperationStrategyFactory.java
```diff
@@ -0,0 +1,21 @@
...
+    private final Map<FruitTransaction.Operation, OperationStrategy> strategies;
+
+    public OperationStrategyFactory() {
+        strategies = new HashMap<>();
+        strategies.put(FruitTransaction.Operation.BALANCE, new BalanceOperationStrategy());
+        strategies.put(FruitTransaction.Operation.SUPPLY, new SupplyOperationStrategy());
+        strategies.put(FruitTransaction.Operation.PURCHASE, new PurchaseOperationStrategy());
+        strategies.put(FruitTransaction.Operation.RETURN, new ReturnOperationStrategy());
```
olekskov, line 15: And implementation provided is already not strategy. So better to change naming to represent what it's doing
some possible examples (but choose same naming pattern for all of them)
```suggestion
        strategies.put(FruitTransaction.Operation.BALANCE, new BalanceOperationExecutor());
        strategies.put(FruitTransaction.Operation.SUPPLY, new SupplyOperationHandler());
        strategies.put(FruitTransaction.Operation.PURCHASE, new PurchaseOperationFacilitator());
        strategies.put(FruitTransaction.Operation.RETURN, new ReturnOperationPerformer());
```

src/main/java/strategy/ReturnOperationStrategy.java
```diff
@@ -0,0 +1,10 @@
+package strategy;
+
+import java.util.Map;
+
+public class ReturnOperationStrategy implements OperationStrategy {
+    @Override
+    public void execute(String fruit, int quantity, Map<String, Integer> fruitQuantities) {
+        fruitQuantities.put(fruit, fruitQuantities.getOrDefault(fruit, 0) + quantity);
```
olekskov, line 8: please check usage of .merge() method instead of .put()
here it should look more pretty

src/main/java/strategy/OperationStrategyFactory.java
```diff
@@ -0,0 +1,21 @@
...
+    }
```
olekskov, line 16: please move map initialization to main method and in constructor just assign obtained map to variable

olekskov (changes requested):
Good job overall but requires some improvements

kozhukhovsky (approved):

PR #1057, opened 2024-02-09
firstVersionFruitShop

shumYevh:


src/main/java/db/FruitStorage.java
```diff
@@ -0,0 +1,12 @@
+package db;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+public class FruitStorage {
+    private static Map<String, Integer> fruits = new TreeMap<>();
```
olekskov, line 7: For get/put operations HashMap is better choise. TreeMap better if we have to do iterations by keys in sorted order
```suggestion
    private static Map<String, Integer> fruits = new HashMap<>();
```

src/main/java/strategy/FruitStorageHandler.java
```diff
@@ -0,0 +1,37 @@
...
+        StringBuilder report = new StringBuilder();
+        report.append(REPORT_HEAD);
+        for (Map.Entry entry : FRUIT_STORAGE.getFruits().entrySet()) {
+            report.append(System.lineSeparator())
+                    .append(entry.getKey())
+                    .append(",")
+                    .append(entry.getValue());
+        }
+        return report.toString();
+    }
```
olekskov, line 36: This method should be extracted to separate service. Overall this service looks like it's designed to work with storage and forming report is out of this scope.
Probably you should replace it with some getAll() operation and extract report formation to specialized component

src/main/java/strategy/FruitStorageHandler.java
```diff
@@ -0,0 +1,37 @@
+package strategy;
+
+import db.FruitStorage;
+import java.util.Map;
+
+public class FruitStorageHandler implements StorageHandler<String,Integer> {
+    private static final String REPORT_HEAD = "fruit,quantity";
+    private static final FruitStorage FRUIT_STORAGE = new FruitStorage();
```
olekskov, line 8: Components we rely on are better to be non-static and injected via constructor
```suggestion
    private final FruitStorage storage;
    
    public FruitStorageHandler(FruitStorage storage) {
        this.storage = storage;
    }
```

src/main/java/strategy/FruitStorageHandler.java
```diff
@@ -0,0 +1,37 @@
+package strategy;
+
+import db.FruitStorage;
+import java.util.Map;
+
+public class FruitStorageHandler implements StorageHandler<String,Integer> {
```
olekskov, line 6: By current design this class is much more connected to DB that to Strategy so it should be placed and named properly
```suggestion
package db;

import java.util.Map;

public class FruitStorageDao implements StorageHandler<String,Integer> {
```

src/main/java/strategy/operation/FruitBalanceOperation.java
```diff
@@ -0,0 +1,13 @@
+package strategy.operation;
+
+import strategy.FruitStorageHandler;
+import strategy.StorageHandler;
+
+public class FruitBalanceOperation implements OperationHandler<String, Integer> {
+    private static StorageHandler<String, Integer> STORAGE_HANDLER = new FruitStorageHandler();
```
olekskov, line 7: same as were above
private final and injected via constructor
please recheck all operations

olekskov (changes requested):
Some enhancements required but it's a very good job for first try!

src/main/java/Main.java
```diff
@@ -0,0 +1,57 @@
...
+import strategy.TransactionStrategy;
+import strategy.operation.FruitBalanceOperation;
+import strategy.operation.FruitPurchaseOperation;
+import strategy.operation.FruitReturnOperation;
+import strategy.operation.FruitSupplyOperation;
+import strategy.operation.OperationHandler;
+
+public class Main {
+    public static void main(String[] args) {
+        String nameReadingFile = "Fruits.csv";
```
okuzan, line 28: make filenames constants in this class

src/main/java/Main.java
```diff
@@ -0,0 +1,57 @@
...
+        FruitStorage fruitStorage = new FruitStorage();
+        StorageDao<String,Integer> storageHandler = new FruitStorageDao(fruitStorage);
+
+        Map<FruitTransaction.Operation,
+                OperationHandler<String, Integer>> operationHandlerMap = new HashMap<>();
+        operationHandlerMap.put(FruitTransaction.Operation.BALANCE,
+                new FruitBalanceOperation(storageHandler));
```
okuzan, line 35: init with Map.of for more concise code

src/main/java/service/impl/FruitCsvWriter.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class FruitCsvWriter implements Writer {
+    private static final String PATH_TO_WRITE = "src/main/resources/";
+
+    @Override
+    public void write(String file, String fileName) {
+        try {
+            FileWriter fileWriter = new FileWriter(PATH_TO_WRITE + fileName);
+            fileWriter.write(file);
+            fileWriter.close();
```
okuzan, line 15: use try with resources

src/main/java/service/impl/FruitOperationPerformer.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public FruitOperationPerformer(TransactionStrategy transactionStrategy) {
+        this.transactionStrategy = transactionStrategy;
+    }
+
+    @Override
+    public void performProcess(FruitTransaction transaction) {
+        OperationHandler handler = transactionStrategy.get(transaction.getOperation());
+        handler.doOperation(transaction.getFruit(), transaction.getQuantity());
+    }
```
okuzan, line 20: unused as of now
make it a private method, call it inside of a public method, there's some code duplication

src/main/java/service/impl/FruitTransactionParser.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private FruitTransaction.Operation validOperation(String codeOperation) {
+        for (int i = 0; i < VALIBLE_OPERATIONS.length; i++) {
+            if (VALIBLE_OPERATIONS[i].getCode().equals(codeOperation)) {
+                return VALIBLE_OPERATIONS[i];
+            }
+        }
+        throw new RuntimeException("Can't parse this Operation " + codeOperation);
+    }
```
okuzan, line 48: valible isn't a word
remove this constant
such methods are usually places inside of their enums, move it there

src/main/java/service/impl/FruitTransactionParser.java
```diff
@@ -0,0 +1,49 @@
...
+                transactions.add(new FruitTransaction(newOperation,
+                        nameOfFruit,
+                        quantityOfOperation));
+            }
+            return transactions;
+        }
+        throw new RuntimeException("Can't parse this List" + input);
+    }
+
+    private FruitTransaction.Operation validOperation(String codeOperation) {
```
okuzan, line 41: method names shouldn't be nouns -> `fromCode`, `of`

okuzan (changes requested):
Nicely done

Vasul-Pankiv (approved):

PR #1062, opened 2024-02-14
added solution to the task

yanna-stepanova:


src/main/java/Main.java
```diff
@@ -0,0 +1,46 @@
...
+public class Main {
+    private static final String FILE_INPUT = "src/main/resources/input.cvs";
+    private static final String FILE_OUTPUT = "src/main/resources/output.cvs";
+    private static final ReaderService readerService = new ReaderServiceImpl();
+    private static final ParserService parserService = new ParserServiceImpl();
+    private static final WriterService writerService = new WriterServiceImpl();
+    private static final Map<Operation, OperationHandler> operationHandlerMap = new HashMap<>();
+
+    public static void main(String[] args) {
+        List<String> dataReader = readerService.readFromFile(FILE_INPUT);
```
olekskov, line 31: dataReader name looks like some service name but this is just content of file represented as List<String>
```suggestion
        List<String> fileContent = readerService.readFromFile(FILE_INPUT);
```

src/main/java/Main.java
```diff
@@ -0,0 +1,46 @@
...
+        ActionDao actionDao = parserService.parseToObject(dataReader);
```
olekskov, line 32: parser is a service that should create some object from raw representation (often String)
you shouldn't delegate it any other additional logic
```suggestion
        List<FruitTransaction> fruitTransactions = parserService.parseTransactions(fileContent);
```

src/main/java/db/Storage.java
```diff
@@ -0,0 +1,21 @@
+package db;
+
+import java.util.ArrayList;
+import java.util.List;
+import model.FruitTransaction;
+
+public class Storage {
+    private final List<FruitTransaction> transactions = new ArrayList<>();
```
olekskov, line 8: Inside storage it's better to have structure that looks like storage.
You anyway goes to Map by collecting results in FruitService. Would be nicer to stick to it from the start
```suggestion
    private final Map<String, Integer> fruitQuantityByName = new HashMap<>();
```

src/main/java/Main.java
```diff
@@ -0,0 +1,46 @@
...
+        initMapOperationHandler();
+        OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlerMap);
+        FruitService fruitService = new FruitServiceImpl(actionDao, operationStrategy);
+        Map<String, Integer> report = fruitService.createReport();
```
olekskov, line 36: Processing of transactions and report generation are separate tasks that would be better to split into separate services.
Also report would be better if it's in format ready to write to file, in this way we can simplify writer
```suggestion
        FruitService fruitService = new FruitServiceImpl(operationStrategy);
        fruitService.processTransactions(fruitTransactions); // this method will change state of our storage
        ReportService reportService = new ReportService();
        String report = reportService.generateReport(); // and this will make String based on storage content
```

src/main/java/strategy/OperationHandler.java
```diff
@@ -0,0 +1,5 @@
+package strategy;
+
+public interface OperationHandler {
+    int getAmount(int quantity);
+}
```
olekskov, line 5: Handler should process transaction by changing storage state accordingly
```suggestion
public interface OperationHandler {
    void processTransaction(FruitTransaction transaction);
}
```

src/main/java/strategy/BalanceHandlerImpl.java
```diff
@@ -0,0 +1,13 @@
...
+public class BalanceHandlerImpl implements OperationHandler {
+    private static final int BORDER_POSITIVE = 0;
+
+    @Override
+    public int getAmount(int quantity) {
+        if (quantity < BORDER_POSITIVE) {
+            throw new RuntimeException("Balance can't be negative quantity!");
+        }
+        return quantity;
+    }
```
olekskov, line 12: and implementation would be something like
```suggestion
    @Override
    public void processTransaction(FruitTransaction transaction) {
        Storage.getStorage.put(transaction.getFruit(), transaction.getQuantity());
    }
```
Please recheck all handlers

olekskov (changes requested):

kozhukhovsky (approved):
Good job!

PR #1064, opened 2024-02-18
First Attempt. Implemented Fruit Shop

kacper-kukula:


src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    private static Map<String, Integer> fruitStorage = new HashMap<>();
```
nowickimj, line 7: This approach is correct according to mentor guidelines. 
However in real life, in order to provide a globally accessible storage service with dynamically changing internal state, you should follow the Singleton pattern:

1. Make constructors private - no one should be able to instantiate the singleton.
2. Store the only class instance internally as the private static field (T instance). The state can be maintained as a non-static private field within the instance - provide necessary accessors and mutators to singleton class (e.g. T.addFruit(...).
3. Create public static T getInstance() method that provides he internally stored instance:
```java
public static T getInstance() {
  if(instance == null) {
    instance = new T();
  }
  return instance;
}
```

src/main/java/core/basesyntax/service/impl/DataConverterServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+public class DataConverterServiceImpl implements DataConverterService {
+    private static final String SPLIT_REGEX = ",";
+    private static final String NOT_FOUND_ERR = "No such operation found.";
+    private static final int FIRST_INDEX = 0;
+    private static final int SECOND_INDEX = 1;
+    private static final int THIRD_INDEX = 2;
+
+    @Override
+    public List<FruitTransaction> getListOfTransactions(List<String> lines) {
+        Predicate<String> startsWithValidCode = new Predicate<>() {
```
nowickimj, line 21: `Predicate` is a functional interface, therefore you should use in a functional fashion rather than creating an anonymous class.

nowickimj (changes requested):
Small change requested, regarding the Predicate use, other comments are hints.

nowickimj (approved):

PR #1082, opened 2024-03-14
Hw jv fruit shop

EduarfF:


Bohadan-Markatov (approved):

src/main/java/core/basesyntax/dao/FruitDaoImpl.java
```diff
@@ -0,0 +1,44 @@
...
+                stringBuilder.append("\n");
+            }
+            return stringBuilder.toString();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read a file");
+        }
+    }
+
+    @Override
+    public void writeToFile(File file, List<String[]> data) {
```
karma-o, line 29: kind of weird to receive a list of string arrays to write, let's replace it with just a list of strings

src/main/java/core/basesyntax/service/FruitServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Map;
+
+public class FruitServiceImpl implements FruitService {
+    private static final int REMOVE_NAMES_INDEX = 0;
+    private static final int OPERATION_INDEX = 0;
+    private static final int FRUIT_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public List<String[]> report(String data) {
```
karma-o, line 16: this method and service has too many responsibilities:
* parses string into a transaction
* processes the data inside it
* saves it to somewhere
* builds a report

All these points should be separate services, like `TransactionParser`, `DataProcessor`(uses strategy, because I do not see one in your solution at all), `ReportProvider`, and called in main in respective order.
In general try to "dive deep" in the common mistakes and tips in the task description, especially in the diagram there. Ask any questions that you may have.
Do that first and then we will proceed with a more detailed review :wink:

src/main/java/core/basesyntax/HelloWorld.java
```diff
@@ -1,9 +1,25 @@
...
- */
+import core.basesyntax.dao.FruitDao;
+import core.basesyntax.dao.FruitDaoImpl;
+import core.basesyntax.service.FruitService;
+import core.basesyntax.service.FruitServiceImpl;
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+
 public class HelloWorld {
```
karma-o, line 11: we can also rename this class to something like FruitShop, but that is not important

karma-o (changes requested):
Check my comment. Dive into the common mistakes and diagram there, feel free to ask any questions

PR #1083, opened 2024-03-14
Ðimplemented classes to create a report

Bohadan-Markatov:


DmytroHadiuchko (changes requested):
Good job, I left some minor comments

src/main/java/core/basesyntax/service/impl/FruitTransactionProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+    public void process(FruitTransaction transaction) {
+        storage.add(transaction.getProductType(), strategy(transaction));
+    }
+
+    private int strategy(FruitTransaction transaction) {
+        return switch (transaction.getTransactionType()) {
+            case PURCHASE -> subtraction(transaction);
+            case RETURN, BALANCE, SUPPLY -> add(transaction);
+        };
+    }
```
karma-o, line 24: we should replace it with the strategy map of `<OperationType, OperationHandler>` check the theory and the example that Bohdan did and why is it better than the switch
Also that `OperationHandler` can just directly access the storage instead of returning something

src/main/java/core/basesyntax/model/Storage.java
```diff
@@ -0,0 +1,25 @@
...
+    public Storage() {
+        storage = new HashMap<>();
+    }
+
+    public void add(String key, Integer value) {
+        if (storage.containsKey(key)) {
+            storage.put(key, storage.get(key) + value);
+        } else {
+            storage.put(key, value);
+        }
```
karma-o, line 19: this is a business logic of addition which should not be here in the DAO layer of our application

src/main/java/core/basesyntax/model/Storage.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public Stream<Map.Entry<String, Integer>> stream() {
+        return storage.entrySet().stream();
+    }
```
karma-o, line 24: we do not need that, I think you can just make the map public final, and it should be fine, this is just a simulation of a DB tho
```suggestion
```

src/main/java/core/basesyntax/service/impl/CsvReportProvider.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private final Predicate<Integer> predicate = operationValue -> {
+        if (operationValue < 0) {
+            throw new RuntimeException("Operation value cannot be negative");
+        }
+        return true;
+    };
+
+    @Override
+    public String make(Storage storage) {
```
karma-o, line 20: we can also have a direct access to the `storage` here instead of passing it as a param

src/main/java/core/basesyntax/service/impl/CsvReportProvider.java
```diff
@@ -0,0 +1,28 @@
...
+public class CsvReportProvider implements ReportProvider {
+    public static final String REPORT_TITLE = "fruit,quantity";
+    public static final String COLUMN_SEPARATOR = ",";
+    public static final String ROW_SEPARATOR = System.lineSeparator();
+
+    private final Predicate<Integer> predicate = operationValue -> {
+        if (operationValue < 0) {
+            throw new RuntimeException("Operation value cannot be negative");
+        }
+        return true;
```
karma-o, line 16: bad variable name

src/main/java/core/basesyntax/service/impl/CsvReportProvider.java
```diff
@@ -0,0 +1,28 @@
...
+    };
+
+    @Override
+    public String make(Storage storage) {
+        StringBuilder builder = new StringBuilder(REPORT_TITLE + ROW_SEPARATOR);
+        storage.stream()
+                .filter(n -> predicate.test(n.getValue()))
```
karma-o, line 23: use `peek()` instead of filter, because we do not need to filter, but to check everything and throw an ex if the data is invalid

src/main/java/core/basesyntax/service/impl/CsvReportProvider.java
karma-o, line 20:
```suggestion
    public String provide(Storage storage) {
```

src/main/java/core/basesyntax/service/impl/CsvReportProvider.java
```diff
@@ -0,0 +1,28 @@
...
+                .map(n -> n.getKey() + COLUMN_SEPARATOR + n.getValue())
+                .forEach(n -> builder.append(n).append(ROW_SEPARATOR));
```
karma-o, line 25: You can use `.collect(Collectors.joining())` instead of this and the `StringBuilder`, it has a prefix, delimiter, and suffix configurable, look up the use cases of it

karma-o (changes requested):
Overall a really good solution, so you did a good job! But we still need to tweak a couple things

src/main/java/core/basesyntax/service/impl/CsvFileReader.java
```diff
@@ -0,0 +1,30 @@
...
+
+import core.basesyntax.service.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.stream.Stream;
+
+public class CsvFileReader implements FileReader {
+    private static final int MIN_NUMBER_OF_ROWS = 2;
```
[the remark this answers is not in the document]
Bohadan-Markatov: because the first line is the name of the columns

src/main/java/core/basesyntax/service/impl/CsvWriterService.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class CsvWriterService implements FileWriter {
+
+    @Override
+    public void write(String data, String newReportName) {
+        if (Files.exists(Paths.get(newReportName + ".csv"))) {
+            throw new RuntimeException("A file with this name already exists");
+        }
+        try {
+            Files.write(Paths.get(newReportName + ".csv"), data.getBytes());
```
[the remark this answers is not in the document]
andrii-hoienko: not resolved

src/main/java/core/basesyntax/service/impl/FruitTransactionParser.java
```diff
@@ -0,0 +1,19 @@
...
+import core.basesyntax.service.util.CsvRowValidator;
+import core.basesyntax.service.util.TransactionSplitter;
+
+public class FruitTransactionParser implements TransactionParser<FruitTransaction> {
+
+    @Override
+    public FruitTransaction parse(String row) {
+        if (row == null || !CsvRowValidator.validate(row)) {
+            throw new RuntimeException("Invalid data in row: " + row);
+        }
```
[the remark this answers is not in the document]
andrii-hoienko: not fully resolved

src/main/java/core/basesyntax/service/impl/CsvWriterService.java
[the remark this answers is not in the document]
Bohadan-Markatov: if it is necessary to work with a new format, I will create a new implementation of FileReader and FileWriter

src/main/java/core/basesyntax/service/impl/FruitTransactionProcessor.java
```diff
@@ -0,0 +1,25 @@
...
+public class FruitTransactionProcessor implements TransactionProcessor<FruitTransaction> {
+    private OperationStrategy<FruitTransaction> strategy;
+
+    public FruitTransactionProcessor(OperationStrategy<FruitTransaction> strategy) {
+        this.strategy = strategy;
+    }
+
+    @Override
+    public void process(FruitTransaction transaction) {
+        OperationHandler<FruitTransaction> operationStrategy = strategy.get(transaction);
```
karma-o, line 17: this is a variable of type `OperationHandler` it should not be named `operationStrategy`

src/main/java/core/basesyntax/service/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public OperationStrategyImpl(Map<OperationType,
+            OperationHandler<FruitTransaction>> operationHandlerMap) {
+        this.operationHandlerMap = operationHandlerMap;
+    }
+
+    @Override
+    public OperationHandler<FruitTransaction> get(FruitTransaction transaction) {
+        return operationHandlerMap.get(transaction.getTransactionType());
+    }
```
karma-o, line 20: this whole class is essentially for just incapsulating the map in here, but I think that is complicating things, so I would just use this map directly in the `FruitTransactionProcesor`

src/main/java/core/basesyntax/service/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,21 @@
...
+    private final Map<OperationType, OperationHandler<FruitTransaction>> operationHandlerMap;
+
+    public OperationStrategyImpl(Map<OperationType,
+            OperationHandler<FruitTransaction>> operationHandlerMap) {
+        this.operationHandlerMap = operationHandlerMap;
+    }
+
+    @Override
+    public OperationHandler<FruitTransaction> get(FruitTransaction transaction) {
+        return operationHandlerMap.get(transaction.getTransactionType());
```
karma-o, line 19: I also do not like the idea of passing the whole transaction here, where we only need it's type, that's one of the reasons why if we fix that - it will be just a map with non-direct access, which is kind of strange

src/main/java/core/basesyntax/service/impl/FruitTransactionProcessor.java
```diff
@@ -0,0 +1,25 @@
...
+        if (operationStrategy != null) {
+            operationStrategy.handle(transaction);
+        } else {
+            throw new RuntimeException("Unknown operation type: "
+                    + transaction.getTransactionType());
+        }
```
karma-o, line 23: do not forget about the "return early" pattern
```suggestion
        if (operationStrategy == null) {
            throw new RuntimeException("Unknown operation type: "
                    + transaction.getTransactionType());
        }
        operationStrategy.handle(transaction);
```

src/main/java/core/basesyntax/service/impl/ReportProviderImpl.java
```diff
@@ -0,0 +1,27 @@
...
+        }
+    };
+
+    @Override
+    public String provide() {
+        String header = REPORT_TITLE + ROW_SEPARATOR;
+        return header + StorageDao.storage.entrySet().stream()
+                .peek(trans -> positiveValueCheck.accept(trans.getValue()))
+                .map(trans -> trans.getKey() + COLUMN_SEPARATOR + trans.getValue())
+                .collect(Collectors.joining(ROW_SEPARATOR));
```
karma-o, line 25: you are still using the usual string concat in 2 places. I was talking about that:

image

https://www.geeksforgeeks.org/java-8-streams-collectors-joining-method-with-examples/ - screenshot of the last example here

not sure what to do with the column separator, and how to do that in the best possible way, but we can definetly set `header` as a prefix in `Collectors.joining()`. And now that I thing about this column separator - maybe your `StringBuilder` approach was actually better, but now you know about the `joining()` 😅

src/main/java/core/basesyntax/service/impl/FruitTransactionParser.java
```diff
@@ -0,0 +1,26 @@
...
+    private static final int OPERATION_TYPE_COLUMN = 0;
+    private static final int PRODUCT_TYPE_COLUMN = 1;
+    private static final int OPERATION_VALUE_COLUMN = 2;
+
+    @Override
+    public FruitTransaction parse(String row) {
+        if (!CsvRowValidator.validate(row)) {
+            throw new RuntimeException("Invalid data in row: " + row
+            + " The expected format should be: b,apple,100");
+        }
```
karma-o, line 19: I still think that this whole thing should be in the `CsvRowValidator.validate()` since we have a whole class dedicated to validation

karma-o (approved):
Approved with a couple more comments 🔥  please make use of them before starting the work on the fruit shop tests

PR #1088, opened 2024-03-19
Added an implementation for task fruit shop. Created all necessary cl…

nklimovych:
…ass and methods according to the SOLID principles.

### DIFF
--- a/src/main/java/core/basesyntax/db/Storage.java
+++ b/src/main/java/core/basesyntax/db/Storage.java
@@ -1,0 +1,17 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    private static Map<String, Integer> products;
+
+    public Storage() {
+        products = new HashMap<>();
+    }
+
+    public Map<String, Integer> getProducts() {
+        return products;
+    }
+
+}
--- a/src/main/java/core/basesyntax/model/Transaction.java
+++ b/src/main/java/core/basesyntax/model/Transaction.java
@@ -1,0 +1,63 @@
+package core.basesyntax.model;
+
+public class Transaction {
+    private Operation operation;
+    private String name;
+    private int quantity;
+
+    public Transaction(String operationCode, String name, int quantity) {
+        this.operation = Operation.getByCode(operationCode);
+        this.name = name;
+        this.quantity = quantity;
+    }
+
+    public Operation getOperation() {
+        return operation;
+    }
+
+    public void setOperation(Operation operation) {
+        this.operation = operation;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public int getQuantity() {
+        return quantity;
+    }
+
+    public void setQuantity(int quantity) {
+        this.quantity = quantity;
+    }
+
+    public enum Operation {
+        BALANCE("b"),
+        SUPPLY("s"),
+        PURCHASE("p"),
+        RETURN("r");
+
+        private final String code;
+
+        Operation(String code) {
+            this.code = code;
+        }
+
+        public String getCode() {
+            return code;
+        }
+
+        public static Operation getByCode(String operationCode) {
+            for (Operation operation : Operation.values()) {
+                if (operation.code.equals(operationCode)) {
+                    return operation;
+                }
+            }
+            throw new IllegalArgumentException("Invalid operation code: " + operationCode);
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/service/FileDataParser.java
+++ b/src/main/java/core/basesyntax/service/FileDataParser.java
@@ -1,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.Transaction;
+import java.util.List;
+
+public interface FileDataParser {
+    List<Transaction> parse(List<String> data);
+}
--- a/src/main/java/core/basesyntax/service/WriterService.java
+++ b/src/main/java/core/basesyntax/service/WriterService.java
@@ -1,0 +1,7 @@
+package core.basesyntax.service;
+
+import java.util.List;
+
+public interface WriterService {
+    void writeToFile(List<String> report, String filePath);
+}
--- a/src/main/java/core/basesyntax/service/impl/FileDataParserImpl.java
+++ b/src/main/java/core/basesyntax/service/impl/FileDataParserImpl.java
@@ -1,0 +1,28 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.Transaction;
+import core.basesyntax.service.FileDataParser;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class FileDataParserImpl implements FileDataParser {
+    @Override
+    public List<Transaction> parse(List<String> data) {
+        return data.stream()
+                .skip(1)
+                .map(this::parseTransaction)
+                .collect(Collectors.toList());
+    }
+
+    private Transaction parseTransaction(String data) {
+        String[] dataArray = data.split(",");
+        if (dataArray.length != 3) {
+            throw new IllegalArgumentException("Invalid transaction data format");
+        }
+
+        String operation = dataArray[0];
+        String name = dataArray[1];
+        int quantity = Integer.parseInt(dataArray[2]);
+        return new Transaction(operation, name, quantity);
+    }
+}
--- a/src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
@@ -1,0 +1,18 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.ReaderService;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class ReaderServiceImpl implements ReaderService {
+    @Override
+    public List<String> readFromFile(String filePath) {
+        try {
+            return Files.readAllLines(Path.of(filePath));
+        } catch (IOException e) {
+            throw new RuntimeException("Error reading from a file: " + filePath, e);
+        }
+    }
+}
--- a/src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
@@ -1,0 +1,26 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.ReportService;
+import core.basesyntax.service.StorageService;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class ReportServiceImpl implements ReportService {
+    private final StorageService storageService;
+
+    public ReportServiceImpl() {
+        storageService = new StorageServiceImpl();
+    }
+
+    @Override
+    public List<String> generateReport() {
+        List<String> reportData = new ArrayList<>();
+        reportData.add("fruit,quantity");
+
+        for (Map.Entry<String, Integer> entry : storageService.getAll().entrySet()) {
+            reportData.add(entry.getKey() + "," + entry.getValue());
+        }
+        return reportData;
+    }
+}
--- a/src/main/java/core/basesyntax/service/impl/StorageServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/impl/StorageServiceImpl.java
@@ -1,0 +1,33 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.StorageService;
+import java.util.HashMap;
+
+public class StorageServiceImpl implements StorageService {
+    private final Storage storage;
+
+    public StorageServiceImpl() {
+        storage = new Storage();
+    }
+
+    @Override
+    public void save(String name, Integer quantity) {
+        storage.getProducts().put(name, quantity);
+    }
+
+    @Override
+    public Integer get(String key) {
+        return storage.getProducts().get(key);
+    }
+
+    @Override
+    public HashMap<String, Integer> getAll() {
+        return new HashMap<>(storage.getProducts());
+    }
+
+    @Override
+    public void update(String name, Integer value) {
+        storage.getProducts().put(name, value);
+    }
+}
--- a/src/main/java/core/basesyntax/service/impl/TransactionProcessorImpl.java
+++ b/src/main/java/core/basesyntax/service/impl/TransactionProcessorImpl.java
@@ -1,0 +1,20 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.Transaction;
+import core.basesyntax.service.TransactionProcessor;
+import core.basesyntax.strategy.OperationStrategy;
+import core.basesyntax.strategy.impl.OperationStrategyImpl;
+import java.util.List;
+
+public class TransactionProcessorImpl implements TransactionProcessor {
+    private final OperationStrategy operationStrategy;
+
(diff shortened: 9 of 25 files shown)


src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public Storage() {
+        products = new HashMap<>();
+    }
+
+    public Map<String, Integer> getProducts() {
+        return products;
+    }
+
+}
```
ValentynSharshon, line 17: I think you can use in line declaration of map here and not use constructor and getter. But your approach is ok too, nobody cancels encapsulation.

src/main/java/core/basesyntax/service/impl/FileDataParserImpl.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public List<Transaction> parse(List<String> data) {
+        return data.stream()
+                .skip(1)
+                .map(this::parseTransaction)
+                .collect(Collectors.toList());
+    }
+
+    private Transaction parseTransaction(String data) {
+        String[] dataArray = data.split(",");
```
ValentynSharshon, line 18: I think it's better to use a constant instead of ",".

src/main/java/core/basesyntax/service/impl/FileDataParserImpl.java
```diff
@@ -0,0 +1,28 @@
...
+        if (dataArray.length != 3) {
```
ValentynSharshon, line 19: I think it's better to use a constant instead of 3.

src/main/java/core/basesyntax/service/impl/FileDataParserImpl.java
```diff
@@ -0,0 +1,28 @@
...
+            throw new IllegalArgumentException("Invalid transaction data format");
+        }
+
+        String operation = dataArray[0];
+        String name = dataArray[1];
+        int quantity = Integer.parseInt(dataArray[2]);
```
ValentynSharshon, line 25: I think it's better to use a constant instead of 0, 1 and 2.

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class ReportServiceImpl implements ReportService {
+    private final StorageService storageService;
+
+    public ReportServiceImpl() {
+        storageService = new StorageServiceImpl();
+    }
```
ValentynSharshon, line 14: I think it's better to pass StorageService as a constructor parameter, but I'm not sure, let mentor take a look.

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Override
+    public List<String> generateReport() {
+        List<String> reportData = new ArrayList<>();
+        reportData.add("fruit,quantity");
```
ValentynSharshon, line 19: I think it's better to use a constant instead of "fruit,quantity".

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+
+        for (Map.Entry<String, Integer> entry : storageService.getAll().entrySet()) {
+            reportData.add(entry.getKey() + "," + entry.getValue());
```
ValentynSharshon, line 22: I think it's better to use a constant instead of ",".

src/main/java/core/basesyntax/service/impl/TransactionProcessorImpl.java
```diff
@@ -0,0 +1,20 @@
...
+import core.basesyntax.strategy.OperationStrategy;
+import core.basesyntax.strategy.impl.OperationStrategyImpl;
+import java.util.List;
+
+public class TransactionProcessorImpl implements TransactionProcessor {
+    private final OperationStrategy operationStrategy;
+
+    public TransactionProcessorImpl() {
+        operationStrategy = new OperationStrategyImpl();
+    }
```
ValentynSharshon, line 14: I think it's better to pass OperationStrategy as a constructor parameter, but I'm not sure, let mentor take a look.

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class WriterServiceImpl implements WriterService {
+    @Override
+    public void writeToFile(List<String> data, String filePath) {
+        try (BufferedWriter writer = Files.newBufferedWriter(Path.of(filePath))) {
+            if (data != null) {
+                for (int i = 0; i < data.size() - 1; i++) {
```
ValentynSharshon, line 15: Magic number

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+                    writer.write(data.get(i));
+                    writer.newLine();
+                }
+                writer.write(data.get(data.size() - 1));
```
ValentynSharshon, line 19: Magic number

src/main/java/core/basesyntax/strategy/impl/BalanceOperation.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.strategy.impl;
+
+import core.basesyntax.model.Transaction;
+import core.basesyntax.service.StorageService;
+import core.basesyntax.service.impl.StorageServiceImpl;
+import core.basesyntax.strategy.Operation;
+
+public class BalanceOperation implements Operation {
+    private final StorageService storageService;
```
ValentynSharshon, line 9: It may be better to make StorageService a field of the Operation interface to avoid duplicating code in each implementation.

ValentynSharshon (approved):

src/main/java/core/basesyntax/db/Storage.java
nklimovych: I'm not sure too, so let's wait for a mentor))

src/main/java/core/basesyntax/service/impl/FileDataParserImpl.java
nklimovych: 🥲

src/main/java/core/basesyntax/service/impl/ReportServiceImpl.java
nklimovych: done 👌🏻

src/main/java/core/basesyntax/strategy/impl/BalanceOperation.java
nklimovych: you may be right, but I'm not sure, so I'll wait for the mentor's feedback

src/main/java/core/basesyntax/model/Transaction.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public int getQuantity() {
+        return quantity;
+    }
+
+    public void setQuantity(int quantity) {
+        this.quantity = quantity;
+    }
+
+    public enum Operation {
```
karma-o, line 38: I'd rather move this enum to separate file

src/main/java/core/basesyntax/model/Transaction.java
```diff
@@ -0,0 +1,63 @@
+package core.basesyntax.model;
+
+public class Transaction {
+    private Operation operation;
+    private String name;
```
karma-o, line 5: to add a bit of context about what this transaction is, and do not forget about the getters and setters renaming
```suggestion
    private String fruitName;
```

src/main/java/core/basesyntax/service/impl/StorageServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public Map<String, Integer> getAll() {
+        return storage.getProducts();
+    }
+
+    @Override
+    public void update(String name, Integer value) {
+        storage.getProducts().put(name, value);
+    }
```
karma-o, line 31: save and update are identical, remove one of them

src/main/java/core/basesyntax/service/impl/StorageServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.StorageService;
+import java.util.Map;
+
+public class StorageServiceImpl implements StorageService {
+    private final Storage storage = new Storage();
+
+    public StorageServiceImpl() {
+    }
```
karma-o, line 11: redundant constructor we have it available by default if we do not have any others

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import java.nio.file.Path;
+import java.util.List;
+
+public class ReaderServiceImpl implements ReaderService {
+    @Override
+    public List<String> readFromFile(String filePath) {
+        try {
+            return Files.readAllLines(Path.of(filePath));
+        } catch (IOException e) {
+            throw new RuntimeException("Error reading from a file: " + filePath, e);
```
karma-o, line 15: Let's replace `RuntimeException` with something custom

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,28 @@
...
+import core.basesyntax.service.WriterService;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class WriterServiceImpl implements WriterService {
+
+    private static final int LINE_OFFSET = 1;
```
karma-o, line 12: why do we need that?

src/main/java/core/basesyntax/strategy/Operation.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.model.Transaction;
+
+public interface Operation {
+    void execute(Transaction transaction);
```
karma-o, line 6: we can't have enum and this class named identically, this is a handler for a strategy, so:
```suggestion
public interface OperationHandler {
    void handle(Transaction transaction);
```

src/main/java/core/basesyntax/strategy/impl/OperationStrategyImpl.java
```diff
@@ -0,0 +1,28 @@
...
+
+        operations.entrySet().stream()
+                .filter(entry -> entry.getKey().equals(operationCode))
+                .map(Map.Entry::getValue)
+                .findFirst()
+                .ifPresentOrElse(
+                        operation -> operation.execute(transaction),
+                        () -> {
+                            throw new RuntimeException("Unsupported operation: " + operationCode);
+                        });
```
karma-o, line 26: filtering the whole map to get a single element by the key is ineffective, use `get()`
```suggestion
        var handler = operations.get(operationCode):
        // throw custom ex if handler is null
        handler.handle(transaction)
```

src/main/java/core/basesyntax/strategy/impl/BalanceOperation.java
karma-o: As per your question regarding the code duplication of common field for all Handlers: interfaces can have only constants, so we can not do that. But if you really want to, you can add an abstract class in between interface and class, put this field and a constructor there, and remove from here and that will work, but I would say that this is unnecessary, so up to you

src/main/java/core/basesyntax/strategy/OperationStrategy.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.model.Transaction;
+
+public interface OperationStrategy {
+    void executeOperation(Transaction transaction);
```
karma-o, line 6: Also, I would actually rather move the execution and handling of the transaction itself to the `TransactionProcessor` judging by its name, thus leaving a this class only a single responsibility of providing a proper `OperationHandler` which will be used in the processor, so:
```suggestion
public interface OperationStrategy {
    OperationHandler getHandlerFor(Transaction transaction);
```

src/main/java/core/basesyntax/service/WriterService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+import java.util.List;
+
+public interface WriterService {
+    void writeToFile(List<String> report, String filePath);
```
karma-o, line 6: we should not adapt abstraction to the implementation, this writer can write anything, not only reports, to any file
```suggestion
public interface WriterService {
    void writeToFile(List<String> data, String filePath);
```

src/main/java/core/basesyntax/service/FileDataParser.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.Transaction;
+import java.util.List;
+
+public interface FileDataParser {
```
karma-o, line 6:
```suggestion
public interface TransactionParser {
```
same for impl class

src/main/java/core/basesyntax/service/impl/FileDataParserImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final String DIVIDER = ",";
+    private static final int REQUIRED_LENGTH = 3;
+    private static final int OPERATION_INDEX = 0;
+    private static final int NAME_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public List<Transaction> parse(List<String> data) {
+        return data.stream()
+                .skip(1)
```
karma-o, line 19: magic number

karma-o (changes requested):
Looks good so far, left some comments

src/main/java/core/basesyntax/model/Transaction.java
nklimovych: There was an example in README.

```
  private Operation operation;
  private String fruit;
  private int quantity;

  // getters, setters, ...
  
  public enum Operation {
    BALANCE("b"),
    SUPPLY("s"),
    PURCHASE("p"),
    RETURN("r");

    private String code;

    Operation(String code) {
      this.code = code;
    }

    public String getCode() {
      return code;
    }
  }
}
```

src/main/java/core/basesyntax/db/Storage.java
karma-o: actually this is a bit confusing because every time you create a new `Storage`, you reinitialize the map, which is unclear and shouldn't work like that, I would actually just make it 
`public static final Map<...> products = new hashMap<>()` without any getters etc. because we have another level of abstration for it in the form of `StorageService`

src/main/java/core/basesyntax/strategy/impl/OperationStrategyImpl.java
nklimovych: 🥲

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
nklimovych: student left me a comment that it was a magic number 😅 
I'll change it back)

src/main/java/core/basesyntax/service/impl/StorageServiceImpl.java
nklimovych: i agree. I've deleted update method

PR #1089, opened 2024-03-19
Solution SOLID

fedaktaras:
Completed 
Using sealed classes


src/main/java/core/basesyntax/service/impl/ReaderCsv.java
```diff
@@ -0,0 +1,38 @@
...
+
+    static class DataProcessor {
+        private static final Set<String> TITLES_TO_REMOVE = Set.of("type,fruit,quantity");
+
+        public static List<String> filterTitles(List<String> data) {
+            return data.stream()
+                    .filter(row -> !TITLES_TO_REMOVE.contains(row))
+                    .collect(Collectors.toList());
+        }
+    }
```
TonyH277, line 37: could be replaced with lines.remove(0);

TonyH277 (changes requested):
Good job! Pay attention to constants

src/main/java/core/basesyntax/model/Operation.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax.model;
+
+public abstract sealed class Operation permits Balance, Purchase, Supply, Return {
```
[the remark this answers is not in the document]
fedaktaras: > Interesting aproach woth sealed classes. IMHO: Move all handlers from "model" to "strategy/impl" will be easier + if we had a lot of handlers -> you should permit them all (sound like extra work)

I will not be able to get invalid operation in any case. The best to use switch but it is in later java

src/main/java/core/basesyntax/strategy/OperationStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.Balance;
+import core.basesyntax.model.Operation;
+import core.basesyntax.model.Purchase;
+import core.basesyntax.model.Return;
+import core.basesyntax.model.Supply;
+
+public class OperationStrategy {
+
+    public void processData(Operation operation) {
```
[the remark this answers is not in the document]
fedaktaras: > It's more flexible if you use Map<> or switch/case for strategy

The idea was to use switch/case but with java 17 it doesn't work as required.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,26 @@
...
+public class Main {
+    public static void main(String[] args) {
+        Reader reader = new ReaderCsv();
+        List<String> lines = reader.readData();
+        Parser parser = new ParserImpl();
+        List<Operation> parsedData = parser.parseToOperations(lines);
+        OperationStrategy operationStrategy = new OperationStrategy();
+        for (var pd : parsedData) {
+            operationStrategy.processData(pd);
+        }
```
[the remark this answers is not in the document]
fedaktaras: > You could pass parsedData to operationStrategy trough constructor, and choose sertain strategy inside

It could be static field. It is created only once

pom.xml
```diff
@@ -41,6 +41,14 @@
...
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>17</source>
+                    <target>17</target>
+                </configuration>
+            </plugin>
```
andrii-hoienko, line 51: What is the reason to add this?

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,29 @@
...
+import core.basesyntax.service.impl.WriterCsv;
+import core.basesyntax.strategy.OperationStrategy;
+import java.util.List;
+
+public class Main {
+    private static final String WRITE_FILE_PATH = "src/files/fruits_report.csv";
+    private static final String READ_FILE_PATH = "src/files/fruits.csv";
+
+    public static void main(String[] args) {
+        Reader reader = new ReaderCsv(READ_FILE_PATH);
```
andrii-hoienko, line 18: What if we want to reuse reader and read not only from this file READ_FILE_PATH? (same with writer)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,29 @@
...
+        List<String> lines = reader.readData();
+        Parser parser = new ParserImpl();
+        List<Operation> parsedData = parser.parseToOperations(lines);
+        OperationStrategy operationStrategy = new OperationStrategy();
+        for (var pd : parsedData) {
```
andrii-hoienko, line 23:
```suggestion
        for (var data: parsedData) {
```

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    public static final Map<String, Integer> data = new HashMap<>();
```
andrii-hoienko, line 7: Don`t forget about name conventions

src/main/java/core/basesyntax/model/Purchase.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.model;
+
+public final class Purchase extends Operation {
```
andrii-hoienko, line 3: I suggest making Operation enum, operation objects always should be the same and in limited amount. And create FruitTransaction class to keep operation, amount and fruit name

src/main/java/core/basesyntax/service/Parser.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.Operation;
+import java.util.List;
+
+public interface Parser {
+    List<Operation> parseToOperations(List<String> record);
```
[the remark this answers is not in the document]
andrii-hoienko: agree

src/main/java/core/basesyntax/service/impl/ReaderCsv.java
```diff
@@ -0,0 +1,42 @@
...
+                lines.add(line);
+            }
+            return DataProcessor.filterTitles(lines);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file");
+        }
+    }
+
+    static class DataProcessor {
+        private static final Set<String> TITLES_TO_REMOVE = Set.of("type,fruit,quantity");
```
andrii-hoienko, line 34: What if we change the title, for example, we will add whitespace or something? In this case, we will go to the code and add a new value to the set. I suggest just skipping first element of the array to avoid the header

src/main/java/core/basesyntax/service/impl/ReaderCsv.java
```diff
@@ -0,0 +1,42 @@
...
+
+        public static List<String> filterTitles(List<String> data) {
+            return data.stream()
+                    .filter(row -> !TITLES_TO_REMOVE.contains(row))
```
andrii-hoienko, line 38:
```suggestion
                    .skip(1)
```

src/main/java/core/basesyntax/service/impl/ReaderCsv.java
```diff
@@ -0,0 +1,42 @@
...
+            while ((line = br.readLine()) != null) {
+                lines.add(line);
+            }
+            return DataProcessor.filterTitles(lines);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file");
+        }
+    }
+
+    static class DataProcessor {
```
andrii-hoienko, line 33: No reason to create separate nested class for that, just method is enough

src/main/java/core/basesyntax/service/impl/WriterCsv.java
```diff
@@ -0,0 +1,26 @@
...
+    public WriterCsv(String filePath) {
+        this.filePath = filePath;
+    }
+
+    public void writeData() {
+        try (FileWriter fileWriter = new FileWriter(filePath)) {
+            fileWriter.write("fruit,quantity" + System.lineSeparator());
+            for (Map.Entry<String, Integer> entry : Storage.data.entrySet()) {
+                fileWriter.write(entry.getKey() + "," + entry.getValue() + System.lineSeparator());
+            }
```
andrii-hoienko, line 21: Please make string literals constants

src/main/java/core/basesyntax/service/impl/WriterCsv.java
```diff
@@ -0,0 +1,26 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t write file.");
```
andrii-hoienko, line 23:
```suggestion
            throw new RuntimeException("Can`t write file:" + filePath);
```

andrii-hoienko (changes requested):
Good job, but some significant improvements required

pom.xml
fedaktaras: > What is the reason to add this?

I made switch/case for sealed classes. I somehow changed project settings and it wasn`t pass tests on github.

yevheniisynytsia (approved):
Great job!

PR #1092, opened 2024-03-19
Created fruit shop

Recdt:


src/main/java/core/basesyntax/service/strategy/ActivityStrategy.java
```diff
@@ -0,0 +1,18 @@
...
+
+import core.basesyntax.service.strategy.impl.BalanceService;
+import core.basesyntax.service.strategy.impl.PurchaseService;
+import core.basesyntax.service.strategy.impl.ReturnService;
+import core.basesyntax.service.strategy.impl.SupplyService;
+import core.basesyntax.utility.FruitTransaction;
+
+public class ActivityStrategy {
+    public ActivityService getActivityService(FruitTransaction.Operation activityType) {
+        return switch (activityType) {
```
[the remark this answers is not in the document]
karma-o: @MolchanovAlexander You should not approve the PR untill a big mistake like that is fixed

src/main/java/core/basesyntax/service/strategy/ActivityService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service.strategy;
+
+import core.basesyntax.utility.FruitTransaction;
+
+public interface ActivityService {
+    int execute(FruitTransaction transaction);
```
karma-o, line 6: I would not call that a service it is the template to implement logic of handling some operation, so 
```suggestion
public interface OperationHandler {
    int handle(FruitTransaction transaction);
```
Word `service` is usualy used when you have some logic in relation to some entity, like `UserService`, `FruitService`, `BookService`, `ProductService` etc. there are other cases as well but mostly it is used like that,
There are plenty other words that  we can use to define waht the class does, see the next comments

src/main/java/core/basesyntax/utility/service/ReportService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.utility.service;
+
+import core.basesyntax.utility.Pair;
+import java.util.List;
+
+public interface ReportService {
+    List<Pair<String, Integer>> createReport();
```
karma-o, line 7: For example this can be named as follows:
```suggestion
public interface ReportProvider {
    List<Pair<String, Integer>> provide();
```
because it has only 1 method without any parameters and returns some value. There are also useful words like `Consumer`, `Executor`, `Mapper`, and many others, so we always should keep in mind good naming

src/main/java/core/basesyntax/service/strategy/ActivityStrategy.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax.service.strategy;
+
+import core.basesyntax.service.strategy.impl.BalanceService;
+import core.basesyntax.service.strategy.impl.PurchaseService;
+import core.basesyntax.service.strategy.impl.ReturnService;
+import core.basesyntax.service.strategy.impl.SupplyService;
+import core.basesyntax.utility.FruitTransaction;
+
+public class ActivityStrategy {
```
karma-o, line 7: Same here, this is an `OperationStrategy` because it provides Handlers based on the `Operation` of the transaction, do not forget to correct the method here respectively

src/main/java/core/basesyntax/service/strategy/ActivityStrategy.java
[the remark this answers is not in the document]
karma-o: Agree, we should have a `Map<Operation, OperationHandler>`, initialize it in the main, and pass here through the constructor to get the right handler by operation instead of switch, check the theory and Bohdan's example on why it is better

src/main/java/core/basesyntax/utility/service/ParserService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.utility.service;
+
+import core.basesyntax.utility.FruitTransaction;
+import java.util.List;
+
+public interface ParserService {
+    List<FruitTransaction> parseData(List<String> data);
```
karma-o, line 7:
```suggestion
public interface TransactionParser {
    List<FruitTransaction> parseData(List<String> data);
```
OR
```suggestion
public interface Parser<T> {
    List<T> parseData(List<String> data);
```

src/main/java/core/basesyntax/ProgramStart.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public static void main(String[] args) {
+        FileService fileService = new FileServiceImpl();
+        List<FruitTransaction> fruitTransactions = new FruitDataParserImpl()
+                .parseData(fileService.readFromFile(DATA_FILE));
+
+        ActivityStrategy activityStrategy = new ActivityStrategy();
+        for (var element : fruitTransactions) {
+            activityStrategy.getActivityService(element.getOperation()).execute(element);
+        }
```
karma-o, line 38: we should not iterate over data in main, add some `TransactionProcessor` class and do it there

karma-o (changes requested):

src/main/java/core/basesyntax/utility/FileServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+import core.basesyntax.utility.service.FileService;
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FileServiceImpl implements FileService {
```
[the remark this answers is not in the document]
Recdt: We are discussing that on q&a and Yevhen said that this is ok

src/main/java/core/basesyntax/utility/FruitTransaction.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public void setFruit(String fruit) {
+        this.fruit = fruit;
+    }
+
+    public void setQuantity(int quantity) {
+        this.quantity = quantity;
+    }
+
+    public enum Operation {
```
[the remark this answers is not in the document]
Recdt: This was written in readme
image

src/main/java/core/basesyntax/utility/service/ReportService.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.utility.service;
```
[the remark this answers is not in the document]
Recdt: I'm sorry, thats because first I made the util classes with a static methouds. But on Q&A was discussed that, and I replaced this in code, but don't managed packages after that

src/main/java/core/basesyntax/service/impl/ReportProviderImpl.java
```diff
@@ -0,0 +1,20 @@
...
+public class ReportProviderImpl implements ReportProvider {
+    private static final String REPORT_HEADER = "fruit,quantity";
+    private static final String ELEMENT_SEPARATOR = ",";
+
+    @Override
+    public String provide() {
+        StringBuilder report = new StringBuilder(REPORT_HEADER);
+
+        for (var element : Storage.fruitStorage.entrySet()) {
+            report.append("\n").append(element.getKey())
```
d1sam, line 15: let's better to use `System.lineSeparator();`

d1sam (changes requested):
Good job, but there are some changes that should be done according to my comments!

d1sam (changes requested):
Please make changes according to my comment.

d1sam (approved):
Well done!

PR #1095, opened 2024-03-19
Java core solid homework fruit shop

azzilian:


src/main/java/core/basesyntax/service/impl/FruitFileReaderImpl.java
```diff
@@ -0,0 +1,24 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class FruitFileReaderImpl implements FruitFileReader {
+    private CheckFilePathImpl filePathExist = new CheckFilePathImpl();
+
+    @Override
+    public List<String> readFile(String filePath) {
+        if (filePathExist.checkFilePath(filePath)) {
```
VolandevlodD, line 15: Here, you're using your custom check for a string representing the path to a file and then throwing an exception if the file does not exist. It's okay, I think, but the static factory method `Path.of` already throws an `InvalidPathException` if the path string cannot be converted to a `Path`. So maybe we can simplify the logic by catching this exception and handling it. 
**For example, by adding some custom message to this error and rethrowing it further because it is unchecked.**
IMHO

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class Main {
+    private static String fileOpen = "src/main/resources/fruitts.csv";
+    //    private static String fileOpen = "src/main/resources/fruitsWithNulls.csv";
+    private static String fileSave = "src/main/resources/report.csv";
+
+    public static void main(String[] args) {
+        FruitFileReader reader = new FruitFileReaderImpl();
+        List<String> fileString = reader.readFile(fileOpen);
+        FruitRawStringParserImpl parser = new FruitRawStringParserImpl();
```
VolandevlodD: It`s better to use abstraction
```suggestion
        FruitRawStringParser parser = new FruitRawStringParserImpl();
```

src/main/java/core/basesyntax/service/impl/FruitRawStringParserImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class FruitRawStringParserImpl implements FruitRawStringParser {
+    private static final int OPERATION_NEXT_INDEX = 0;
+    private static final int FRUIT_NAME_INDEX = 1;
+    private static final int FIRST_CVS_LINE = 1;
+    private static final int FRUIT_QUANTITY_INDEX = 2;
+
+    private FruitStringsCheckImpl checkFruitsInFile = new FruitStringsCheckImpl();
```
VolandevlodD, line 15:
```suggestion
    private final FruitStringsCheck checkFruitsInFile;

    public FruitRawStringParserImpl(FruitStringsCheckImpl checkFruitsInFile) {
        this.checkFruitsInFile = checkFruitsInFile;
    }

```
Field could be final, it`s better to use abstraction, and i`m almost sure that it is better to create an instance of FruitStringsCheck in main and pass it as the parameter in constuctor.

VolandevlodD: And also, since this check is used only in one class, I think it may be even better to create a private method in the same class rather than a separate interface and class.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,47 @@
...
+        var balance = new BalanceOperation();
+        var supply = new SupplyOperation();
+        var returns = new ReturnOperation();
+        var purchase = new PurchaseOperation();
+        List<OperationHandler> handlers = List.of(balance,returns,purchase,supply);
+        FruitStrategy strategy = new FruitStrategy(handlers);
+        List<FruitTransactionDto> dtos = readerService;
+        for (var dto : dtos) {
+            strategy.getHandlers(dto).forEach(oh -> oh.apply(dto));
+        }
```
VolandevlodD: Do we really need this variable assigment?
```suggestion
     for (var dto : readerService) {
            strategy.getHandlers(dto).forEach(oh -> oh.apply(dto));
        }
```

src/main/java/core/basesyntax/service/strategy/impl/BalanceOperation.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax.service.strategy.impl;
+
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.service.strategy.OperationHandler;
+import java.util.HashMap;
+
+public class BalanceOperation implements OperationHandler {
+    private StorageDaoImpl actionDB = new StorageDaoImpl();
```
VolandevlodD, line 9: For all handlers, you're creating a new `StorageDaoImpl();`. Here, you can also simply make a field and pass a reference for the DAO as a parameter to the constructor.

VolandevlodD: This way u will create ur DAO once and use the same dao in all handlers.

src/main/java/core/basesyntax/service/strategy/FruitStrategy.java
```diff
@@ -0,0 +1,18 @@
...
+    public FruitStrategy(List<OperationHandler> handlers) {
+        this.handlers = handlers;
+    }
+
+    public List<OperationHandler> getHandlers(FruitTransactionDto dto) {
+        return handlers.stream()
+                .filter(h -> h.isApplicable(dto))
+                .toList();
+    }
+}
```
VolandevlodD, line 18: I remember during the lesson with the coach, we were exploring applicable handlers like this, but let's simplify. In this specific task, we have only one handler for each operation type, and every time we receive only one handler. Therefore, there's no reason to put it into a list.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,47 @@
...
+        System.out.println(Storage.fruitsQuantity);
+        FruitReportCreate prapaireReport = new FruitReportCreateImpl();
```
VolandevlodD: `prapaireReport` typo?
```suggestion
        FruitReportCreate prepareReport = new FruitReportCreateImpl();
```

src/main/java/core/basesyntax/service/impl/FruitReportCreateImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+import core.basesyntax.exception.NegativeBalanceException;
+import core.basesyntax.service.interfaces.FruitReportCreate;
+import java.util.HashMap;
+import java.util.Map;
+
+public class FruitReportCreateImpl implements FruitReportCreate {
+    public String createReport(HashMap<String, Integer> fruitQuantity) {
+        StringBuilder builder = new StringBuilder();
+        builder.append("fruit").append(",").append("quantity").append("\n");
```
VolandevlodD, line 11:
```suggestion
        builder.append("fruit").append(",").append("quantity").append("System.lineSeparator");
```

VolandevlodD: It's better to use `System.lineSeparator()` for compatibility.

src/main/java/core/basesyntax/service/impl/FruitReportCreateImpl.java
```diff
@@ -0,0 +1,25 @@
...
+        for (Map.Entry<String, Integer> fruits: fruitQuantity.entrySet()) {
+            if (fruits.getValue() < 0) {
+                throw new NegativeBalanceException("Incorrect quantity,"
+                        + " some operations probably missing,"
+                        + " quantity for fruit"
+                        + fruits.getKey()
+                        + " is "
+                        + fruits.getValue());
+            }
+            builder.append(fruits.getKey()).append(",").append(fruits.getValue()).append("\n");
```
VolandevlodD, line 21: It's better to make `","` a constant, since it's a repetitive string.

VolandevlodD (changes requested):
Good, leave comments below. Please check all classes for the same mistakes, ensuring dependency on abstractions rather than implementations. There is no reason to create interfaces this way. Also, please check for common mistakes.

src/main/java/core/basesyntax/service/interfaces/FruitFileSaver.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.service.interfaces;
+
+public interface FruitFileSaver {
+    void saveToFile(String report, String filePath);
```
karma-o, line 4: There is no need to be specific on what you save here, because this interface should be reusable in other parts of the program and not just fruits
```suggestion
public interface FileWriter {
    void writeDataToFile(String data, String filePath);
```

src/main/java/core/basesyntax/service/serviceimpl/FruitRawStringParserImpl.java
```diff
@@ -0,0 +1,41 @@
+package core.basesyntax.service.serviceimpl;
+
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.exception.DataFileCorrupted;
+import core.basesyntax.service.interfaces.FruitRawStringParser;
+import core.basesyntax.service.interfaces.FruitStringsCheck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FruitRawStringParserImpl implements FruitRawStringParser {
```
karma-o, line 10: You are parsing a fruit transaction here, and implementing some parser, so I'd do something like this:
```suggestion
public class FruitTransactionParser implements TransactionParser<FruitTransaction> {
```
and generified the method in the interface to make it return `<T>`, which would make our `TransactionParser` even more handy to reuse in the future

src/main/java/core/basesyntax/service/serviceimpl/FruitRawStringParserImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final int OPERATION_NEXT_INDEX = 0;
+    private static final int FRUIT_NAME_INDEX = 1;
+    private static final int FIRST_CVS_LINE = 1;
+    private static final int FRUIT_QUANTITY_INDEX = 2;
+
+    private final FruitStringsCheck checkFruitsInFile = new FruitStringsCheckImpl();
+
+    @Override
+    public List<FruitTransactionDto> parsedFruitData(List<String> rawStrings) {
```
karma-o, line 19: `persedFruitData` is a noun, method names should be verbs, `Pareser` -> `parse`
```suggestion
    public List<FruitTransactionDto> parse(List<String> rawStrings) {
```

src/main/java/core/basesyntax/service/serviceimpl/FruitStringsCheckImpl.java
```diff
@@ -0,0 +1,37 @@
...
+                + columns.length
+                + " is present");
+            }
+            for (String column : columns) {
+                if (column == null || column.trim().isEmpty()) {
+                    throw new DataFileCorrupted("Data in column is absent");
+                }
+            }
+        }
+        return true;
```
karma-o, line 35: all of that is so unnecessary, would have been enough to make a private method in the transaction parser where you check the length of the list and that is it
There are cases when we can bring out the validation logic outside but we would name it like `RawDataValidator` with method `validate()` and your class and method names just do not represent what it actually does, or are not completely understandable

src/main/java/core/basesyntax/service/serviceimpl/CheckFilePathImpl.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.service.serviceimpl;
+
+import core.basesyntax.service.interfaces.CheckFilePath;
+
+public class CheckFilePathImpl implements CheckFilePath {
+    @Override
+    public boolean checkFilePath(String filePath) {
+        return (filePath == null || filePath.isEmpty());
+    }
```
karma-o, line 9: do we really need a whole class for that?

src/main/java/core/basesyntax/service/serviceimpl/FruitFileReaderImpl.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.List;
+
+public class FruitFileReaderImpl implements FruitFileReader {
+    private CheckFilePathImpl filePathExist = new CheckFilePathImpl();
+
+    @Override
+    public List<String> readFile(String filePath) {
+        if (filePathExist.checkFilePath(filePath)) {
+            throw new NoFileToReadException("File path is empty cannot read data");
+        }
```
karma-o, line 17: this is redundant, there is a try-catch for that which will handle this situation

src/main/java/core/basesyntax/service/serviceimpl/FruitFileReaderImpl.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax.service.serviceimpl;
+
+import core.basesyntax.exception.NoFileToReadException;
+import core.basesyntax.service.interfaces.FruitFileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class FruitFileReaderImpl implements FruitFileReader {
```
karma-o, line 10: Again, there is not logic that works only with fruits, it is universal and reusable with any file, so no need to specify that
```suggestion
public class FileReaderImpl implements FileReader {
```

src/main/java/core/basesyntax/service/serviceimpl/FruitFileSaverImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import java.io.IOException;
+
+public class FruitFileSaverImpl implements FruitFileSaver {
+    private CheckFilePathImpl filePathExist = new CheckFilePathImpl();
+
+    @Override
+    public void saveToFile(String report, String filePath) {
+        if (filePathExist.checkFilePath(filePath)) {
+            throw new NoFileToReadException("File path is invalid cannot read data");
+        }
```
karma-o, line 16: this is also redundant

src/main/java/core/basesyntax/service/serviceimpl/FruitRawStringParserImpl.java
```diff
@@ -0,0 +1,41 @@
...
+import core.basesyntax.exception.DataFileCorrupted;
+import core.basesyntax.service.interfaces.FruitRawStringParser;
+import core.basesyntax.service.interfaces.FruitStringsCheck;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FruitRawStringParserImpl implements FruitRawStringParser {
+    private static final int OPERATION_NEXT_INDEX = 0;
+    private static final int FRUIT_NAME_INDEX = 1;
+    private static final int FIRST_CVS_LINE = 1;
```
karma-o, line 13:
```suggestion
    private static final int FIRST_CSV_LINE = 1;
```

src/main/resources/fruitts.csv
```diff
@@ -0,0 +1,10 @@
+type,fruit,quantity
+b,banana,20
+b,apple,100
+s,banana,100
+p,banana,10
+s, kiwi, 20
```
karma-o, line 6: without spaces is a standard format, no need to complicate things
```suggestion
s,kiwi,20
```

src/main/java/core/basesyntax/service/serviceimpl/FruitRawStringParserImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final int FRUIT_QUANTITY_INDEX = 2;
+
+    private final FruitStringsCheck checkFruitsInFile = new FruitStringsCheckImpl();
+
+    @Override
+    public List<FruitTransactionDto> parsedFruitData(List<String> rawStrings) {
+        if (!checkFruitsInFile.checkFruitQuantity(rawStrings)) {
+            throw new DataFileCorrupted("Some data in file is missing");
+        }
+        var fruitDto = new ArrayList<FruitTransactionDto>(rawStrings.size() - FIRST_CVS_LINE);
```
karma-o, line 23: this is a list of transactions, not a fruit dto
```suggestion
        var transactions = new ArrayList<FruitTransactionDto>(rawStrings.size() - FIRST_CVS_LINE);
```

src/main/java/core/basesyntax/service/serviceimpl/FruitRawStringParserImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        for (int i = FIRST_CVS_LINE; i < rawStrings.size(); i++) {
+            String line = rawStrings.get(i);
+            String[] columns = line.split(",");
+            try {
+                int fruitQuantity = Integer.parseInt(columns[FRUIT_QUANTITY_INDEX].trim());
+                var dto = new FruitTransactionDto(columns[OPERATION_NEXT_INDEX].trim(),
+                        columns[FRUIT_NAME_INDEX].trim(),
+                        fruitQuantity);
```
karma-o, line 31: let's be consistent with the variables and constructor, either create the variables of fruitName, quantity and operation  and then create objects with them, or create the object and pass columns[index] right away for all fields in the constructor

src/main/java/core/basesyntax/service/serviceimpl/FruitRawStringParserImpl.java
```diff
@@ -0,0 +1,41 @@
...
+                fruitDto.add(dto);
+            } catch (IllegalArgumentException e) {
```
karma-o, line 33: we should not catch runtime exceptions, we should prevent them from happening or ignore to show user what the problem is

src/main/java/core/basesyntax/service/serviceimpl/FruitRawStringParserImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public List<FruitTransactionDto> parsedFruitData(List<String> rawStrings) {
+        if (!checkFruitsInFile.checkFruitQuantity(rawStrings)) {
+            throw new DataFileCorrupted("Some data in file is missing");
+        }
+        var fruitDto = new ArrayList<FruitTransactionDto>(rawStrings.size() - FIRST_CVS_LINE);
+        for (int i = FIRST_CVS_LINE; i < rawStrings.size(); i++) {
+            String line = rawStrings.get(i);
+            String[] columns = line.split(",");
```
karma-o, line 26: make coma a constant

src/main/java/core/basesyntax/service/serviceimpl/FruitReportCreateImpl.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax.service.serviceimpl;
+
+import core.basesyntax.exception.NegativeBalanceException;
+import core.basesyntax.service.interfaces.FruitReportCreate;
+import java.util.HashMap;
+import java.util.Map;
+
+public class FruitReportCreateImpl implements FruitReportCreate {
+    private static final String SEMICOL = ",";
```
karma-o, line 9: semicolumn is this - `;`, what you have here is just coma

src/main/java/core/basesyntax/service/serviceimpl/FruitReportCreateImpl.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public String createReport(HashMap<String, Integer> fruitQuantity) {
```
karma-o, line 11: better to add a StorageDao as a field of this class and call it to get the storage info instead of passing it as a param here
```suggestion
    public String createReport() {
```

src/main/java/core/basesyntax/service/serviceimpl/FruitReportCreateImpl.java
```diff
@@ -0,0 +1,30 @@
...
+        StringBuilder builder = new StringBuilder();
+        builder.append("fruit").append(SEMICOL).append("quantity").append(System.lineSeparator());
```
karma-o, line 13: no need to append separate words, jsut have a while first row as a constant and put it in the constructor of the StringBuilder instead of appending

src/main/java/core/basesyntax/service/serviceimpl/FruitReportCreateImpl.java
```diff
@@ -0,0 +1,30 @@
...
+        for (Map.Entry<String, Integer> fruits: fruitQuantity.entrySet()) {
+            if (fruits.getValue() < 0) {
```
karma-o, line 15: we should prevent this situation in the `PurchaseOperationHandler`, by throwing an exception there, where it is closer to the reason why it happened, on the attempt to buy more fruits than we have available in the storage

src/main/java/core/basesyntax/service/strategy/FruitStrategy.java
karma-o: Agree, not resolved, we do not need this `isApplicable()` method. 
Use`Map<Operation, OperationHandler>` instead of this list to easily get handler by operation

src/main/java/core/basesyntax/service/strategy/FruitStrategy.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.List;
+
+public class FruitStrategy {
+    private final List<OperationHandler> handlers;
+
+    public FruitStrategy(List<OperationHandler> handlers) {
+        this.handlers = handlers;
+    }
+
+    public List<OperationHandler> getHandlers(FruitTransactionDto dto) {
```
karma-o, line 13: Strategy's main point is to give you the correct handler depending on some value, so:
```suggestion
    public OperationHandler findHandlerFor(FruitTransactionDto dto) {
```
or
```suggestion
    public OperationHandler findHandlerFor(Operation operation) {
```

src/main/java/core/basesyntax/service/strategy/OperationHandler.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.service.strategy;
+
+import core.basesyntax.dto.FruitTransactionDto;
+import java.util.HashMap;
+
+public interface OperationHandler {
+    HashMap<String, Integer> apply(FruitTransactionDto dto);
```
karma-o, line 7: handler -> handle, we also do not need to return anything, just update storage in it
```suggestion
public interface OperationHandler {
    void handle(FruitTransactionDto dto);
```

src/main/java/core/basesyntax/service/strategy/strategyimpl/BalanceOperation.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax.service.strategy.strategyimpl;
+
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.dto.FruitTransactionDto;
+import core.basesyntax.service.strategy.OperationHandler;
+import java.util.HashMap;
+
+public class BalanceOperation implements OperationHandler {
+    private final StorageDao actionDB;
```
karma-o, line 10: this is confusing, same for all handlers
```suggestion
    private final StorageDao storageDao;
```

src/main/java/core/basesyntax/service/strategy/strategyimpl/SupplyOperation.java
```diff
@@ -0,0 +1,33 @@
...
+    public HashMap<String, Integer> apply(FruitTransactionDto dto) {
+        try {
+            HashMap<String, Integer> oldFrutitValue = actionDB.get(dto);
+            FruitTransactionDto newFruitValue = new FruitTransactionDto(dto.operationType(),
+                    dto.fruitName(),
+                    oldFrutitValue.get(dto.fruitName()) + dto.quantity());
+            return actionDB.change(newFruitValue);
+        } catch (NullPointerException e) {
+            return actionDB.add(dto);
+        }
```
karma-o, line 26: do not catch nullpointer, this is very bad, we should prevent it from happening

src/main/java/core/basesyntax/service/strategy/strategyimpl/PurchaseOperation.java
```diff
@@ -0,0 +1,37 @@
...
+            HashMap<String,Integer> oldFrutitValue = actionDB.get(dto);
+            FruitTransactionDto newFruitValue = new FruitTransactionDto(dto.operationType(),
+                    dto.fruitName(),
+                    oldFrutitValue.get(dto.fruitName()) - dto.quantity());
+            return actionDB.change(newFruitValue);
+        } catch (NullPointerException e) {
+            throw new DataFileCorrupted("Trying to purchase fruits that we dont have in Storage"
+                    + " this means there wasnt any supply or balance operations"
+                    + "in the incoming file");
+        }
```
karma-o, line 29: same here

src/main/java/core/basesyntax/service/strategy/strategyimpl/ReturnOperation.java
```diff
@@ -0,0 +1,35 @@
...
+        try {
+            HashMap<String,Integer> oldFrutitValue = actionDB.get(dto);
+            FruitTransactionDto newFruitValue = new FruitTransactionDto(dto.operationType(),
+                    dto.fruitName(),
+                    oldFrutitValue.get(dto.fruitName()) + dto.quantity());
+            return actionDB.change(newFruitValue);
+        } catch (NullPointerException e) {
+            throw new DataFileCorrupted("Trying to return fruits but"
+                    + " there was no fruit in balance or supply which were purchased");
+        }
```
karma-o, line 28: same

src/main/java/core/basesyntax/service/strategy/strategyimpl/ReturnOperation.java
```diff
@@ -0,0 +1,35 @@
...
+    @Override
+    public HashMap<String, Integer> apply(FruitTransactionDto dto) {
+        try {
+            HashMap<String,Integer> oldFrutitValue = actionDB.get(dto);
+            FruitTransactionDto newFruitValue = new FruitTransactionDto(dto.operationType(),
+                    dto.fruitName(),
+                    oldFrutitValue.get(dto.fruitName()) + dto.quantity());
+            return actionDB.change(newFruitValue);
+        } catch (NullPointerException e) {
+            throw new DataFileCorrupted("Trying to return fruits but"
```
karma-o, line 26: this NPE has nothing to do with the corrupted file, which is misguiding

karma-o (changes requested):

src/main/java/core/basesyntax/service/serviceimpl/CheckFilePathImpl.java
azzilian: Well I sued this check in two other classes and planned to use in tests but I guess there is no need of it now)

src/main/java/core/basesyntax/dao/StorageDaoImpl.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Override
+    public HashMap<String, Integer> get(String fruitName) {
+        Integer quantity = Storage.fruitsQuantity.get(fruitName);
+        if (quantity != null) {
+            HashMap<String, Integer> result = new HashMap<>();
+            result.put(fruitName, quantity);
+            return result;
+        }
+        return null;
```
karma-o, line 22: why not just return the quantity instead of map here? seems more reasonable

src/main/java/core/basesyntax/service/interfaces/TransactionParser.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service.interfaces;
+
+import core.basesyntax.dto.FruitTransactionDto;
+import java.util.List;
+
+public interface TransactionParser<T> {
+    List<FruitTransactionDto> parse(List<String> rawString);
```
karma-o, line 7: It'll become a Fuit transaction in you implementation if you will pass `FruitTransaction` as `T` in the `<>` operator
```suggestion
    List<T> parse(List<String> rawString);
```

src/main/java/core/basesyntax/service/serviceimpl/FruitTransactionParser.java
```diff
@@ -0,0 +1,47 @@
...
+            String fruitName = columns[FRUIT_NAME_INDEX].trim();
+            int fruitQuantity = Integer.parseInt(columns[FRUIT_QUANTITY_INDEX].trim());
+            var dto = new FruitTransactionDto(operation, fruitName, fruitQuantity);
+            transactions.add(dto);
+        }
+        return transactions;
+    }
+
+    private void validateRawData(String[] columns) {
+        if (columns.length <= FRUIT_QUANTITY_INDEX) {
```
karma-o, line 38: we shouldn't reuse the constant just because it has the same value that we need, it should bring us more context and understanding, but in that case it just brings more confusion
```suggestion
        if (columns.length <= MIN_COLUMNS_NUMBER) {
```

src/main/java/core/basesyntax/service/serviceimpl/FruitTransactionParser.java
```diff
@@ -0,0 +1,47 @@
...
+            throw new DataFileCorruptedException("Invalid number of columns in data");
+        }
+        for (String column : columns) {
+            if (column == null || column.trim().isEmpty()) {
```
karma-o, line 42: some hint for the future, you can use `StringUtils.isBlank(String)` instead

src/main/java/core/basesyntax/service/strategy/OperationHandler.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.service.strategy;
+
+import core.basesyntax.dto.FruitTransactionDto;
+
+public interface OperationHandler {
+    void handle(FruitTransactionDto dto);
+
+    boolean isApplicable(FruitTransactionDto dto);
```
karma-o, line 8:
```suggestion
```

src/main/java/core/basesyntax/service/strategy/FruitStrategy.java
```diff
@@ -0,0 +1,21 @@
...
+        this.handlers = handlers;
+    }
+
+    public OperationHandler findHandlerFor(FruitTransactionDto dto) {
+        for (OperationHandler handler : handlers) {
+            if (handler.isApplicable(dto)) {
+                return handler;
+            }
+        }
+        return null;
```
karma-o, line 19: Use `Map<Operation, OperationHandler>` to get the needed handler by operation enum with simply calling `map.get(operation)`
and initialize that map in main and pass it here in the constructor

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,72 @@
...
+        TransactionParser parser = new FruitTransactionParser();
+        return parser.parse(fileData);
+    }
+
+    private static FruitStrategy initializeStrategy() {
+        var balance = new BalanceOperation(storageDao);
+        var supply = new SupplyOperation(storageDao);
+        var returns = new ReturnOperation(storageDao);
+        var purchase = new PurchaseOperation(storageDao);
+        List<OperationHandler> handlers = List.of(balance, returns, purchase, supply);
```
karma-o, line 59: use map instead and `Map.of()` to initialize it

karma-o (changes requested):
Good job! 🔥🚒 just a couple very small things left

src/main/java/core/basesyntax/service/serviceimpl/FruitTransactionParser.java
azzilian: I need new library to add to dependecy then?

src/main/java/core/basesyntax/dao/StorageDaoImpl.java
azzilian: Well I got some test but Ok I can fix it

src/main/java/core/basesyntax/service/serviceimpl/FruitTransactionParser.java
karma-o: I thought it was default java method but looks like it is from `apache-commons` 3 or 4th version, try to google, cause I do not remember for sure 😄

PR #1096, opened 2024-03-19
implemented fruit shop

VolandevlodD:
Changes Made:
Added classes for CSV data parsing (CsvParser) and report generation (ReportCreatorImpl).
Implemented handling of fruit transactions using strategies (OperationStrategyImpl) and handlers (BalanceHandler, PurchaseHandler, ReturnHandler, SupplyHandler).
Updated the Main class to utilize the newly added functionality for processing data from input CSV files and generating reports.

### DIFF
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,0 +1,53 @@
+package core.basesyntax;
+
+import core.basesyntax.dao.CsvDao;
+import core.basesyntax.dao.FruitShopDao;
+import core.basesyntax.db.FruitStorage;
+import core.basesyntax.dto.FruitTransaction;
+import core.basesyntax.service.FruitTransactionParser;
+import core.basesyntax.service.ReportCreator;
+import core.basesyntax.service.impl.CsvParser;
+import core.basesyntax.service.impl.ReportCreatorImpl;
+import core.basesyntax.strategy.OperationStrategy;
+import core.basesyntax.strategy.OperationStrategyImpl;
+import core.basesyntax.strategy.handler.OperationHandler;
+import core.basesyntax.strategy.handler.impl.BalanceHandler;
+import core.basesyntax.strategy.handler.impl.PurchaseHandler;
+import core.basesyntax.strategy.handler.impl.ReturnHandler;
+import core.basesyntax.strategy.handler.impl.SupplyHandler;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+public class Main {
+    public static void main(String[] args) {
+        FruitShopDao fruitShopDao = new CsvDao();
+
+        FruitStorage fruitStorage = new FruitStorage(new TreeMap<>());
+
+        Map<FruitTransaction.Operation, OperationHandler> operationHandlerMap = new HashMap<>();
+        operationHandlerMap.put(FruitTransaction.Operation.BALANCE, new BalanceHandler(fruitStorage));
+        operationHandlerMap.put(FruitTransaction.Operation.PURCHASE, new PurchaseHandler(fruitStorage));
+        operationHandlerMap.put(FruitTransaction.Operation.RETURN, new ReturnHandler(fruitStorage));
+        operationHandlerMap.put(FruitTransaction.Operation.SUPPLY, new SupplyHandler(fruitStorage));
+
+        Path readPath = Path.of("src/main/resources/input.csv");
+        Path writePath = Path.of("src/main/resources/report.csv");
+
+        FruitTransactionParser fruitTransactionParser = new CsvParser();
+        OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlerMap);
+
+        List<String> data = fruitShopDao.loadDataFromFile(readPath);
+        List<FruitTransaction> transactions = fruitTransactionParser.parseTransactions(data);
+        OperationHandler operationHandler;
+        for (var transaction : transactions) {
+            operationHandler = operationStrategy.findHandler(transaction.operation());
+            operationHandler.handle(transaction.fruit(), transaction.quantity());
+        }
+
+        ReportCreator reportCreator = new ReportCreatorImpl();
+        fruitShopDao.saveDataToFile(writePath, reportCreator.create(fruitStorage));
+    }
+}
--- a/src/main/java/core/basesyntax/dao/CsvDao.java
+++ b/src/main/java/core/basesyntax/dao/CsvDao.java
@@ -1,0 +1,30 @@
+package core.basesyntax.dao;
+
+import core.basesyntax.exception.CsvFileException;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class CsvDao implements FruitShopDao {
+
+    @Override
+    public List<String> loadDataFromFile(Path readPath) {
+        try {
+            List<String> lines = Files.readAllLines(readPath);
+            lines.remove(0);
+            return lines;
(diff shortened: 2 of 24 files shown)


slizko1 (approved):
Nice job!

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,57 @@
...
+        FruitTransactionParser fruitTransactionParser = new CsvParser();
+        OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlers);
+
+        List<String> data = fruitShopDao.loadDataFromFile(readPath);
+        List<FruitTransaction> transactions = fruitTransactionParser.parseTransactions(data);
+        OperationHandler operationHandler;
+        for (var transaction : transactions) {
+            operationHandler = operationStrategy.findHandler(transaction.operation());
+            operationHandler.handle(transaction.fruit(), transaction.quantity());
+        }
```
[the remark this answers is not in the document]
VolandevlodD: I considered it, but I kept it this way because the readme or common mistakes mentioned that all services should be called in the main to avoid connecting or binding services. I toyed with the idea of creating a controller to link certain services, simply calling them and returning values, but I'm unsure if it's the best approach.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,57 @@
...
+
+        FruitTransactionParser fruitTransactionParser = new CsvParser();
+        OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlers);
+
+        List<String> data = fruitShopDao.loadDataFromFile(readPath);
+        List<FruitTransaction> transactions = fruitTransactionParser.parseTransactions(data);
+        OperationHandler operationHandler;
+        for (var transaction : transactions) {
+            operationHandler = operationStrategy.findHandler(transaction.operation());
+            operationHandler.handle(transaction.fruit(), transaction.quantity());
```
karma-o, line 51:
```suggestion
            operationHandler.handle(transaction);
```

src/main/java/core/basesyntax/Main.java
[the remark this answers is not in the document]
karma-o: I think it would be nice to have some `TransactionProcessor` with method `process(List<Transactions>)` to encapsulate this loop and a strategy there

src/main/java/core/basesyntax/dao/FruitShopDao.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.dao;
+
+import java.nio.file.Path;
+import java.util.List;
+
+public interface FruitShopDao {
```
karma-o, line 6: * interface by the look of it's methods has nothing to do with the fruits so far, it jsut reads and writes files, so it should be named correctly
```suggestion
public interface FileService {
```
* overall combining this sort of logic can be possible, but it's better to separate this into 2 interfaces: `FileReader` & `FileWriter` as we are practicing the **SOLID** principles in this HW, and we need t keep up with that `S`

src/main/java/core/basesyntax/dao/CsvDao.java
```diff
@@ -0,0 +1,34 @@
...
+public class CsvDao implements FruitShopDao {
+
+    private static final int FIRST_LINE = 0;
+
+    @Override
+    public List<String> loadDataFromFile(Path readPath) {
+        try {
+            List<String> lines = Files.readAllLines(readPath);
+            if (!lines.isEmpty() && lines.get(FIRST_LINE).matches("\\D+,\\D+,\\D+")) {
+                lines.remove(FIRST_LINE);
```
karma-o, line 18: this custom line removal has nothing to do with the responsibility of reading the file, this class should just read/write the data it has received without any modifications. 
Because it'll be really confusing if you or you colleague were to use that class somewhere else, reading some data, facing the problem, and after several hours of debugging - you find out that it is removing the first line, and nobody really knows why 😄 because it just says `readDataFromFile()` and nothing about the skip of the line

src/main/java/core/basesyntax/db/FruitStorage.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.db;
+
+import core.basesyntax.models.Fruit;
+import java.util.Map;
+
+public record FruitStorage(Map<Fruit, Integer> storage) {
```
karma-o, line 6: `record` is usually used in the DTO classes and makes them immutable as far as I am aware, in this case `Storage` acts the role of the DB, or DAO, or Service, so let's make it a usual calss with `public/private static final Map`

src/main/java/core/basesyntax/models/Fruit.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax.models;
+
+public record Fruit(String fruitName) implements Comparable<Fruit> {
+
+    @Override
+    public int compareTo(Fruit fruit) {
+        return this.fruitName.compareTo(fruit.fruitName);
+    }
```
karma-o, line 8: I do not think we need all of that, this class has only 1 string field - so it can easily be replaced with just string in storage, or if you want to make use of it, add `amount` or `quantity` field here, and make `storage` not a `map`, but a `list of fruits`.
But I would prefer the first approach

src/main/java/core/basesyntax/service/impl/CsvParser.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.dto.FruitTransaction;
+import core.basesyntax.models.Fruit;
+import core.basesyntax.service.FruitTransactionParser;
+import java.util.List;
+
+public class CsvParser implements FruitTransactionParser {
```
karma-o, line 8: I'd revert the naming or did something like this:
```suggestion
public class FruitTransactionParser implements TransactionParser<FruitTransaction> {
```
because both interface and and impl class sound concrete and interface name should be more abstract, here we have the ability to make some other `MeatTransaction` and `MeatTransactionParser` or even parametrize `TransactionParser` and its method to return `<T>` for example, to make it even more reusable in the future

src/main/java/core/basesyntax/service/impl/ReportCreatorImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class ReportCreatorImpl implements ReportCreator {
+
+    public static final String TITLE = "fruit,quantity";
+    public static final String COMMA = ",";
+
+    @Override
+    public List<String> create(FruitStorage storage) {
```
karma-o, line 14: we can have a direct access to the storage instead of passing it as a parameter, if we use it's static map or pass some `StorageDao`(if we would have one) to the constructor of the `ReportCreator`

karma-o (changes requested):

karma-o:
Looks good so far. left not too much comments, so good work!

src/main/java/core/basesyntax/service/impl/TransactionProcessorImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public TransactionProcessorImpl(
+            Map<Operation, OperationHandler> operationHandlers) {
+        this.operationHandlers = operationHandlers;
+    }
+
+    @Override
+    public void process(List<ProductTransaction> productTransactions) {
+        OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlers);
+        OperationHandler operationHandler;
```
karma-o, line 23: Why new Strategy object every time? make it a field or just remove that class at all and use the map directly, because basically all it does is encapsulates this map and I do not really think it is necessary

karma-o (approved):
Approved with one comment

PR #1105, opened 2024-03-20
java-fruit-shop-solution

kaiiiseeel:


src/main/java/core/basesyntax/service/impl/StrategyServiceImpl.java
d1sam, line 1: if we create new implementation of OperationHandler we will need to add one case. Please rewrite strategy with Map<Operation, OperationHandler> strategyMap and then get necessary OperationHandler from this map regarding to the operation.

src/main/java/core/basesyntax/service/impl/FruitDataParser.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final String DELIMITER = ",";
+
+    @Override
+    public List<FruitTransactionDto> parse(List<String> rawData) {
+        List<FruitTransactionDto> fruitsList = new ArrayList<>(rawData.size());
+        for (int i = 1; i < rawData.size(); i++) {
+            String line = rawData.get(i);
+            String[] strings = line.split(DELIMITER);
+            FruitTransactionDto dto =
+                    new FruitTransactionDto(strings[0], strings[1], Integer.parseInt(strings[2]));
```
d1sam, line 18: please avoid magic numbers.

src/main/java/core/basesyntax/service/impl/DataValidatorImpl.java
```diff
@@ -0,0 +1,53 @@
...
+        String operation = fruitTransactionDto.operation();
+
+        if ((operation.equalsIgnoreCase(BALANCE)
+                || operation.equalsIgnoreCase(SUPPLY)
+                || operation.equalsIgnoreCase(RETURN)
+                || operation.equalsIgnoreCase(PURCHASE))) {
+            return;
+        }
+        throw new InvalidOperationException("Invalid operation: " + operation);
+    }
```
d1sam, line 35:
```suggestion
```
let's better remove this operation validation and throw InvalidOperationException , if there is no operation in `Map<Operation, OperationHandler> strategyMap`

src/main/java/core/basesyntax/service/impl/DataValidatorImpl.java
```diff
@@ -0,0 +1,53 @@
...
+import core.basesyntax.exception.InvalidOperationException;
+import core.basesyntax.exception.InvalidQuantityException;
+import core.basesyntax.service.DataValidator;
+
+public class DataValidatorImpl implements DataValidator {
+    private static final String BALANCE = FruitDataParser.Operation.BALANCE.getCode();
+    private static final String SUPPLY = FruitDataParser.Operation.SUPPLY.getCode();
+    private static final String RETURN = FruitDataParser.Operation.RETURN.getCode();
+    private static final String PURCHASE = FruitDataParser.Operation.PURCHASE.getCode();
+    private static final String SPECIAL_CHARACTERS = ".*[$&+,/:;=?@#|'<>.-^*()%!].*";
```
d1sam, line 14: what will be if your String will contain `_`? let's better check that string contains only numbers

src/main/java/core/basesyntax/service/impl/FileWriterImpl.java
```diff
@@ -0,0 +1,26 @@
...
+    private static final String HEADER = "fruit,quantity";
+
+    @Override
+    public boolean write(List<String> processedData, String fileName) {
+        try (FileWriter fileWriter = new FileWriter(fileName)) {
+            fileWriter.write(HEADER);
+            for (String data : processedData) {
+                fileWriter.write(LINE_SEPARATOR);
+                fileWriter.write(data);
+            }
```
d1sam, line 20:
```suggestion
            fileWriter.write(HEADER);
            fileWriter.write(LINE_SEPARATOR);
            for (String data : processedData) {
                fileWriter.write(data);
                fileWriter.write(LINE_SEPARATOR);
            }
```

d1sam: your file should have empty last line

d1sam (changes requested):
Good job, but there some changes, that should be done!

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    public static final Map<String, Integer> fruitStorage = new HashMap<>();
+
+    public Map<String, Integer> getStorageData() {
+        return fruitStorage;
+    }
```
karma-o, line 11: `fruitStorage` is public so no need in that method
```suggestion
```

src/main/java/core/basesyntax/service/impl/CsvFruitDataReaderImpl.java
```diff
@@ -0,0 +1,22 @@
...
+import java.nio.file.Path;
+import java.util.List;
+
+public class CsvFruitDataReaderImpl implements FruitDataReaderService {
+    @Override
+    public List<FruitTransactionDto> readData(String fileName) {
+        try {
+            List<String> strings = Files.readAllLines(Path.of(fileName));
+            FruitDataParser fruitDataParser = new FruitDataParser();
+            return fruitDataParser.parse(strings);
```
karma-o, line 17: first of all services like this parser shouldn't be created in a method like this, they should be a class fields. But in that case it should not be a part of FileReader at all, because it has 2 responsibilities - read data from file and parse in into transactions, so we should split them into 2

src/main/java/core/basesyntax/service/DataValidator.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax.service;
+
+import core.basesyntax.dto.FruitTransactionDto;
+
+public interface DataValidator {
+    void validate(FruitTransactionDto fruitTransactionDto);
+
+    void validateFruitType(FruitTransactionDto fruitTransactionDto);
+
+    void validateQuantity(FruitTransactionDto fruitTransactionDto);
```
karma-o, line 10: no point in having them public if you call them both in `validate()`, just make them private in the impl class
```suggestion
```

src/main/java/core/basesyntax/service/impl/FileWriterImpl.java
```diff
@@ -0,0 +1,27 @@
...
+
+import core.basesyntax.exception.CantWriteToFileException;
+import core.basesyntax.service.FileWriterService;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.List;
+
+public class FileWriterImpl implements FileWriterService {
+    private static final String LINE_SEPARATOR = System.lineSeparator();
+    private static final String HEADER = "fruit,quantity";
```
karma-o, line 11: adding this header is not the responsibility of the file writer, it should just right the data it recieves withouth any modifications to it

src/main/java/core/basesyntax/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,19 @@
...
+import core.basesyntax.service.ReportGenerator;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class ReportGeneratorImpl implements ReportGenerator {
+    private static final String DELIMITER = ",";
+
+    @Override
+    public List<String> createReport(Map<String, Integer> fruitTransactionDtoList) {
```
karma-o, line 12: Bad name + Better to create report using the direct Storage access instead of passing it as a parameter here
```suggestion
    public List<String> createReport() {
```

src/main/java/core/basesyntax/dto/FruitTransactionDto.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.dto;
+
+public record FruitTransactionDto(
+        String operation,
```
karma-o, line 4: `Operation` should be enum, check readme for example

src/main/java/core/basesyntax/service/impl/StrategyServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+import core.basesyntax.service.DataValidator;
+import core.basesyntax.service.OperationHandler;
+import core.basesyntax.service.OperationStrategy;
+import core.basesyntax.service.StrategyService;
+import java.util.List;
+import java.util.Map;
+
+public class StrategyServiceImpl implements StrategyService, OperationStrategy {
+    private static final DataValidator DATA_VALIDATOR = new DataValidatorImpl();
+    private final Map<String, OperationHandler> operations;
```
karma-o, line 14:
```suggestion
    private final Map<Operation, OperationHandler> operationStrategyMap;

```

src/main/java/core/basesyntax/service/impl/StrategyServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    public void processData(List<FruitTransactionDto> fruitTransactionDtoList,
+                            Map<String, OperationHandler> operationTypeList) {
+        for (FruitTransactionDto fruitTransactionDto : fruitTransactionDtoList) {
+            DATA_VALIDATOR.validate(fruitTransactionDto);
+            getHandlers(fruitTransactionDto).handle(fruitTransactionDto);
+        }
+    }
+
+    @Override
+    public OperationHandler getHandlers(FruitTransactionDto fruitTransactionDto) {
```
karma-o, line 30:
```suggestion
    public OperationHandler getHandler(FruitTransactionDto fruitTransactionDto) {

```

karma-o (changes requested):

PR #1109, opened 2024-03-25
built and implemented fruit shop architecture

ghost:


src/main/java/mate/fruitshop/service/strategy/TransactionHandlerStrategy.java
```diff
@@ -0,0 +1,17 @@
...
+public class TransactionHandlerStrategy {
+    public static final Map<FruitTransaction.Operation, FruitTransactionHandler>
+            TRANSACTIONS_MAP = new HashMap<>();
+
+    static {
+        TRANSACTIONS_MAP.put(FruitTransaction.Operation.BALANCE, new FruitBalanceHandler());
+        TRANSACTIONS_MAP.put(FruitTransaction.Operation.SUPPLY, new FruitSupplyHandler());
+        TRANSACTIONS_MAP.put(FruitTransaction.Operation.PURCHASE, new FruitPurchaseHandler());
+        TRANSACTIONS_MAP.put(FruitTransaction.Operation.RETURN, new FruitReturnHandler());
+    }
```
okuzan, line 16: you used constant case for this map, but it's not a constant, because static final isn't enough, it should be deeply immutable
You can use Map.of, it will create immutable map, and you won't need static init block

src/main/java/mate/fruitshop/service/FruitService.java
```diff
@@ -0,0 +1,46 @@
...
+    public FruitService(FruitTransactionDao dao) {
+        this.dao = dao;
+    }
+
+    public Map<String, Integer> calculateFruitsLeft() {
+        List<FruitTransaction> transactions = dao.getAll();
+        Map<String, Integer> fruitQuantityMap = new HashMap<>();
+        int currentFruitQuantity;
+
+        for (FruitTransaction transaction: transactions) {
```
okuzan, line 25:
```suggestion
        for (FruitTransaction transaction : transactions) {
```
don't forget to autoformat your files (Ctrl Alt L)

src/main/java/mate/fruitshop/service/FruitService.java
```diff
@@ -0,0 +1,46 @@
...
+            currentFruitQuantity = TRANSACTIONS_MAP.get(transaction.getOperation())
+                    .conductTransaction(transaction,
+                            fruitQuantityMap.getOrDefault(transaction.getFruit(), 0));
+            fruitQuantityMap.put(transaction.getFruit(), currentFruitQuantity);
```
okuzan, line 29:
```suggestion
            fruitQuantityMap.compute(transaction.getFruit(), (k, v) ->
                    TRANSACTIONS_MAP.get(transaction.getOperation())
                            .conductTransaction(transaction, v == null ? 0 : v));
```
there are a lot of useful methods in hashmap, I think we can try smth like this

src/main/java/mate/fruitshop/service/FruitService.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+        return fruitQuantityMap;
+    }
+
+    public String createReport(Map<String, Integer> fruitQuantityMap) {
+        StringBuilder reportBuilder = new StringBuilder();
+        reportBuilder.append(FRUIT + ", " + QUANTITY);
+        fruitQuantityMap.entrySet().stream()
+                        .forEach(e -> reportBuilder.append(System.lineSeparator())
+                                        .append(e.getKey()).append(", ").append(e.getValue()));
```
okuzan, line 39: use Map's forEach instead

src/main/java/mate/fruitshop/model/FruitTransaction.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        public static Operation pickOperation(String type) {
+            return switch (type) {
+                case "b" -> Operation.BALANCE;
+                case "s" -> Operation.SUPPLY;
+                case "p" -> Operation.PURCHASE;
+                case "r" -> Operation.RETURN;
+                default -> throw new RuntimeException("Invalid operation type " + type);
+            };
```
okuzan, line 55: 1. Better to throw more specific exceptions where possible, here, use IllegalArgument
2. Not a flexible approach, you now need to make changes in multiple places if you decide to add/remove some enum value. Iterate through .values() instead

src/main/java/mate/fruitshop/dao/FruitTransactionDaoCsv.java
```diff
@@ -0,0 +1,65 @@
...
+import mate.fruitshop.model.FruitTransaction;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVRecord;
+
+public class FruitTransactionDaoCsv implements FruitTransactionDao {
+    public static final String INPUT_FILE_NAME = "src/main/resources/input.csv";
+    public static final String OUTPUT_FILE_NAME = "src/main/resources/output.csv";
+    public static final String OPERATION_TYPE = "type";
+    public static final String FRUIT = "fruit";
+    public static final String QUANTITY = "quantity";
```
okuzan, line 20: revise access modifiers

src/main/java/mate/fruitshop/dao/FruitTransactionDaoCsv.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public List<FruitTransaction> getAll() {
+        Reader in = getFileReader();
```
okuzan, line 24: `in` is a bad naming, just go for a fileReader or reader

okuzan (changes requested):
Clean and readable code, good job!

Vasul-Pankiv (approved):

PR #1111, opened 2024-03-29
Fruit shop

ChabVlad:


src/main/java/core/basesyntax/strategy/OperationStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+            return new SupplyOperation();
+        }
+        if (operation.getCode().equals(Operation.PURCHASE.getCode())) {
+            return new PurchaseOperation();
+        }
+        if (operation.getCode().equals(Operation.RETURN.getCode())) {
+            return new ReturnOperation();
+        }
+        throw new RuntimeException("This operation doesn't exist " + operation);
+    }
```
olekskov, line 25: 1) Make method non-static. It may look easier now but static elements are much harder to be tested and maintained properly
2) It's strange when method that accepts Operation is called getOperation. With such naming I expect to see single line of code ```return operation``` :)  Should be getOperationHandler() or anything that indicates that we will obtain special component that is ready to perform some logic using our operation
3) Please use map instead of ```if``` chain

src/main/java/core/basesyntax/service/impl/FileReaderServiceImpl.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.FileReaderService;
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class FileReaderServiceImpl implements FileReaderService {
```
olekskov, line 10: I personally do not see any sense to add Service word into name of class when there is a noun that describes action (like Reader, Writer, Processor, Handler etc.)
```suggestion
public class FileReaderImpl implements FileReader {
```

src/main/java/core/basesyntax/service/impl/FileWriterServiceImpl.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.FileWriterService;
+import java.io.FileWriter;
+import java.util.Map;
+
+public class FileWriterServiceImpl implements FileWriterService {
```
olekskov, line 7: same

src/main/java/core/basesyntax/service/impl/BalanceOperation.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.OperationService;
+
+public class BalanceOperation implements OperationService {
+    @Override
+    public void operation(String line) {
```
olekskov, line 8: naming is confusing
1) method name should contain verb to get idea what will be done
2) BalanceOperation name looks like it should extend Operation but it extends OperationService
3) OperationService is very abstract and not giving idea about purpose (looks like basic CRUD service that should perform save/get/update/delete operations on Operation model)
4) Why method accepts String. This thing should accept operation
```suggestion
public class BalanceTransactionHandler implements TransactionHandler {
    @Override
    public void process(Transaction transaction) {
```

similar for other types of operations

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class Main {
+    private static final String PATH = "src/main/java/core/basesyntax/db/data.csv";
+    private static final String REPORT = "src/main/java/core/basesyntax/db/report.csv";
+
+    public static void main(String[] args) {
+        FileWriterService writer = new FileWriterServiceImpl();
+        FruitShopService transactions = new FruitShopServiceImpl();
+        writer.writeToFile(transactions.fruitTransaction(PATH), REPORT);
+    }
```
olekskov, line 16: Your classes performs many things instead of concentration on single responsibility. Much more operations
should be here
1. Read content of file
2. Parse raw content to Transactions
3. Process transactions to change Storage state
4. Form report using new Storage state
5. Write report to file

And each of this logical steps should be performed by separate service


Please recheck task description after ```Hint``` section, there are some nice advises

olekskov (changes requested):

src/main/java/core/basesyntax/service/Parser.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.service;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.List;
+
+public interface Parser {
+    List<FruitTransaction> parseTransactions(List<String> lines);
```
olekskov, line 7: Better to specify model in class name as we may have multiple parsers in our application and this is standard to have separate parser for each one
```suggestion
public interface TransactionParser {
    List<FruitTransaction> parseAll(List<String> lines);
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,29 @@
...
+    private static final String REPORT = "src/main/java/core/basesyntax/db/report.csv";
+
+    public static void main(String[] args) {
+        FileReader reader = new FileReaderImpl();
+        FileWriter writer = new FileWriterImpl();
+        Parser parser = new ParserImpl();
+        FruitShopService transactions = new FruitShopServiceImpl();
+        List<String> dataFromFile = reader.readData(PATH);
+        List<FruitTransaction> parsedTransactions = parser.parseTransactions(dataFromFile);
+        transactions.fruitTransaction(parsedTransactions);
```
olekskov, line 26: Need to work on naming a bit to increase code clarity
```suggestion
        FruitShopService fruitShopService = new FruitShopServiceImpl();
        List<String> dataFromFile = reader.readData(PATH);
        List<FruitTransaction> parsedTransactions = parser.parseTransactions(dataFromFile);
        fruitShopService.processTransactions(parsedTransactions);
```

src/main/java/core/basesyntax/service/impl/FruitShopServiceImpl.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.FruitShopService;
+import core.basesyntax.service.TransactionHandler;
+import core.basesyntax.strategy.OperationStrategy;
+import java.util.List;
+
+public class FruitShopServiceImpl implements FruitShopService {
+    private OperationStrategy strategy = new OperationStrategy();
```
olekskov, line 10: Please use dependency injection via constructor instead of field initialization. It will make code testing and maintaining more simple

src/main/java/core/basesyntax/strategy/OperationStrategy.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class OperationStrategy {
+    private Map<String, TransactionHandler> operationMap;
+
+    public OperationStrategy() {
+        operationMap = new HashMap<>();
+        operationMap.put(Operation.BALANCE.getCode(), new BalanceTransactionHandler());
+        operationMap.put(Operation.SUPPLY.getCode(), new SupplyTransactionHandler());
+        operationMap.put(Operation.PURCHASE.getCode(), new PurchaseTransactionHandler());
+        operationMap.put(Operation.RETURN.getCode(), new ReturnTransactionHandler());
```
olekskov, line 20: Please move map initialization to Main class and pass ready to use map via constructor

src/main/java/core/basesyntax/db/report.csv
```diff
@@ -0,0 +1 @@
+{banana=152, apple=90}
```
olekskov, line 1: File foormat doesn't comply with requirement
in task description:
    fruit,quantity
    banana,152
    apple,90

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,29 @@
...
+        writer.writeToFile(Storage.storage, REPORT);
```
olekskov, line 27: Better to split logic of report generation and writing into file:
```suggestion
        String report = reportService.generate(Storage.storage);
        writer.writeToFile(report, REPORT_PATH);
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,29 @@
...
+import core.basesyntax.service.Parser;
+import core.basesyntax.service.impl.FileReaderImpl;
+import core.basesyntax.service.impl.FileWriterImpl;
+import core.basesyntax.service.impl.FruitShopServiceImpl;
+import core.basesyntax.service.impl.ParserImpl;
+import java.util.List;
+
+public class Main {
+    private static final String PATH = "src/main/java/core/basesyntax/db/data.csv";
+    private static final String REPORT = "src/main/java/core/basesyntax/db/report.csv";
```
olekskov, line 17: Better to use more precise naming
```suggestion
    private static final String DATA_SOURCE_PATH = "src/main/java/core/basesyntax/db/data.csv";
    private static final String REPORT_PATH = "src/main/java/core/basesyntax/db/report.csv";
```

olekskov (changes requested):
Much better but there are still some comments

Vasul-Pankiv (approved):
Good job!

PR #1112, opened 2024-04-01
Solution

Leanchers:


src/main/java/core/basesyntax/strategy/SupplyOperationHandler.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public SupplyOperationHandler(FruitDao fruitDao) {
+        this.fruitDao = fruitDao;
+    }
+
+    @Override
+    public void apply(FruitTransaction fruitTransaction) {
+        if (fruitDao.containsFruit(fruitTransaction.getFruit())) {
+            fruitDao.update(fruitTransaction.getFruit(),
+                    fruitDao.get(fruitTransaction.getFruit()) + fruitTransaction.getQuantity());
```
okuzan, line 17: better create a DAO method for this, inside of which use map.merge with Integer::sum as function

okuzan (changes requested):
Very good job overall! Clean code

olekskov (approved):
Good job

PR #1123, opened 2024-04-10
Solved jv-fruit-shop task:

Shimeriyu:
Created storage class simulating database of fruits. Created FruitTransaction model with Fruit and Transaction Enums Implemented services for: reading csv, generating report, saving report to csv, parsing input to Fruit Transactions, Transaction handling and strategy for it Added validation for input data and if there is enough stock of fruit for purchase transaction

### DIFF
--- a/src/main/java/fruit/shop/model/Transaction.java
+++ b/src/main/java/fruit/shop/model/Transaction.java
@@ -1,0 +1,5 @@
+package fruit.shop.model;
+
+public enum Transaction {
+    BALANCE, SUPPLY, PURCHASE, RETURN
+}
--- a/src/main/java/fruit/shop/service/InputDataValidator.java
+++ b/src/main/java/fruit/shop/service/InputDataValidator.java
@@ -1,0 +1,37 @@
+package fruit.shop.service;
+
+import fruit.shop.model.Fruit;
+import java.util.Arrays;
+import java.util.Objects;
+
+public class InputDataValidator {
+    private static final int FRUIT_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    public String[] test(String[] inputData) {
+        if (inputData.length != 3) {
+            throwException("Row contains wrong number of parameters");
+        }
+        if (Arrays.stream(inputData).anyMatch(Objects::isNull)) {
+            throwException("Row contains null value");
+        }
+        try {
+            Fruit.valueOf(inputData[FRUIT_INDEX].toUpperCase());
+        } catch (IllegalArgumentException e) {
+            throwException("Fruit: " + inputData[FRUIT_INDEX] + "is not defined in shop");
+        }
+        try {
+            Integer.parseInt(inputData[QUANTITY_INDEX]);
+        } catch (NumberFormatException e) {
+            throwException("Quantity column contains non-numerical data");
+        }
+        if (Integer.parseInt(inputData[QUANTITY_INDEX]) < 0) {
+            throwException("Quantity cannot be negative");
+        }
+        return inputData;
+    }
+
+    private void throwException(String message) {
+        throw new RuntimeException("Input csv is invalid. " + message);
+    }
+}
--- a/src/main/java/fruit/shop/service/TransactionParserServiceImpl.java
+++ b/src/main/java/fruit/shop/service/TransactionParserServiceImpl.java
@@ -1,0 +1,39 @@
+package fruit.shop.service;
+
+import fruit.shop.model.Fruit;
+import fruit.shop.model.FruitTransaction;
+import fruit.shop.model.Transaction;
+import java.util.List;
+
+public class TransactionParserServiceImpl implements TransactionParserService {
+    private static final int TRANSACTION_INDEX = 0;
+    private static final int FRUIT_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+    private static final InputDataValidator INPUT_DATA_VALIDATOR = new InputDataValidator();
+
+    @Override
+    public List<FruitTransaction> parse(List<String> transactionInput) {
+        return transactionInput.stream()
+                .map(input -> input.split(","))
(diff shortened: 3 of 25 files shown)


src/main/java/fruit/shop/model/Transaction.java
```diff
@@ -0,0 +1,5 @@
+package fruit.shop.model;
+
+public enum Transaction {
```
nowickimj, line 3: The instance of this class won't be transaction - `TransactionType` would be more descriptive.

src/main/java/fruit/shop/service/InputDataValidator.java
```diff
@@ -0,0 +1,37 @@
...
+        }
+        if (Integer.parseInt(inputData[QUANTITY_INDEX]) < 0) {
+            throwException("Quantity cannot be negative");
+        }
+        return inputData;
+    }
+
+    private void throwException(String message) {
+        throw new RuntimeException("Input csv is invalid. " + message);
+    }
```
nowickimj, line 36: How about the dedicated exception type? `TransactionInputValidationException extends RuntimeException` ?

src/main/java/fruit/shop/service/TransactionParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+                .map(input -> new FruitTransaction(
+                                Fruit.valueOf(input[FRUIT_INDEX].toUpperCase()),
+                                transactionShort(input[TRANSACTION_INDEX]),
+                                Integer.parseInt(input[QUANTITY_INDEX])
+                        )
+                )
+                .toList();
+    }
+
+    private Transaction transactionShort(String transaction) {
```
nowickimj, line 28: `parseTransactionType()` would be more descriptive

src/main/java/fruit/shop/service/transaction/PurchaseTransactionHandler.java
```diff
@@ -0,0 +1,19 @@
...
+
+    @Override
+    public void execute(FruitTransaction transaction) {
+        Integer stock = Storage.get(transaction.getFruit());
+        Integer transactionQuantity = transaction.getQuantity();
+        if (quantityValidator.validate(stock, transactionQuantity)) {
+            Storage.put(transaction.getFruit(), stock - transactionQuantity);
+            return;
+        }
+        throw new RuntimeException("Shop doesn't have enough stock of: " + transaction.getFruit());
```
nowickimj, line 17: This case is pretty trivial, but for more complex code you might want to reverse the order:
1. Go through negative case scenarios and end execution when it occurrs (`if (!quantityValidator.validate()...)`)
2. Process the positive scenario (`Storage.put(...)`)

you lose the obscure `return` call.

nowickimj (changes requested):

src/main/java/fruit/shop/service/InputDataValidator.java
```diff
@@ -0,0 +1,39 @@
...
+        }
+        if (Integer.parseInt(inputData[QUANTITY_INDEX]) < MINIMUM_QUANTITY_VALUE) {
+            throwException("Quantity cannot be negative");
+        }
+        return inputData;
+    }
+
+    private void throwException(String message) {
+        throw new TransactionInputValidationException("Input csv is invalid. " + message);
+    }
```
nowickimj, line 38: Please include the default message in exception constructor:
```java
public TransactionInputValidationException(String message) {
  super("Invalid input. " + message);
}
```
Then, throw the exception immediately within `test()` method.

Also, you can give it a bit more generic name, like `InvalidTransactionInputException` so it can also be used in greater context, see the comment below.

src/main/java/fruit/shop/service/TransactionParserServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private TransactionType parseTransactionType(String transaction) {
+        return switch (transaction) {
+            case "b" -> TransactionType.BALANCE;
+            case "s" -> TransactionType.SUPPLY;
+            case "p" -> TransactionType.PURCHASE;
+            case "r" -> TransactionType.RETURN;
+            default -> throw new RuntimeException("Input csv is invalid. Transaction: "
+                    + transaction
+                    + " is not defined in shop");
```
nowickimj, line 36: `InvalidTransactionInputException`

src/main/java/fruit/shop/service/TransactionStrategyImpl.java
```diff
@@ -0,0 +1,24 @@
...
+import fruit.shop.service.transaction.TransactionHandler;
+import java.util.Map;
+
+public class TransactionStrategyImpl implements TransactionStrategy {
+    private final Map<TransactionType, TransactionHandler> transactionHandlerMap = Map.ofEntries(
+            Map.entry(TransactionType.BALANCE, new BalanceTransactionHandler()),
+            Map.entry(TransactionType.SUPPLY, new SupplyTransactionHandler()),
+            Map.entry(TransactionType.PURCHASE, new PurchaseTransactionHandler()),
+            Map.entry(TransactionType.RETURN, new ReturnTransactionHandler())
+    );
```
nowickimj, line 18: Let's make this implementation a bit more flexible.

1. Make this field initialized in constructor
2. Create map and pass it to constructor in `main()` method.

src/main/java/fruit/shop/service/TransactionStrategy.java
```diff
@@ -0,0 +1,7 @@
+package fruit.shop.service;
+
+import fruit.shop.model.FruitTransaction;
+
+public interface TransactionStrategy {
+    void executeTransactionHandler(FruitTransaction transaction);
```
nowickimj, line 6: `handleTransaction()` - `executeTransactionHandler` is misleading as you're not passing any handler. Handler is not known to outside world, hence the name says nothing to user.

nowickimj (changes requested):

PR #1141, opened 2024-05-24
Hw jv fruit shop

snowfallen:


src/main/java/core/basesyntax/services/impl/ReportWriterImpl.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax.services.impl;
+
+import core.basesyntax.services.ReportWriter;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.IOException;
+
+public class ReportWriterImpl implements ReportWriter {
```
kerrrusha, line 8: it can write not only reports
```suggestion
public class FileWriterImpl implements FileWriter {
```

kerrrusha (approved):

PR #1143, opened 2024-05-27
MateAcademy/Task/jv-fruit-shop - Bartosz Wójcik 

banderan:


MateuszRostek (changes requested):
You have squeezed everything too tightly, you should break down the operations shown in the diagram into more services, preferably also introduce a strategy pattern for operations, create a storage (a simple HashMap will be sufficient). Below I have posted you how the structure of the program can look like after splitting it into smaller parts as an example. Also your ShopDao is doing things that should be done in these services, check out the screenshots for better understanding.

Example of how should project structure look like:
Zrzut ekranu 2024-05-28 095329

Example of how should your Dao layer look like:
image

image

And also how can storage look like:
image

PR #1146, opened 2024-06-07
solved style-code issues

Yaromyros:
First, I created packages of db, model, services, and implementations of these services. Also, I added strategy and resources. 

Briefly, my storage has a map and there are methods to add, set, get, and subtract fruits. 

FruitTransaction has each fruit object with all working fields. 

Services handled interfaces and their implementations with files, data convertion, and report generation.

Strategy is in charge of which new object-operation based on operation.

Finally, a main method that combines all of the above and resources contains two files(output and input)

### DIFF
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,0 +1,46 @@
+package core.basesyntax;
+
+import core.basesyntax.service.FileReader;
+import core.basesyntax.service.FileWriter;
+import core.basesyntax.service.impl.FileReaderImpl;
+import core.basesyntax.service.impl.FileWriterImpl;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.*;
+import core.basesyntax.service.impl.*;
+import core.basesyntax.strategy.*;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    public static void main(String[] arg) {
+        // 1. Read the data from the input CSV file
+        FileReader fileReader = new FileReaderImpl();
+        List<String> inputReport = fileReader.read("src/main/resources/reportToRead.csv");
+
+        // 2. Convert the incoming data into FruitTransactions list
+        DataConverter dataConverter = new DataConverterImpl();
+        List<FruitTransaction> transactions = dataConverter.convertToTransaction(inputReport);
+
+        // 3. Create and feel the map with all OperationHandler implementations
+        Map<FruitTransaction.Operation, OperationHandler> operationHandlers = new HashMap<>();
+        operationHandlers.put(FruitTransaction.Operation.BALANCE, new BalanceOperation());
+        operationHandlers.put(FruitTransaction.Operation.PURCHASE, new PurchaseOperation());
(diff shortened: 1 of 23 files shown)


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,55 @@
...
+import core.basesyntax.strategy.SupplyOperation;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    public static void main(String[] arg) {
+        // 1. Read the data from the input CSV file
+        FileReader fileReader = new FileReaderImpl();
+        List<String> inputReport = fileReader.read("src/main/resources/reportToRead.csv");
```
Elena-Bruyako, line 29: Let's create constants for input and output files path

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,55 @@
...
+
+        // 2. Convert the incoming data into FruitTransactions list
+        DataConverter dataConverter = new DataConverterImpl();
+        final List<FruitTransaction> transactions = dataConverter.convertToTransaction(inputReport);
+
+        // 3. Create and feel the map with all OperationHandler implementations
+        Map<FruitTransaction.Operation, OperationHandler> operationHandlers = new HashMap<>();
```
Elena-Bruyako, line 36: It's better to create separate method like getMapOperationHandler where you create a Map

src/main/java/core/basesyntax/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.service.ReportGenerator;
+import java.util.stream.Collectors;
+
+public class ReportGeneratorImpl implements ReportGenerator {
+    @Override
+    public String getReport() {
+        String header = "fruit,quantity";
```
Elena-Bruyako, line 10: Create constant `private static final String HEADER = "fruit,quantity";`

src/main/java/core/basesyntax/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,16 @@
...
+        String reportBody = Storage.getAllFruits().entrySet().stream()
+                .map(entry -> entry.getKey() + "," + entry.getValue())
```
Elena-Bruyako, line 12: Create constant `private static final String DELIMITER = ",";`

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    private static final Map<String, Integer> fruits = new HashMap<>();
+
+    public static Integer getFruitQuantity(String fruit) {
```
Elena-Bruyako, line 9: It's better to move these all logic to strategy operations

Elena-Bruyako (changes requested):

Vasul-Pankiv (approved):
Dude, nice job!

PR #1148, opened 2024-06-12
MateAcademy/Task/jv-fruit-shop - Bartosz Wójcik second solution

banderan:


MateuszRostek (approved):
Really good job! Your code for the most part looks good, due to the fact that you are facing writing tests for this task, this is where you can best check and see if your solution is completely correct and if you need to add some additional validation in some places.

Keep it in mind:
image

PR #1149, opened 2024-06-16
First Attempt

jszymanski97:
After long brake but I'm back... Hope didn't forget much

### DIFF
--- a/src/main/java/core/basesyntax/FruitTransaction.java
+++ b/src/main/java/core/basesyntax/FruitTransaction.java
@@ -1,0 +1,44 @@
+package core.basesyntax;
+
+public class FruitTransaction {
+    private Operation operation;
+    private String fruit;
+    private int quantity;
+
+    public FruitTransaction(Operation operation, String fruit, int quantity) {
+        this.operation = operation;
+        this.fruit = fruit;
+        this.quantity = quantity;
+    }
+
+    public Operation getOperation() {
+        return operation;
+    }
+
+    public void setOperation(Operation operation) {
+        this.operation = operation;
+    }
+
+    public String getFruit() {
+        return fruit;
+    }
+
+    public void setFruit(String fruit) {
+        this.fruit = fruit;
+    }
+
+    public int getQuantity() {
+        return quantity;
+    }
+
+    public void setQuantity(int quantity) {
+        this.quantity = quantity;
+    }
+
+    public enum Operation {
+        BALANCE,
+        SUPPLY,
+        PURCHASE,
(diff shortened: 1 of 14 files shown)


src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,27 @@
...
+    public List<FruitTransaction> convert(List<String> transactions) {
+        List<FruitTransaction> transactionList = new ArrayList<>();
+        for (String transaction : transactions) {
+            String[] parts = transaction.split(",");
+            if (parts.length > 0 && parts.length < 3) {
+                throw new ArrayIndexOutOfBoundsException("Invalid data entered");
+            }
+            Operation operation = new OperationStrategyImpl().getOperationFromCode(parts[0]);
+            String fruit = parts[1];
+            int quantity = Integer.parseInt(parts[2]);
```
MateuszRostek, line 22: Magic numbers should be constants

src/main/java/core/basesyntax/strategy/OperationStrategy.java
MateuszRostek, line 1: This is not how the strategy pattern should be implemented, here is a proper structure of it:
image

And one of the operation handler for example:
image

MateuszRostek: You can stick to your solution, but you should now that your solution does not follow the strategy pattern.

src/main/java/core/basesyntax/service/impl/TransitionProcessorImpl.java
```diff
@@ -0,0 +1,25 @@
...
+    public void processTransaction(Map<String, Integer> balanceMap, FruitTransaction transaction) {
+        String fruit = transaction.getFruit();
+        int quantity = transaction.getQuantity();
+        {
+
+            switch (transaction.getOperation()) {
+                case BALANCE -> balanceMap.put(fruit, quantity);
+                case SUPPLY, RETURN -> balanceMap.put(fruit,
+                        balanceMap.getOrDefault(fruit, 0) + quantity);
+                case PURCHASE -> balanceMap.put(fruit,
```
MateuszRostek, line 18: What if balance is < 0 after this operation? You should do some check for it.

MateuszRostek (approved):
I approved your solution, look at my comments and implement it (you can skip the strategy pattern). You will learn best and most when you write tests for this solution and see where there are possible errors and correct them yourself.

PR #1158, opened 2024-06-25
implement logic with solid rules

BelitskyiAlexandr:


src/main/java/service/impl/FruitTransactionParserImpl.java
```diff
@@ -0,0 +1,20 @@
+package service.impl;
```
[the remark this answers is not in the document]
BelitskyiAlexandr: it doesn't even pass checkstyle!

src/main/java/strategy/handlers/SupplyHandler.java
```diff
@@ -0,0 +1,20 @@
...
+public class SupplyHandler implements OperationHandler {
+    @Override
+    public void apply(FruitTransaction transaction) {
+        if (transaction == null) {
+            throw new IllegalArgumentException("Transaction cannot be null");
+        }
+
+        if (transaction.getQuantity() <= 0) {
+            throw new IllegalArgumentException("Quantity must be a positive number");
+        }
```
okuzan, line 16: these checks would make sense in each handler, let's create a utility method for this somewhere, maybe in operation handler, and reuse it across handlers

src/main/java/strategy/handlers/BalanceHandler.java
```diff
@@ -0,0 +1,21 @@
...
+        validateTransaction(transaction);
+        Storage.addFruit(transaction.getFruit(), transaction.getQuantity());
+    }
+
+    private void validateTransaction(FruitTransaction transaction) {
+        if (transaction.getQuantity() < 0) {
+            throw new IllegalArgumentException("Quantity cannot be negative: "
+                    + transaction.getQuantity());
+        }
+    }
```
okuzan, line 20: replace with a utility that can be used in each handler

src/main/java/strategy/handlers/BalanceHandler.java
```diff
@@ -0,0 +1,21 @@
+package strategy.handlers;
+
+import db.Storage;
+import model.FruitTransaction;
+import strategy.OperationHandler;
+
+public class BalanceHandler implements OperationHandler {
+
```
okuzan, line 8:
```suggestion
```
don't leave an empty line after method/class definition

src/main/java/service/impl/FruitTransactionParserImpl.java
```diff
@@ -0,0 +1,36 @@
...
+        try {
+            operation = FruitTransaction.Operation.fromCode(parts[0]);
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException("Invalid operation code: " + parts[0], e);
+        }
+
+        String fruit = parts[1];
+        int quantity;
+        try {
+            quantity = Integer.parseInt(parts[2]);
```
okuzan, line 29: would be nice to extract indexes to constants like OPERATION_INDEX

src/main/java/service/impl/FruitServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+    private final Map<FruitTransaction.Operation, OperationHandler> operationHandlers;
+
+    public FruitServiceImpl(Map<FruitTransaction.Operation, OperationHandler> operationHandlers) {
+        this.operationHandlers = operationHandlers;
+    }
+
+    @Override
+    public void applyTransaction(FruitTransaction transaction) {
+        OperationHandler handler = operationHandlers.get(transaction.getOperation());
+        if (handler != null) {
```
okuzan, line 19: in cases like this, always invert if to throw exception first, and put rest of the logic after it - guard clause pattern. It will be easier to read since you don't need else after `throw`/`return`

src/main/java/service/impl/CsvFileReaderServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+import java.nio.file.Paths;
+import java.util.List;
+import service.CsvFileReaderService;
+
+public class CsvFileReaderServiceImpl implements CsvFileReaderService {
+    @Override
+    public List<String> readFromFile(String filePath) {
+        try (var lines = Files.lines(Paths.get(filePath))) {
+            return lines
+                    .skip(1)
```
okuzan, line 14:
```suggestion
                    .skip(OFFSET)
```
let's make it more visible

src/main/java/Main.java
```diff
@@ -0,0 +1,59 @@
...
+        String inputFilePath = args[0];
+
+        CsvFileReaderService readerService = new CsvFileReaderServiceImpl();
+        FruitTransactionParser parser = new FruitTransactionParserImpl();
+
+        Map<FruitTransaction.Operation, OperationHandler> operationHandlers = new HashMap<>();
+        operationHandlers.put(FruitTransaction.Operation.BALANCE, new BalanceHandler());
+        operationHandlers.put(FruitTransaction.Operation.SUPPLY, new SupplyHandler());
+        operationHandlers.put(FruitTransaction.Operation.PURCHASE, new PurchaseHandler());
+        operationHandlers.put(FruitTransaction.Operation.RETURN, new ReturnHandler());
```
okuzan, line 36: you can use Map.of for brevity

src/main/java/Main.java
```diff
@@ -0,0 +1,59 @@
...
+
+        for (FruitTransaction transaction : transactions) {
+            fruitService.applyTransaction(transaction);
+        }
+
+        Map<String, Integer> reportData = fruitService.getReportData();
+        List<String> report = reportData.entrySet().stream()
+                .map(entry -> entry.getKey() + "," + entry.getValue())
+                .collect(Collectors.toList());
+        report.add(0, "fruit,quantity"); // add header line
```
okuzan, line 53: REPORT_HEADER

okuzan, line 53: this logic deserves its own service - ReportGenerator

src/main/java/Main.java
```diff
@@ -0,0 +1,59 @@
...
+        FruitService fruitService = new FruitServiceImpl(operationHandlers);
+
+        List<String> lines = readerService.readFromFile(inputFilePath);
+        List<FruitTransaction> transactions = lines.stream()
+                .map(parser::parse)
+                .toList();
+
+        for (FruitTransaction transaction : transactions) {
+            fruitService.applyTransaction(transaction);
+        }
```
okuzan, line 47: same for this - better encapsulate in a service(s). You can create another method in Parser for bulk parsing (list to a list)

okuzan (changes requested):
Good job, clean code!  Left some comments

src/main/java/util/TransactionValidator.java
```diff
@@ -0,0 +1,15 @@
...
+import model.FruitTransaction;
+
+public class TransactionValidator {
+    public static void validate(FruitTransaction transaction) {
+        if (transaction == null) {
+            throw new IllegalArgumentException("Transaction cannot be null");
+        }
+
+        if (transaction.getQuantity() <= 0) {
+            throw new IllegalArgumentException("Quantity must be a positive number");
```
okuzan, line 12:
```suggestion
            throw new IllegalArgumentException("Quantity must be a positive number, but was: " + transaction.getQuantity());
```

src/main/java/util/TransactionValidator.java
```diff
@@ -0,0 +1,15 @@
+package util;
+
+import model.FruitTransaction;
+
+public class TransactionValidator {
+    public static void validate(FruitTransaction transaction) {
```
okuzan, line 6: if it's a utility class, common convention is to add a private constructor to signal that no instantiation is supposed to happen for this class

src/main/java/Main.java
```diff
@@ -0,0 +1,56 @@
...
+
+        FruitService fruitService = new FruitServiceImpl(operationHandlers);
+
+        List<String> lines = readerService.readFromFile(inputFilePath);
+        List<FruitTransaction> transactions = parser.parseLines(lines);
+
+        fruitService.applyTransactions(transactions);
+
+        List<String> report = reportGenerator.generateReport(fruitService.getReportData());
+        report.add(0, REPORT_HEADER);
```
okuzan, line 52: this logic (52 line) belongs to a reportGenerator service itself

src/main/java/Main.java
```diff
@@ -0,0 +1,56 @@
...
+public class Main {
+    private static final String REPORT_HEADER = "fruit,quantity";
+
+    public static void main(String[] args) {
+        if (args.length != 2) {
+            throw new IllegalArgumentException("Please provide input and "
+                    + "output file paths as arguments.");
+        }
+
+        String inputFilePath = args[0];
```
okuzan, line 29:
```suggestion
        String inputFilePath = args[INPUT_FILE_PATH_ARG_INDEX];
```
and same for the second one

okuzan (changes requested):
Almost there!

src/main/java/db/Storage.java
```diff
@@ -0,0 +1,24 @@
+package db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    private static final Map<String, Integer> fruitStorage = new HashMap<>();
+
+    public static void addFruit(String fruit, int quantity) {
+        fruitStorage.put(fruit, fruitStorage.getOrDefault(fruit, 0) + quantity);
```
MentorMateAcademy, line 10: try to use method merge from map

src/main/java/service/impl/FruitTransactionParserImpl.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class FruitTransactionParserImpl implements FruitTransactionParser {
+    private static final int OPERATION_INDEX = 0;
+    private static final int FRUIT_INDEX = 1;
+    private static final int QUANTITY_INDEX = 2;
+
+    @Override
+    public FruitTransaction parse(String line) {
+        String[] parts = line.split(",");
+        if (parts.length != 3) {
```
MentorMateAcademy, line 15: can save to constant this value

MentorMateAcademy (changes requested):

MentorMateAcademy (changes requested):
Fix comments

src/main/java/service/impl/FruitTransactionParserImpl.java
MentorMateAcademy: Create constant for number 3, some limiter or smth else

src/main/java/service/impl/FruitTransactionParserImpl.java
```diff
@@ -0,0 +1,46 @@
...
+            throw new IllegalArgumentException("Invalid quantity: " + parts[QUANTITY_INDEX], e);
+        }
+
+        return new FruitTransaction(operation, fruit, quantity);
+    }
+
+    @Override
+    public List<FruitTransaction> parseLines(List<String> lines) {
+        return lines.stream()
+                .skip(1) // skip header line
```
MentorMateAcademy, line 42:
```suggestion
                .skip(1)

```

src/main/java/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import service.ReportGenerator;
+
+public class ReportGeneratorImpl implements ReportGenerator {
+    private static final String REPORT_HEADER = "fruit,quantity";
+
+    public List<String> generateReport(Map<String, Integer> reportData) {
+        List<String> report = reportData.entrySet().stream()
+                .map(entry -> entry.getKey() + "," + entry.getValue())
+                .collect(Collectors.toList());
+        report.add(0, REPORT_HEADER); // Add header line
```
MentorMateAcademy, line 15:
```suggestion
        report.add(0, REPORT_HEADER); 
```

MentorMateAcademy: don`t need comments

src/main/java/util/TransactionValidator.java
```diff
@@ -0,0 +1,21 @@
+package util;
+
+import model.FruitTransaction;
+
+public class TransactionValidator {
+
+    private TransactionValidator() {
+        throw new UnsupportedOperationException("Utility class");
+    }
```
MentorMateAcademy, line 9: Don`t do this anymore)

src/main/java/util/TransactionValidator.java
```diff
@@ -0,0 +1,21 @@
...
+    public static void validate(FruitTransaction transaction) {
+        if (transaction == null) {
+            throw new IllegalArgumentException("Transaction cannot be null");
+        }
+
+        if (transaction.getQuantity() <= 0) {
+            throw new IllegalArgumentException("Quantity must be a positive number, but was: "
+                + transaction.getQuantity());
+        }
+    }
```
MentorMateAcademy, line 20: why it static?

MentorMateAcademy (changes requested):

src/main/java/util/TransactionValidator.java
BelitskyiAlexandr: ok, but why? 
Is it better to make this class static so that no object of this class can be created?

It's just logical(maybe😁) that it's like an utility class, so you don't need to make an object of this class

sarakhmen: private constructors :)

src/main/java/util/TransactionValidator.java
sarakhmen: static method is what we try to avoid. It's better to have a class with a regular method

sarakhmen: it will be easier to maintain, test and follow best practices this way

sarakhmen (approved):
you can improve your solution if you want ;)

PR #1168, opened 2024-07-16
First solution

Wojtas1996PL:


src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.WriterService;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class WriterServiceImpl implements WriterService {
+    @Override
+    public void writeToCsv(String generatedReport, String filePath) {
```
MateuszRostek, line 9: Missing some checks before, like what if: 
`fileName == null
                || dataToWrite == null
                || !fileName.endsWith(".csv")`
                You are missing some checks + throwing exception in this case.

MateuszRostek (changes requested):
There are a lot of mistakes that need to be corrected, I have highlighted some of the most important ones, go through them and read carefully again the requirements and the most common mistakes. After the changes are made, I will review your solution again:)

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class DataConverterImpl implements DataConverter {
+
+    @Override
+    public List<FruitTransaction> convertToTransactions(List<String> csvLines) {
+        List<String> csvLinesCopy = csvLines;
+        csvLinesCopy.remove(0);
+        OperationHandlerSwitch operationHandlerSwitch = new OperationHandlerSwitchImpl();
+        List<FruitTransaction> listOfTransactions = new ArrayList<>();
+        for (int i = 0; i < csvLinesCopy.size(); i++) {
```
beatarjn, line 18: The code to extract the type, fruit, and quantity looks repeated for each line, consider extracting this logic into a single method to make the code more concise and reduce redundancy. Try something like this:

`
    for (String csvLine : csvLinesCopy) {
        String[] parts = csvLine.split(",");
        if (parts.length != 3) {
            throw new IllegalArgumentException("CSV line format is incorrect: " + csvLine);
        }

        String type = parts[0].trim();
        String fruit = parts[1].trim();
        int quantity = parseQuantity(parts[2].trim(), csvLine);

        listOfTransactions.add(new FruitTransaction(
            operationHandlerSwitch.getOperation(type), fruit, quantity));
    }

private int parseQuantity(String quantityStr, String csvLine) {
    int quantity;
    try {
        quantity = Integer.parseInt(quantityStr);
        throw new IllegalArgumentException("Invalid quantity value in line: " + csvLine, e);
    }
    if (quantity < 0) {
        throw new IllegalArgumentException("Quantity cannot be less than 0, actual quantity: " + quantity);
    }
    return quantity;
}`

src/main/java/core/basesyntax/service/impl/FruitTransactionServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public FruitTransactionServiceImpl(FruitShopDao fruitShopDao, FruitStrategy fruitStrategy) {
+        this.fruitShopDao = fruitShopDao;
+        this.fruitStrategy = fruitStrategy;
+    }
+
+    @Override
+    public Map<String, Integer> processTransactions(List<FruitTransaction> transactions) {
+        for (FruitTransaction transaction : transactions) {
+            fruitStrategy.getOperationHandler(transaction.getOperation())
```
beatarjn, line 22: Method getOperationHandler() by default is returning null. Please add check for handling possible NPE.

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+    @Override
+    public List<String> readFile(String filePath) {
+        if (filePath == null || !filePath.endsWith("csv")) {
+            throw new RuntimeException("Incorrect data");
+        }
+        List<String> list;
+        try {
+            list = Files.readAllLines(Path.of(filePath));
+        } catch (IOException ex) {
+            throw new RuntimeException();
```
beatarjn, line 19: Consider adding some descriptive message to the exception. It is a good practice to include the original exception's message and stack trace when rethrowing it. This helps in debugging.

src/main/java/core/basesyntax/service/impl/ReaderServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import core.basesyntax.service.ReaderService;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class ReaderServiceImpl implements ReaderService {
+    @Override
+    public List<String> readFile(String filePath) {
+        if (filePath == null || !filePath.endsWith("csv")) {
```
beatarjn, line 12: Consider extracting this checks to a separate method e.g
`private void validateFilePath(String filePath) {
        Objects.requireNonNull(filePath, "File path cannot be null");
        
        if (!filePath.endsWith(".csv")) {
            throw new IllegalArgumentException("File must be a CSV file");
        }

        if (!Files.exists(Path.of(filePath))) {
            throw new IllegalArgumentException("File does not exist: " + filePath);
        }
    }`

src/main/java/core/basesyntax/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,19 @@
...
+        if (generatedReport == null || filePath == null || !filePath.endsWith("csv")) {
```
beatarjn, line 10: Similarly as above please extract the checks to a method and use more descriptive messages.

src/main/java/core/basesyntax/strategy/impl/BalanceHandlerImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import core.basesyntax.utility.ListService;
+import core.basesyntax.utility.impl.ListServiceImpl;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class BalanceHandlerImpl implements OperationHandler {
+    private static List<FruitTransaction> balanceList = new ArrayList<>();
+    private Map<String, Integer> balanceFruitsMap;
+    private ListService listService = new ListServiceImpl();
```
beatarjn, line 15: Use dependency injection to pass dependencies like listService and balanceFruitsMap instead of instantiating them directly within the class.
This makes the class easier to test and follow the dependency inversion principle.
`private final Map<String, Integer> balanceFruitsMap;
    private final ListService listService;

    public BalanceHandlerImpl(ListService listService) {
        this.listService = listService;
        this.balanceFruitsMap = new HashMap<>();
    }
`

src/main/java/core/basesyntax/strategy/impl/FruitStrategyImpl.java
```diff
@@ -0,0 +1,37 @@
...
+                    return new SupplyHandlerImpl();
+                }
+                case PURCHASE -> {
+                    return new PurchaseHandlerImpl();
+                }
+                case RETURN -> {
+                    return new ReturnHandlerImpl();
+                }
+                default -> {
+                    return null;
```
beatarjn, line 32: Instead of returning null for unsupported operations, it's better to throw an exception to signal an invalid operation more explicitly or at least return an empty Object.

src/main/java/core/basesyntax/strategy/impl/PurchaseHandlerImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import core.basesyntax.utility.ListService;
+import core.basesyntax.utility.impl.ListServiceImpl;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class PurchaseHandlerImpl implements OperationHandler {
+    private static List<FruitTransaction> purchaseList = new ArrayList<>();
+    private Map<String, Integer> purchaseFruitsMap;
+    private ListService listService = new ListServiceImpl();
```
beatarjn, line 15: Same as above, pls use dependency injection to pass dependencies

src/main/java/core/basesyntax/strategy/impl/ReturnHandlerImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import core.basesyntax.utility.ListService;
+import core.basesyntax.utility.impl.ListServiceImpl;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class ReturnHandlerImpl implements OperationHandler {
+    private static List<FruitTransaction> returnList = new ArrayList<>();
+    private Map<String, Integer> returnFruitsMap;
+    private ListService listService = new ListServiceImpl();
```
beatarjn, line 15: Same as above, pls use dependency injection to pass dependencies

src/main/java/core/basesyntax/strategy/impl/SupplyHandlerImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import core.basesyntax.utility.ListService;
+import core.basesyntax.utility.impl.ListServiceImpl;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class SupplyHandlerImpl implements OperationHandler {
+    private static List<FruitTransaction> supplyList = new ArrayList<>();
+    private Map<String, Integer> supplyFruitsMap;
+    private ListService listService = new ListServiceImpl();
```
beatarjn, line 15: Same as above, pls use dependency injection to pass dependencies

src/main/java/core/basesyntax/utility/impl/ListServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class ListServiceImpl implements ListService {
+    @Override
+    public Map<String, Integer> getComputedMap(List<FruitTransaction> fruitTransactionList) {
+        Map<String, Integer> fruitTransactionComputed = new HashMap<>();
+        for (FruitTransaction fruitTransaction : fruitTransactionList) {
+            if (!fruitTransactionComputed.containsKey(fruitTransaction.getFruit())) {
```
beatarjn, line 14: You don't need to check if the key exists twice. Instead, you can streamline the code using the merge method provided by the Map interface. 'merge' will either add a new entry or update an existing entry in a single line. This method takes care of checking the existence of the key and updating the value appropriately.
`for (FruitTransaction fruitTransaction : fruitTransactionList) {
            fruitTransactionComputed.merge(fruitTransaction.getFruit(), 
                                           fruitTransaction.getQuantity(), 
                                           Integer::sum);
        }`

src/main/java/core/basesyntax/utility/impl/OperationHandlerSwitchImpl.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class OperationHandlerSwitchImpl implements OperationHandlerSwitch {
+    @Override
+    public FruitTransaction.Operation getOperation(String operation) {
+        return switch (operation) {
+            case "b" -> FruitTransaction.Operation.BALANCE;
+            case "s" -> FruitTransaction.Operation.SUPPLY;
+            case "p" -> FruitTransaction.Operation.PURCHASE;
+            case "r" -> FruitTransaction.Operation.RETURN;
+            default -> throw new IllegalStateException("Unexpected value: " + operation);
```
beatarjn, line 14: I would refactor the message to "No such operation avaialble"

beatarjn (changes requested):

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    private int parseQuantity(String quantityStr, String csvLine) {
+        int quantity;
+        try {
+            quantity = Integer.parseInt(quantityStr);
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Invalid quantity value in line: " + csvLine, e);
+        }
+        if (quantity < 0) {
```
beatarjn, line 40: This check should be done first, before parsing, so you won't parse a negative number.

src/main/java/core/basesyntax/main/Main.java
```diff
@@ -0,0 +1,62 @@
...
+        FruitTransactionService fruitTransactionService =
+                new FruitTransactionServiceImpl(fruitShopDao, fruitStrategy);
+        Map<String, Integer> fruitsQuantityAfterDay =
+                fruitTransactionService.processTransactions(fruitTransactions);
+
+        ReportGenerator reportGenerator = new ReportGeneratorImpl();
+        String generatedReport = reportGenerator.generateReport(fruitsQuantityAfterDay);
+
+        WriterService writerService = new WriterServiceImpl();
+        writerService.writeToCsv(generatedReport, "src/main/resources/outputFile.csv");
```
beatarjn, line 60: Avoid hardcoding the path for the output file. Consider using a configuration or define a constant.

beatarjn (changes requested):

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -31,16 +31,16 @@ public List<FruitTransaction> convertToTransactions(List<String> csvLines) {
     }
 
     private int parseQuantity(String quantityStr, String csvLine) {
+        if (Integer.parseInt(quantityStr) < 0) {
```
beatarjn, line 34: this could be simplified:
`try {
    int quantity = Integer.parseInt(quantityStr);
    if (quantity < 0) {
        throw new IllegalArgumentException("Quantity cannot be less than 0, actual quantity: " + quantity);
    }
    throw new IllegalArgumentException("Invalid quantity format: " + quantityStr, e);
}
`

beatarjn (approved):

PR #1186, opened 2024-08-21
Fruit Shop

4Vitalii5:
Implemented a program for processing transactions of a fruit shop, which includes reading input data, converting them, processing various operations (balance, purchase, return, supply), generating a report and writing the final result to a file.

### DIFF
--- a/src/main/java/Main.java
+++ b/src/main/java/Main.java
@@ -1,17 +1,20 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
+import dao.FruitDaoImpl;
 import model.FruitTransaction;
 import service.DataConverter;
 import service.ReaderService;
 import service.ReportGenerator;
 import service.ShopService;
 import service.WriterService;
-import service.imp.DataConverterImpl;
-import service.imp.ReaderServiceImpl;
-import service.imp.ReportGeneratorImpl;
-import service.imp.ShopServiceImpl;
-import service.imp.WriterServiceImpl;
+import service.impl.DataConverterImpl;
+import service.impl.DataValidatorServiceImpl;
+import service.impl.ReaderServiceImpl;
+import service.impl.ReportGeneratorImpl;
+import service.impl.ShopServiceImpl;
+import service.impl.WriterServiceImpl;
 import service.operation.BalanceOperation;
 import service.operation.OperationHandler;
 import service.operation.PurchaseOperation;
@@ -23,26 +26,29 @@ import strategy.OperationStrategyImpl;
 public class Main {
     private static final String FROM_FILE = "src/main/resources/reportToRead.csv";
     private static final String TO_FILE = "src/main/resources/finalReport.csv";
-    private static List<FruitTransaction> transactions;
 
     public static void main(String[] arg) {
         ReaderService fileReader = new ReaderServiceImpl();
         List<String> inputReport = fileReader.read(FROM_FILE);
 
-        DataConverter dataConverter = new DataConverterImpl();
-        transactions = dataConverter.convertToTransaction(inputReport);
-
         Map<FruitTransaction.Operation, OperationHandler> operationHandlers = new HashMap<>();
-        operationHandlers.put(FruitTransaction.Operation.BALANCE, new BalanceOperation());
-        operationHandlers.put(FruitTransaction.Operation.PURCHASE, new PurchaseOperation());
-        operationHandlers.put(FruitTransaction.Operation.RETURN, new ReturnOperation());
-        operationHandlers.put(FruitTransaction.Operation.SUPPLY, new SupplyOperation());
+        operationHandlers.put(FruitTransaction.Operation.BALANCE,
+                new BalanceOperation(new FruitDaoImpl()));
+        operationHandlers.put(FruitTransaction.Operation.PURCHASE,
+                new PurchaseOperation(new FruitDaoImpl()));
+        operationHandlers.put(FruitTransaction.Operation.RETURN,
+                new ReturnOperation(new FruitDaoImpl()));
+        operationHandlers.put(FruitTransaction.Operation.SUPPLY,
+                new SupplyOperation(new FruitDaoImpl()));
         OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlers);
+
+        DataConverter dataConverter = new DataConverterImpl(new DataValidatorServiceImpl());
+        List<FruitTransaction> transactions = dataConverter.convertToTransaction(inputReport);
 
         ShopService shopService = new ShopServiceImpl(operationStrategy);
         shopService.process(transactions);
 
-        ReportGenerator reportGenerator = new ReportGeneratorImpl();
+        ReportGenerator reportGenerator = new ReportGeneratorImpl(new FruitDaoImpl());
         String resultingReport = reportGenerator.getReport();
 
         WriterService writerService = new WriterServiceImpl();
--- a/src/main/java/Main.java
+++ b/src/main/java/Main.java
@@ -1,8 +1,7 @@
+import dao.FruitDaoImpl;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-
-import dao.FruitDaoImpl;
 import model.FruitTransaction;
 import service.DataConverter;
 import service.ReaderService;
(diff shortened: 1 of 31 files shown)


src/main/java/Main.java
```diff
@@ -0,0 +1,56 @@
...
+
+        Map<FruitTransaction.Operation, OperationHandler> operationHandlers = new HashMap<>();
+        operationHandlers.put(FruitTransaction.Operation.BALANCE,
+                new BalanceOperation(new FruitDaoImpl()));
+        operationHandlers.put(FruitTransaction.Operation.PURCHASE,
+                new PurchaseOperation(new FruitDaoImpl()));
+        operationHandlers.put(FruitTransaction.Operation.RETURN,
+                new ReturnOperation(new FruitDaoImpl()));
+        operationHandlers.put(FruitTransaction.Operation.SUPPLY,
+                new SupplyOperation(new FruitDaoImpl()));
```
okuzan, line 41: use map.of for brevity

src/main/java/dao/FruitDao.java
```diff
@@ -0,0 +1,15 @@
...
+
+public interface FruitDao {
+    Integer getBalance(String fruit);
+
+    boolean addBalance(String fruit, int quantity);
+
+    void updateBalance(String fruit, int quantity);
+
+    Set<Map.Entry<String, Integer>> getAllEntries();
+
```
okuzan, line 14:
```suggestion
```

src/main/java/dao/FruitDaoImpl.java
```diff
@@ -0,0 +1,30 @@
+package dao;
+
+import db.Storage;
+import java.util.Map;
+import java.util.Set;
+
+public class FruitDaoImpl implements FruitDao {
+
```
okuzan, line 8:
```suggestion
```

src/main/java/model/FruitTransaction.java
```diff
@@ -0,0 +1,67 @@
...
+        private String code;
+
+        Operation(String code) {
+            this.code = code;
+        }
+
+        public String getCode() {
+            return code;
+        }
+    }
```
okuzan, line 66: include method to parse string code into operation that would throw illegal arg exception if not found

src/main/java/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        checkReportLength(inputReport);
+        for (int i = REPORT_START_INDEX; i < inputReport.size(); i++) {
+            FruitTransaction fruitTransaction = new FruitTransaction();
+            String[] line = inputReport.get(i).split(DATA_SEPARATOR);
+            validator.validate(line);
+            fruitTransaction.setOperation(Arrays.stream(FruitTransaction.Operation.values())
+                    .filter(o -> o.getCode().equals(line[OPERATION_INDEX]))
+                    .findAny()
+                    .orElseThrow(() -> new NoSuchElementException("Invalid operation name: "
+                            + line[OPERATION_INDEX] + ". Must be one of these 'b,p,r,s'")));
```
okuzan, line 36: move this to enum
better remove must be one of these, you can forget to adapt this hardcode if you change the set of possible enum constant values

src/main/java/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,49 @@
...
+            fruitTransaction.setQuantity(Integer.parseInt(line[QUANTITY_INDEX]));
+            fruitTransactions.add(fruitTransaction);
+        }
+        return fruitTransactions;
+    }
+
+    private void checkReportLength(List<String> data) {
+        if (data.size() <= EMPTY_REPORT_LENGTH) {
+            throw new RuntimeException("This reportFile is empty");
+        }
```
okuzan, line 47: used only once, it's ok to just use it in place

src/main/java/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,49 @@
...
+            fruitTransaction.setFruit(line[FRUIT_INDEX]);
+            fruitTransaction.setQuantity(Integer.parseInt(line[QUANTITY_INDEX]));
+            fruitTransactions.add(fruitTransaction);
+        }
+        return fruitTransactions;
+    }
+
+    private void checkReportLength(List<String> data) {
+        if (data.size() <= EMPTY_REPORT_LENGTH) {
+            throw new RuntimeException("This reportFile is empty");
```
okuzan, line 46: a bit weird that you talk about file in the exception of the method that excepts list of strings :)

src/main/java/service/operation/PurchaseOperation.java
```diff
@@ -0,0 +1,23 @@
...
+    @Override
+    public void handle(String fruit, int quantity) {
+        int balanceAfterPurchase = fruitDao.getBalance(fruit) - quantity;
+        if (balanceAfterPurchase >= 0) {
+            fruitDao.updateBalance(fruit,balanceAfterPurchase);
+        } else {
+            throw new RuntimeException(String.format("You wanted to buy %d %ss "
+                    + "but available only %d %ss",quantity, fruit,fruitDao.getBalance(fruit),fruit
+            ));
+        }
```
okuzan, line 21: in cases like this, invert if condition to throw early - "guard clause pattern". We wont'need else after return/throw
https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

src/main/java/service/operation/ReturnOperation.java
```diff
@@ -0,0 +1,16 @@
...
+public class ReturnOperation implements OperationHandler {
+    private final FruitDao fruitDao;
+
+    public ReturnOperation(FruitDao fruitDao) {
+        this.fruitDao = fruitDao;
+    }
+
+    @Override
+    public void handle(String fruit, int quantity) {
+        fruitDao.updateBalance(fruit,fruitDao.getBalance(fruit) + quantity);
```
okuzan, line 14:
```suggestion
        fruitDao.updateBalance(fruit, fruitDao.getBalance(fruit) + quantity);
```
don't forget to Ctrl Alt L throughout your files

src/main/java/service/operation/SupplyOperation.java
```diff
@@ -0,0 +1,16 @@
...
+public class SupplyOperation implements OperationHandler {
+    private final FruitDao fruitDao;
+
+    public SupplyOperation(FruitDao fruitDao) {
+        this.fruitDao = fruitDao;
+    }
+
+    @Override
+    public void handle(String fruit, int quantity) {
+        fruitDao.updateBalance(fruit,fruitDao.getBalance(fruit) + quantity);
```
okuzan, line 14: same

okuzan (changes requested):
Good job on separating responsibilities, clean code!

src/main/java/service/impl/DataConverterImpl.java
4Vitalii5: Maybe this option will be better? Or is it better to remove this option?
public static Operation parseOperation(String code) {
            return Arrays.stream(FruitTransaction.Operation.values())
                    .filter(o -> o.getCode().equals(code))
                    .findAny()
                    .orElseThrow(() -> new IllegalArgumentException("Invalid operation name: "
                            + code + ". Must be one of these "
                            + Arrays.stream(Operation.values())
                            .map(Operation::getCode)
                            .collect(Collectors.joining(",", "{", "}"))));
        }

okuzan: ahah, nice! This way it's fine

src/main/java/service/impl/ShopServiceImpl.java
```diff
@@ -0,0 +1,20 @@
...
+public class ShopServiceImpl implements ShopService {
+    private final OperationStrategy operationStrategy;
+
+    public ShopServiceImpl(OperationStrategy operationStrategy) {
+        this.operationStrategy = operationStrategy;
+    }
+
+    @Override
+    public void process(List<FruitTransaction> transactions) {
+        transactions.forEach(f -> operationStrategy.get(f.getOperation())
```
okuzan, line 17: better replace f with transaction

okuzan (approved):
well done

src/main/java/service/impl/ShopServiceImpl.java
4Vitalii5: Thank you.

PR #1190, opened 2024-08-25
Implemented Shop

rainett:


src/main/.DS_Store
sarakhmen, line 1: remove this file from version control

src/main/java/core/basesyntax/model/enums/Operation.java
```diff
@@ -0,0 +1,26 @@
...
+    BALANCE("b"), SUPPLY("s"), PURCHASE("p"), RETURN("r");
+
+    private final String code;
+
+    Operation(String code) {
+        this.code = code;
+    }
+
+    public static Operation fromCode(String code) {
+        String exceptionMessage = "Operation with code = [" + code + "] not found";
```
sarakhmen, line 16: let's use this exception message directly without creating an additional variable. Fix same issue in all places

src/main/java/core/basesyntax/model/enums/Operation.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax.model.enums;
+
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
+public enum Operation {
+    BALANCE("b"), SUPPLY("s"), PURCHASE("p"), RETURN("r");
```
sarakhmen, line 7: put each enum on the new line

src/main/java/core/basesyntax/service/impl/CsvWriterImpl.java
```diff
@@ -0,0 +1,19 @@
...
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class CsvWriterImpl implements CsvWriter {
+    @Override
+    public void write(String fileName, String string) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(string);
+            writer.flush();
```
sarakhmen, line 13: you don't need to call flush explicitly. If I'm not wrong it'll be called automatically when the writer is going to be closed

src/main/java/core/basesyntax/strategy/impl/AbstractOperationHandler.java
```diff
@@ -0,0 +1,37 @@
...
+
+    protected AbstractOperationHandler(Operation targetOperation, MapValidator mapValidator) {
+        this.targetOperation = targetOperation;
+        this.mapValidator = mapValidator;
+    }
+
+    @Override
+    public void process(Map<String, Integer> fruitQuantityMap, FruitTransaction transaction) {
+        validateTransactionOperation(transaction);
+        mapValidator.validateMap(fruitQuantityMap, transaction);
```
sarakhmen, line 21: I think these validations are redundant at this point

sarakhmen (changes requested):
Fascinating solution! But let's slightly improve it :) Also I don't see you use something similar to the database/storage out here but I think you can stick to your approach as it stands out and is beautiful :)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,56 @@
...
+                .map(mapper::toObject)
+                .toList();
+
+        QuantityIncrementer incrementer = new QuantityIncrementerImpl();
+
+        Map<Operation, OperationHandler> strategiesMap = new HashMap<>();
+        strategiesMap.put(Operation.BALANCE, new BalanceOperationHandler());
+        strategiesMap.put(Operation.SUPPLY, new SupplyOperationHandler(incrementer));
+        strategiesMap.put(Operation.PURCHASE, new PurchaseOperationHandler());
+        strategiesMap.put(Operation.RETURN, new ReturnOperationHandler(incrementer));
```
okuzan, line 44: use map.of for brevity

okuzan (approved):
Nice!

PR #1243, opened 2024-11-12
Solution

vsupruniuk:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,68 @@
...
+
+        operationHandlerMap.put(
+                FruitTransaction.Operation.PURCHASE,
+                new PurchaseOperationHandler()
+        );
+
+        operationHandlerMap.put(
+                FruitTransaction.Operation.RETURN,
+                new ReturnOperationHandler()
+        );
```
liliia-ponomarenko, line 51: Let’s  move map initialization to the private method. Also, you can use Map.of()

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,68 @@
...
+
+        IOperationStrategy operationStrategy = new OperationStrategy(operationHandlerMap);
+
+        List<String> transactionsStrings = fileReader.readCsvFile(Paths.get("inputFile.csv"));
```
liliia-ponomarenko, line 55: better place files in the `resources`  directory

src/main/java/core/basesyntax/services/operationhandler/IOperationHandler.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax.services.operationhandler;
+
+public interface IOperationHandler {
```
liliia-ponomarenko, line 3: What does `I` prefix mean here?

src/main/java/core/basesyntax/services/reportgenerator/ReportGenerator.java
```diff
@@ -0,0 +1,26 @@
...
+    private static final String CSV_VALUES_SEPARATOR = ",";
+
+    @Override
+    public String getReport() {
+        Set<Map.Entry<String, Integer>> fruitEntries = DB.getFruitsDB().entrySet();
+        StringBuilder builder = new StringBuilder(REPORT_COLUMNS_NAMES);
+
+        for (Map.Entry<String, Integer> entry : fruitEntries) {
+            builder
+                    .append("\n")
```
liliia-ponomarenko, line 18: Better use System.lineseparator()

src/main/java/core/basesyntax/strategies/OperationStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    @Override
+    public IOperationHandler get(FruitTransaction.Operation operation) {
+        return switch (operation) {
+            case BALANCE -> operationHandlerMap.get(FruitTransaction.Operation.BALANCE);
+            case SUPPLY -> operationHandlerMap.get(FruitTransaction.Operation.SUPPLY);
+            case PURCHASE -> operationHandlerMap.get(FruitTransaction.Operation.PURCHASE);
+            case RETURN -> operationHandlerMap.get(FruitTransaction.Operation.RETURN);
+        };
```
liliia-ponomarenko, line 24: Don’t use a switch case here. Just: 
```suggestion
        return operationHandlerMap.get(operation);
```
Also, you can add null check and throw an exception if any handler wasn't found

liliia-ponomarenko (changes requested):
Good job! A few comments ;)

PR #1253, opened 2024-12-02
First approach to the fruit shop task

Pawel-Cp:


slade13 (approved):
Good job on implementing your Fruit Shop! There are some stylistic issues or SOLID violations but in overall for the first approach it is really good code. Don't be afraid of number of comments. :)

PR #1278, opened 2025-01-15
Fruit shop implemented with strategy pattern.

SSXcorp:


src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,41 @@
...
+        inventory.put(fruit, quantity);
+    }
+
+    public void addQuantity(String fruit, int quantity) {
+        inventory.put(fruit, getQuantity(fruit) + quantity);
+    }
+
+    public void subtractQuantity(String fruit, int quantity) {
+        inventory.put(fruit, getQuantity(fruit) - quantity);
+    }
```
ahoienko, line 36: This class should not have business logic, please move it to operation handlers

src/main/java/core/basesyntax/model/operation/PurchaseOperation.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax.model.operation;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.model.OperationHandler;
+
+public class PurchaseOperation implements OperationHandler {
```
ahoienko, line 7:
```suggestion
public class PurchaseOperationHandler implements OperationHandler { //same in other handlers
```

src/main/java/core/basesyntax/service/impl/CsvDataConverterImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            }
+            fruitTransactionsList.add(formFruitTransaction(line));
+        }
+        return fruitTransactionsList;
+    }
+
+    private FruitTransaction formFruitTransaction(String line) {
+        List<String> values = getRecordFromLine(line);
+        if (values.size() != VALUES_IN_LINE) {
+            throw new RuntimeException("Wrong number of values in line: " + line);
```
ahoienko, line 33:
```suggestion
            throw new RuntimeException("Wrong data format: " + line);
```

src/main/java/core/basesyntax/service/impl/CsvDataConverterImpl.java
```diff
@@ -0,0 +1,54 @@
...
+        List<String> records = new ArrayList<>();
+        try (Scanner rowScanner = new Scanner(line)) {
+            rowScanner.useDelimiter(COMMA_DELIMITER);
+            while (rowScanner.hasNext()) {
+                records.add(rowScanner.next());
+            }
+        }
+        return records;
+    }
+
```
ahoienko, line 53:
```suggestion
```

src/main/java/core/basesyntax/service/impl/CsvDataConverterImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    private List<String> getRecordFromLine(String line) {
+        List<String> records = new ArrayList<>();
+        try (Scanner rowScanner = new Scanner(line)) {
+            rowScanner.useDelimiter(COMMA_DELIMITER);
+            while (rowScanner.hasNext()) {
+                records.add(rowScanner.next());
+            }
+        }
+        return records;
+    }
```
ahoienko, line 52: What is the reason to create such complicated method? Is it possible to just line.split(",")?

src/main/java/core/basesyntax/service/impl/CsvFileReaderImpl.java
```diff
@@ -0,0 +1,24 @@
...
+    public List<String> read(String filename) {
+        List<String> lines = new ArrayList<>();
+        try (Scanner scanner = new Scanner(new File(filename))) {
+            while (scanner.hasNextLine()) {
+                lines.add(scanner.nextLine());
+            }
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("Cannot read file: " + filename, e);
+        }
+        return lines;
```
ahoienko, line 22: I believe scanner is not the best option, it worth to use scanner when data is red from console, but here we should work with files, better to use something like that https://www.baeldung.com/reading-file-in-java#%20id=

src/main/java/core/basesyntax/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,24 @@
...
+    private final Storage storage;
+
+    public ReportGeneratorImpl(Storage storage) {
+        this.storage = storage;
+    }
+
+    @Override
+    public String getReport() {
+        StringBuilder report = new StringBuilder();
+        report.append(BASE_REPORT_FIELDS).append("\n");
```
ahoienko, line 18: No need to append each time, you can add "\n" to constant

ahoienko, line 18: Better to use System.lineSeparator()

src/main/java/core/basesyntax/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,24 @@
...
+        for (Map.Entry<String, Integer> entry : storage.getInventory().entrySet()) {
+            report.append(entry.getKey()).append(",").append(entry.getValue()).append("\n");
```
ahoienko, line 20:
```suggestion
            report.append(entry.getKey())
            .append(COMMA_DELIMITER)
            .append(entry.getValue())
            .append(System.lineSeparator());
```

ahoienko (changes requested):
Very nice, you did great job! Left make little improvements

src/main/java/core/basesyntax/service/impl/CsvDataConverterImpl.java
SSXcorp: I decided to create separate methods and to get:
- list of strings from csv at convertToTransaction method
- than get list of params at getRecordFromLine method 
- and form FruitTransaction object at formFruitTransaction method

It could be usefull when we have another amount of params so we need to change only formFruitTransaction method. But otherwise I can agree with you

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax.db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    private static final int DEFAULT_INVENTORY_VALUE = 0;
+    private static Storage instance;
+    private final Map<String, Integer> inventory;
```
Elena-Bruyako, line 9: you can use just 
`public static final Map<String, Integer> storage = new HashMap<>();` and remove other code

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,64 @@
...
+    private String fruit;
+    private int quantity;
+
+    public String getFruit() {
+        return fruit;
+    }
+
+    public void setFruit(String fruit) {
+        if (fruit == null || fruit.isEmpty()) {
+            throw new NullPointerException("Fruit cannot be null or empty");
```
Elena-Bruyako, line 15: don't throw NullPointerException

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/model/operation/PurchaseOperationHandler.java
```diff
@@ -0,0 +1,22 @@
...
+    public PurchaseOperationHandler(Storage storage) {
+        this.storage = storage;
+    }
+
+    @Override
+    public void handle(FruitTransaction transaction) {
+        int currentAmount = storage.getInventory().getOrDefault(transaction.getFruit(),
+                DEFAULT_AMOUNT);
+        storage.getInventory().put(transaction.getFruit(),
+                currentAmount - transaction.getQuantity());
```
liliia-ponomarenko, line 20: is it correct logic? can you purchase fruits if it doesn't exist in storage or the amount isn't enough?

src/main/java/core/basesyntax/model/operation/ReturnOperationHandler.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    @Override
+    public void handle(FruitTransaction transaction) {
+        int currentAmount = storage.getInventory().getOrDefault(transaction.getFruit(),
+                DEFAULT_AMOUNT);
+        storage.getInventory().put(transaction.getFruit(),
+                transaction.getQuantity() + currentAmount);
+    }
+
```
liliia-ponomarenko, line 22:
```suggestion
```

src/main/java/core/basesyntax/service/impl/CsvDataConverterImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    public List<FruitTransaction> convertToTransaction(List<String> lines) {
+        List<FruitTransaction> fruitTransactionsList = new ArrayList<>();
+        boolean isFirstLine = true;
+        for (String line : lines) {
+            if (isFirstLine) {
+                isFirstLine = false;
+                continue;
+            }
+            fruitTransactionsList.add(formFruitTransaction(line));
+        }
```
liliia-ponomarenko, line 25:
```suggestion
        for (int i = 1; i < lines.length; i++) {
            fruitTransactionsList.add(formFruitTransaction(lines[i]));
        }
```
Let's simplify it

src/main/java/core/basesyntax/service/impl/CsvFileWriterImpl.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.FileWriter;
+import java.io.BufferedWriter;
+import java.io.IOException;
+
+public class CsvFileWriterImpl implements FileWriter {
+    @Override
+    public void write(String content, String filename) {
+        try (BufferedWriter writer = new BufferedWriter(new java.io.FileWriter(filename))) {
```
liliia-ponomarenko, line 10: add import

src/main/java/core/basesyntax/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public void execute(FruitTransaction transaction) {
+        OperationHandler handler = operationHandlers.get(transaction.getOperation());
+        if (handler != null) {
+            handler.handle(transaction);
+        } else {
+            throw new UnsupportedOperationException("Operation not supported: "
+                    + transaction.getOperation());
+        }
```
liliia-ponomarenko, line 23:
```suggestion
        if (handler == null) {
         throw new UnsupportedOperationException("Operation not supported: "
                    + transaction.getOperation());  
        } 
        handler.handle(transaction); 
```
Let's avoid if-else construction

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,64 @@
...
+    public int getQuantity() {
+        return quantity;
+    }
+
+    public void setQuantity(int quantity) {
+        if (quantity < ZERO) {
+            throw new RuntimeException("Quantity cannot be negative");
+        }
+        this.quantity = quantity;
+    }
```
liliia-ponomarenko, line 29: don't complicate the setter method. You should validate it in the PurcheseOperationHandler

liliia-ponomarenko (changes requested):
Good job! Let's improve your solution ;)

src/main/java/core/basesyntax/service/impl/CsvFileWriterImpl.java
SSXcorp: What do you mean? I already did FileWriter import. I can't do it twice

src/main/java/core/basesyntax/model/operation/ReturnOperationHandler.java
SSXcorp: > 

?

Elena-Bruyako (approved):

PR #1286, opened 2025-01-29
Implement Report generator

euulu:


src/main/java/core/basesyntax/service/impl/CsvDataConverter.java
```diff
@@ -0,0 +1,44 @@
...
+            Operation operation;
+            try {
+                operation = Operation.getOperationByCode(operationCode);
+            } catch (IllegalArgumentException e) {
+                throw new RuntimeException(e.getMessage(), e);
+            }
+            String fruit = parts[1].trim();
+            if (fruit.isEmpty()) {
+                continue;
+            }
```
liliia-ponomarenko, line 29: is it expected behavior? you should throw an exception if you have invalid data in the input file

src/main/java/core/basesyntax/service/impl/CsvDataConverter.java
```diff
@@ -0,0 +1,44 @@
...
+            int quantity;
+            try {
+                quantity = Integer.parseInt(parts[2].trim());
+            } catch (NumberFormatException e) {
+                continue;
```
liliia-ponomarenko, line 34: the same

src/main/java/core/basesyntax/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.stream.Collectors;
+
+public class ReportGeneratorImpl implements ReportGenerator {
+    @Override
+    public String generateReport(Map<String, Integer> inventory) {
+        StringBuilder reportBuilder = new StringBuilder("fruit,quantity\n");
+        String reportBody = inventory.entrySet()
+                .stream()
+                .map(entry -> entry.getKey() + "," + entry.getValue())
+                .collect(Collectors.joining("\n"));
```
liliia-ponomarenko, line 14: move literals to the constant fields. Use System.lineseparator

src/main/java/core/basesyntax/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.ReportGenerator;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class ReportGeneratorImpl implements ReportGenerator {
+    @Override
+    public String generateReport(Map<String, Integer> inventory) {
```
liliia-ponomarenko, line 9:
```suggestion
    public String generateReport() {
```
Don't pass storage as a parameter. You can call it directly

src/main/java/core/basesyntax/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public OperationStrategyImpl(
+            Map<FruitTransaction.Operation, OperationHandler> operationHandlers
+    ) {
+        this.operationHandlers = operationHandlers;
+    }
+
+    @Override
+    public OperationHandler getOperationHandler(FruitTransaction.Operation operation) {
+        return operationHandlers.get(operation);
```
liliia-ponomarenko, line 18: better add a null check for the handler. and throw an exception if the handler doesn’t  exist

liliia-ponomarenko (changes requested):
Good job! Let’s  improve your solution ;)

Elena-Bruyako (approved):

PR #1313, opened 2025-03-25
completed the task

opetrychenko:
completed the fruit shop report builder implementing the SOLID approach

### DIFF
--- a/src/main/java/core/basesyntax/Main.java
+++ b/src/main/java/core/basesyntax/Main.java
@@ -1,0 +1,54 @@
+package core.basesyntax;
+
+import db.Storage;
+import java.io.FileNotFoundException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import model.Transaction;
+import service.Converter;
+import service.FileReader;
+import service.ReportGenerator;
+import service.ShopService;
+import service.impl.ConverterImpl;
+import service.impl.FileReaderImpl;
+import service.impl.FileWriterImpl;
+import service.impl.OperationStrategyImpl;
+import service.impl.ReportGeneratorImpl;
+import service.impl.ShopServiceImpl;
+import service.operation.BalanceOperation;
+import service.operation.OperationHandler;
+import service.operation.PurchaseOperation;
+import service.operation.ReturnOperation;
+import service.operation.SupplyOperation;
+import strategy.OperationStrategy;
+
+public class Main {
+    private static final String INPUT_FILE = "src/main/reports/inputReport.csv";
+    private static final String OUTPUT_FILE = "src/main/reports/outputReport.csv";
+
+    public static void main(String[] args) throws FileNotFoundException {
+        FileReader fileReader = new FileReaderImpl();
+        List<String> inputReport = fileReader.read(INPUT_FILE);
+
+        Converter converter = new ConverterImpl();
+
+        Map<Transaction.Operation, OperationHandler> operationHandlerMap = new HashMap<>();
+        operationHandlerMap.put(Transaction.Operation.BALANCE, new BalanceOperation());
+        operationHandlerMap.put(Transaction.Operation.SUPPLY, new SupplyOperation());
+        operationHandlerMap.put(Transaction.Operation.PURCHASE, new PurchaseOperation());
+        operationHandlerMap.put(Transaction.Operation.RETURN, new ReturnOperation());
+        OperationStrategy operationStrategy = new OperationStrategyImpl(operationHandlerMap);
+
+        List<Transaction> transactionList = converter.convertTransaction(inputReport);
+        ShopService shopService = new ShopServiceImpl(operationStrategy);
+        shopService.procedure(transactionList);
+
+        Storage storage = new Storage();
+        ReportGenerator reportGenerator = new ReportGeneratorImpl(storage);
+        String reportResult = reportGenerator.getReport();
+
+        FileWriterImpl fileWriter = new FileWriterImpl();
+        fileWriter.write(OUTPUT_FILE, reportResult);
+    }
+}
--- a/src/main/java/db/Storage.java
+++ b/src/main/java/db/Storage.java
@@ -1,0 +1,27 @@
+package db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    public static final Map<String, Integer> fruitList = new HashMap<>();
+
+    public static void add(String fruit, Integer amount) {
+        fruitList.put(fruit, fruitList.getOrDefault(fruit, 0) + amount);
+    }
+
+    public static void remove(String fruit, Integer amount) {
+        if (!fruitList.containsKey(fruit) || fruitList.get(fruit) < amount) {
+            throw new IllegalArgumentException("There's not enough " + fruit + " in storage.");
+        }
+        fruitList.put(fruit, fruitList.get(fruit) - amount);
+    }
+
+    public static int getAmount(String fruit) {
+        return fruitList.getOrDefault(fruit, 0);
+    }
+
+    public Map<String, Integer> getAllFruit() {
+        return new HashMap<>(fruitList);
+    }
+}
--- a/src/main/java/service/FileReader.java
+++ b/src/main/java/service/FileReader.java
@@ -1,0 +1,8 @@
+package service;
+
+import java.io.FileNotFoundException;
+import java.util.List;
+
+public interface FileReader {
+    List<String> read(String fileName) throws FileNotFoundException;
+}
--- a/src/main/java/service/ShopService.java
+++ b/src/main/java/service/ShopService.java
@@ -1,0 +1,8 @@
+package service;
+
+import java.util.List;
+import model.Transaction;
+
+public interface ShopService {
+    void procedure(List<Transaction> transactionList);
+}
--- a/src/main/java/service/impl/ConverterImpl.java
+++ b/src/main/java/service/impl/ConverterImpl.java
@@ -1,0 +1,37 @@
+package service.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+import model.Transaction;
+import service.Converter;
+
+public class ConverterImpl implements Converter {
+    private static final String SEPARATOR = ",";
+    private static final int OPERATION_PARAMETER = 0;
+    private static final int FRUIT_PARAMETER = 1;
+    private static final int QUANTITY_PARAMETER = 2;
+
+    @Override
+    public List<Transaction> convertTransaction(List<String> report) {
+        List<Transaction> transactionList = new ArrayList<>();
+        for (int i = 1; i < report.size(); i++) {
+            Transaction transaction = getFromCsvRow(report.get(i));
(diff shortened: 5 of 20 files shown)


src/main/java/service/FileReader.java
```diff
@@ -0,0 +1,8 @@
+package service;
+
+import java.io.FileNotFoundException;
+import java.util.List;
+
+public interface FileReader {
+    List<String> read(String fileName) throws FileNotFoundException;
```
liliia-ponomarenko, line 7: Don’t  throw exception in the method signature

src/main/java/service/ShopService.java
```diff
@@ -0,0 +1,8 @@
+package service;
+
+import java.util.List;
+import model.Transaction;
+
+public interface ShopService {
+    void procedure(List<Transaction> transactionList);
```
liliia-ponomarenko, line 7:
```suggestion
    void process(List<Transaction> transactionList);
```

src/main/java/service/impl/ConverterImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    private Transaction getFromCsvRow(String row) {
+        String[] fields = row.split(SEPARATOR);
+        Transaction transaction = new Transaction();
+        transaction.setOperation(Transaction.Operation
+                .operationFromCode(fields[OPERATION_PARAMETER]));
+        transaction.setFruit(fields[FRUIT_PARAMETER]);
+        try {
+            transaction.setQuantity(Integer.parseInt(fields[QUANTITY_PARAMETER]));
+        } catch (NumberFormatException e) {
+            throw new RuntimeException("Cannot define quantity. Invalid number found.");
```
liliia-ponomarenko, line 33: provide parameter 
```suggestion
            throw new RuntimeException("Cannot define quantity. Invalid number found: " + fields[QUANTITY_PARAMETER]);
```

src/main/java/service/impl/FileReaderImpl.java
```diff
@@ -0,0 +1,28 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+import service.FileReader;
+
+public class FileReaderImpl implements FileReader {
+
+    public FileReaderImpl() {
+
+    }
```
liliia-ponomarenko, line 14: redundant constructor

src/main/java/service/impl/FileWriterImpl.java
```diff
@@ -0,0 +1,17 @@
+package service.impl;
+
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class FileWriterImpl {
+    public void write(String filePath, String content) {
+        if (!AutoCloseable.class.isAssignableFrom(FileWriter.class)) {
+            throw new IllegalStateException("FileWriter does not implement AutoClosable.");
+        }
```
liliia-ponomarenko, line 10: Do you really need this check?

src/main/java/service/impl/OperationStrategyImpl.java
```diff
@@ -0,0 +1,19 @@
...
+public class OperationStrategyImpl implements OperationStrategy {
+    private Map<Transaction.Operation, OperationHandler> operationHandlerMap;
+
+    public OperationStrategyImpl(Map<Transaction.Operation, OperationHandler> operationHandlerMap) {
+        this.operationHandlerMap = operationHandlerMap;
+    }
+
+    @Override
+    public OperationHandler get(Transaction.Operation operationType) {
+        return operationHandlerMap.getOrDefault(operationType, null);
```
liliia-ponomarenko, line 17: Better throw exception if handler doesn’t  exist

src/main/java/service/operation/BalanceOperation.java
```diff
@@ -0,0 +1,12 @@
+package service.operation;
+
+import db.Storage;
+import model.Transaction;
+
+public class BalanceOperation implements OperationHandler {
+    @Override
+    public void performOperation(Transaction transaction) {
+        Storage storage = new Storage();
```
liliia-ponomarenko, line 9: Make it a class-level variable (fix it everywhere)

liliia-ponomarenko (changes requested):
Good job! Let’s  improve your solution ;)

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final String INPUT_FILE = "src/main/resources/inputReport.csv";
+    private static final String OUTPUT_FILE = "src/main/resources/outputReport.csv";
+
+    public static void main(String[] args) throws FileNotFoundException {
+        FileReader fileReader = new FileReaderImpl();
+        List<String> inputReport = fileReader.read(INPUT_FILE);
+
+        Converter converter = new ConverterImpl();
+
+        Storage storage = new Storage();
```
Elena-Bruyako, line 36:
```suggestion
```

src/main/java/db/Storage.java
```diff
@@ -0,0 +1,31 @@
+package db;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+    private static Map<String, Integer> fruitList = new HashMap<>();
```
Elena-Bruyako, line 7:
```suggestion
    public static final Map<String, Integer> fruitList = new HashMap<>();
```

src/main/java/db/Storage.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public Map<String, Integer> getFruitList() {
+        return fruitList;
+    }
```
Elena-Bruyako, line 11:
```suggestion
```

src/main/java/service/impl/FileWriterImpl.java
```diff
@@ -0,0 +1,14 @@
+package service.impl;
+
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class FileWriterImpl {
+    public void write(String filePath, String content) {
+        try (FileWriter fileWriter = new FileWriter(filePath)) {
```
Elena-Bruyako, line 8: check if you can use FileWriter and fileWriter.write(content); in try with resources, if it implements Closeable interface?

src/main/java/service/operation/PurchaseOperation.java
```diff
@@ -0,0 +1,17 @@
+package service.operation;
+
+import db.Storage;
+import model.Transaction;
+
+public class PurchaseOperation implements OperationHandler {
+    private Storage storage = new Storage();
```
Elena-Bruyako, line 7: you already have storage Map, don't need to create new one

src/main/java/service/operation/SupplyOperation.java
```diff
@@ -0,0 +1,17 @@
+package service.operation;
+
+import db.Storage;
+import model.Transaction;
+
+public class SupplyOperation implements OperationHandler {
+    private Storage storage = new Storage();
```
Elena-Bruyako, line 7: same comment for all handlers

Elena-Bruyako (changes requested):

src/main/java/service/operation/PurchaseOperation.java
opetrychenko: But an operation cannot be performed without a Storage object and using a static method of the Storage class was disapproved during AI check. What needs to be done here then?

src/main/java/service/impl/FileWriterImpl.java
opetrychenko: Checked here. FieWriter does implement Closeable

src/main/java/service/operation/PurchaseOperation.java
Elena-Bruyako: You have `public static final Map<String, Integer> fruitList = new HashMap<>();`
this is like database, you should use it everywhere where is needed
don't create other storages

Elena-Bruyako (changes requested):
answer to you question

src/main/java/db/Storage.java
```diff
@@ -0,0 +1,27 @@
...
+public class Storage {
+    public static final Map<String, Integer> fruitList = new HashMap<>();
+
+    public static void add(String fruit, Integer amount) {
+        fruitList.put(fruit, fruitList.getOrDefault(fruit, 0) + amount);
+    }
+
+    public static void remove(String fruit, Integer amount) {
+        if (!fruitList.containsKey(fruit) || fruitList.get(fruit) < amount) {
+            throw new RuntimeException("There's not enough " + fruit + " in storage.");
```
JJJazl, line 15: Let's use more appropriate exceptions
```suggestion
            throw new IllegalArgumentException("There's not enough " + fruit + " in storage.");
```

src/main/java/service/impl/ConverterImpl.java
```diff
@@ -0,0 +1,38 @@
...
+        transaction.setOperation(Transaction.Operation
+                .operationFromCode(fields[OPERATION_PARAMETER]));
+        transaction.setFruit(fields[FRUIT_PARAMETER]);
+        try {
+            transaction.setQuantity(Integer.parseInt(fields[QUANTITY_PARAMETER]));
+        } catch (NumberFormatException e) {
+            throw new RuntimeException("Cannot define quantity. Invalid number found "
+                    + fields[QUANTITY_PARAMETER]);
+        }
+        return transaction;
```
JJJazl, line 36: What happens if row has an invalid structure, for example less than 3 values or more than 3 values. Let's check it also and throw an IllegalArgumentException with informative message in this case

src/main/java/service/impl/ConverterImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    private Transaction getFromCsvRow(String row) {
+        String[] fields = row.split(SEPARATOR);
+        Transaction transaction = new Transaction();
+        transaction.setOperation(Transaction.Operation
+                .operationFromCode(fields[OPERATION_PARAMETER]));
+        transaction.setFruit(fields[FRUIT_PARAMETER]);
+        try {
+            transaction.setQuantity(Integer.parseInt(fields[QUANTITY_PARAMETER]));
```
JJJazl, line 31: Check if quantity is not negative

src/main/java/service/impl/OperationStrategyImpl.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public OperationStrategyImpl(Map<Transaction.Operation, OperationHandler> operationHandlerMap) {
+        this.operationHandlerMap = operationHandlerMap;
+    }
+
+    @Override
+    public OperationHandler get(Transaction.Operation operationType) {
+        if (operationHandlerMap == null) {
+            throw new RuntimeException("Cannot retrieve operation handler map.");
+        }
```
JJJazl, line 19:
```suggestion
        if (operationHandlerMap.get(operationType) == null) {
            throw new IllegalArgumentException("Cannot retrieve operation handler by type: " + operationType.getOperationCode());
        }
```

JJJazl (approved):
Great! Left a few comments. Please, fix them before moving on to the next task)

PR #1371, opened 2025-09-29
readme completed

fn1z:


src/main/java/core/basesyntax/dao/FileWriterImpl.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class FileWriterImpl implements FileWriter {
+    @Override
+    public void write(String data, String fileName) {
+        if (fileName == null || fileName.isBlank()) {
+            throw new FruitShopException("Output file name is null or blank");
+        }
+        if (data == null) {
+            throw new FruitShopException("Data to write is null");
+        }
```
sarakhmen: this validation is redundant. It pollutes the code and actually for data validation there should be a separate layer (you can skip it for now). Also the exception will occur inside try catch probably and the exception will be thrown in any case

src/main/java/core/basesyntax/dao/FileReaderImpl.java
```diff
@@ -0,0 +1,20 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class FileReaderImpl implements FileReader {
+    @Override
+    public List<String> read(String fileName) {
+        if (fileName == null || fileName.isBlank()) {
+            throw new FruitShopException("Input file name is null or blank");
+        }
```
sarakhmen: same here, please fix in all places

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public FruitTransaction(Operation operation, String fruit, int quantity) {
+        this.operation = Objects.requireNonNull(operation, "operation is null");
+        if (fruit == null || fruit.isBlank()) {
+            throw new FruitShopException("fruit is null or blank");
+        }
+        if (quantity < 0) {
+            throw new FruitShopException("quantity is negative: " + quantity);
+        }
+        this.fruit = fruit;
```
sarakhmen: you're overvalidating your code. It shouldn't be a responsibility of this constructor

src/main/java/core/basesyntax/model/Operation.java
```diff
@@ -0,0 +1,27 @@
...
+        this.code = code;
+    }
+
+    public String getCode() {
+        return code;
+    }
+
+    public static Operation fromCode(String code) {
+        for (Operation op : Operation.values()) {
+            if (op.code.equals(code)) {
```
sarakhmen, line 21: to compare enums you can use == operator. There is only one instance of each one

src/main/java/core/basesyntax/service/DataConverterImpl.java
```diff
@@ -0,0 +1,57 @@
...
+            String line = rawLines.get(i).trim();
+            if (line.isEmpty()) {
+                continue;
+            }
+            String[] parts = line.split(COMMA);
+            if (parts.length != 3) {
+                throw new FruitShopException("Invalid record format (expected 3 values): " + line);
+            }
+            String opCode = parts[0].trim();
+            String fruit = parts[1].trim();
```
sarakhmen, line 45: consider moving hardcoded values/literals/magic numbers to constants

sarakhmen (changes requested):
please review your code and remove excessive validations in all places. There are dozens of such ones

src/main/java/core/basesyntax/dao/FileReaderImpl.java
fn1z: so I should also remove all of throw new FruitShopException("Fruit is null or blank in transaction: in all operation like BalanceOperation, PurchaseOperation etc?

KuryloVolodymyr (changes requested):
please resolve previous comments

sarakhmen (approved):

PR #1377, opened 2025-10-07
 fruit-shoP

MaksymMaksymenko2:


src/main/java/db/Storage.java
```diff
@@ -0,0 +1,29 @@
+package db;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Storage {
+
+    private static final Map<String, Integer> FRUIT_STORAGE = new HashMap<>();
```
sarakhmen: actually it's not a constant as the state is constantly changing inside the map, so we can make the naming in camel case

src/main/java/model/FruitTransaction.java
```diff
@@ -0,0 +1,64 @@
...
+    public FruitTransaction(Operation operation, String fruit, int quantity) {
+        if (operation == null) {
+            throw new IllegalArgumentException("Operation cannot be null");
+        }
+        if (fruit == null || fruit.isBlank()) {
+            throw new IllegalArgumentException("Fruit cannot be null or empty");
+        }
+        if (quantity < 0) {
+            throw new IllegalArgumentException("Quantity cannot be negative");
+        }
```
sarakhmen: no need to have these validations here. Let's not overcomplicate the code

src/main/java/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import model.FruitTransaction;
+import service.DataConverter;
+
+public class DataConverterImpl implements DataConverter {
+
+    @Override
+    public List<FruitTransaction> convertToTransactions(List<String> lines) {
```
sarakhmen, line 15: this method is ok in general. But I'd recommend moving magic numbers/ literals to the constants. And also it'd be great to split the logic on several private methods for readability and to follow clean code principles. Review your code again and fix similar issues in all places if needed

src/main/java/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,20 @@
...
+    @Override
+    public String getReport() {
+        StringBuilder report = new StringBuilder("fruit,quantity\n");
+        for (Map.Entry<String, Integer> entry : Storage.getSnapshot().entrySet()) {
+            report.append(entry.getKey())
+                    .append(",")
+                    .append(entry.getValue())
+                    .append("\n");
+        }
+        return report.toString();
```
sarakhmen: use system.lineseparator for line breaks

src/main/java/service/impl/ShopServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+        if (transactions == null) {
+            throw new IllegalArgumentException("Transactions cannot be null");
+        }
+
+        for (FruitTransaction transaction : transactions) {
+            OperationHandler handler = operationStrategy.getHandler(transaction.getOperation());
+            if (handler == null) {
+                throw new RuntimeException(
+                        "No handler found for operation: " + transaction.getOperation());
+            }
```
sarakhmen: I'd move this validation on strategy class level

src/main/java/service/impl/WriterServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+public class WriterServiceImpl implements WriterService {
+
+    @Override
+    public void writeToFile(String report, String filePath) {
+        if (report == null || report.isBlank()) {
+            throw new IllegalArgumentException("Report cannot be null or empty");
+        }
+        if (filePath == null || filePath.isBlank()) {
+            throw new IllegalArgumentException("File path cannot be null or empty");
+        }
```
sarakhmen: a lot of validations. Your code is polluted with them

src/main/java/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,34 @@
...
+import model.FruitTransaction;
+
+public class OperationStrategyImpl implements OperationStrategy {
+
+    private final Map<FruitTransaction.Operation, OperationHandler> handlers;
+
+    public OperationStrategyImpl(Map<FruitTransaction.Operation, OperationHandler> handlers) {
+        if (handlers == null) {
+            throw new IllegalArgumentException("Handlers map cannot be null");
+        }
```
sarakhmen: same here

sarakhmen: fix in all places

src/main/java/strategy/OperationStrategyImpl.java
```diff
@@ -0,0 +1,34 @@
...
+    @Override
+    public OperationHandler getHandler(FruitTransaction.Operation operation) {
+        if (operation == null) {
+            throw new IllegalArgumentException("Operation cannot be null");
+        }
+        OperationHandler handler = handlers.get(operation);
+        if (handler == null) {
+            throw new RuntimeException("No handler for operation: " + operation);
+        }
+        return handler;
```
sarakhmen: you already have handler!=null validation here so why do you duplicate this logic in other places?

pom.xml
```diff
@@ -18,6 +18,12 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>RELEASE</version>
+            <scope>compile</scope>
+        </dependency>
```
sarakhmen: revert it. No need to add additional dependencies for this task

sarakhmen (changes requested):
let's improve your solution :)

sarakhmen (approved):
good job!

PR #1401, opened 2025-11-09
FruitTransaction - program with a business logic 

SivForces:
Creating a working business logic program to make different types of operations. Read and write to file, check balance


src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,42 @@
...
+
+        Operation(String code) {
+            this.code = code;
+        }
+
+        public String getCode() {
+            return code;
+        }
+    }
+}
```
KuryloVolodymyr: Operation is often used outside of FruitTransaction so it makes more sense to move it to a dediacted file

src/main/java/core/basesyntax/service/impl/ConverterImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    private FruitTransaction.Operation parseOperation(String operationSymbol,
+                                                      String line, int lineIndex) {
+        for (FruitTransaction.Operation operation
+                : FruitTransaction.Operation.values()) {
+            if (operation.getCode().equalsIgnoreCase(operationSymbol)) {
+                return operation;
+            }
+        }
+        throw new RuntimeException("Invalid operation symbol '" + (lineIndex + 1));
+    }
```
KuryloVolodymyr: this looks like more of the Operation specific method. So its better to move it to Operation

src/main/java/core/basesyntax/service/impl/ShopServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class ShopServiceImpl implements ShopService {
+    private final OperationStrategy operationStrategy;
+
+    public ShopServiceImpl(OperationStrategy operationStrategy) {
+        this.operationStrategy = operationStrategy;
+    }
+
+    @Override
+    public void process(List<FruitTransaction> transactions) {
```
KuryloVolodymyr, line 17: Add validation for transactions, if its null

KuryloVolodymyr (changes requested):
see comment

Elena-Bruyako (changes requested):
provide input and report files

Elena-Bruyako (changes requested):
report file is empty, provide file with results data

SivForces:
I don't have report data in the main window because I have instaled a plugin ExcelReader to ues it with CSV files, so I don't really know if this is complete

src/main/resources/report.csv
```diff
@@ -0,0 +1,2 @@
+banana,152
```
Elena-Bruyako, line 1: add header (check task description)

Elena-Bruyako (approved):
see comment
approve

PR #1453, opened 2026-04-09
Solution

Rabico:


src/main/java/core/basesyntax/strategy/impl/SupplyOperation.java
```diff
@@ -0,0 +1,23 @@
...
+public class SupplyOperation implements DataOperation {
+    private StorageDao dao;
+
+    public SupplyOperation(StorageDao dao) {
+        this.dao = dao;
+    }
+
+    @Override
+    public void execute(FruitTransaction fruit) {
+        if (dao.checkFruit(fruit.getName())) {
```
slade13: `SupplyOperation` logic is currently reversed.

If the fruit **already exists**, you should update its quantity.
If it **does not exist yet**, you should add it.
In the current version, supply for an existing fruit will throw `"Fruit already exists"`.

src/main/java/core/basesyntax/strategy/impl/BalanceOperation.java
```diff
@@ -0,0 +1,18 @@
...
+public class BalanceOperation implements DataOperation {
+    private StorageDao dao;
+
+    public BalanceOperation(StorageDao dao) {
+        this.dao = dao;
+    }
+
+    @Override
+    public void execute(FruitTransaction fruit) {
+        dao.add(fruit);
```
slade13: `BalanceOperation` should not fail for repeated fruit entries. `dao.add(fruit)` throws an exception when the fruit already exists.
For this task, balance lines are just input operations that should update storage correctly. Even if the input contains the same fruit more than once, your logic should handle it safely instead of crashing.
A better approach is to make balance work like setting or increasing the initial amount consistently.

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,38 @@
...
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.DataConverter;
+import java.util.List;
+
+public class DataConverterImpl implements DataConverter {
+    @Override
+    public List<FruitTransaction> convertToTransaction(List<String> transactions) {
+        return transactions.stream().map(s -> s.split(","))
+                .map(s -> new FruitTransaction(s[0], s[1],
```
slade13: You created `parseInteger()` and `parseTransaction()`, but `convertToTransaction()` ignores them and uses:
```
.map(s -> s.split(","))
.map(s -> new FruitTransaction(s[0], s[1], Integer.parseInt(s[2])))
```

Because of that:

- transaction length is not validated,
- trimming is not applied,
- integer parsing is not handled through your custom validation.
- You should reuse `parseTransaction()` inside `convertToTransaction()`.

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return i;
+    }
+
+    private FruitTransaction parseTransaction(String[] s) {
+        if (s.length != 3) {
+            throw new IllegalArgumentException("wrong transaction length");
+        }
+        for (String line : s) {
+            line.trim();
```
slade13: This does nothing, because `String `is immutable and the trimmed result is not saved back.
If you want trimmed values, you must assign them, for example:
```
s[i] = s[i].trim();
```

src/main/java/core/basesyntax/db/StorageDaoImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    public void add(FruitTransaction fruit) {
+        if (fruits.containsKey(fruit.getName())) {
+            throw new IllegalArgumentException("Fruit already exists");
+        }
+        fruits.put(fruit.getName(), fruit.getQuantity());
+    }
+
+    @Override
+    public int actualQuantity(String fruitName) {
+        return fruits.get(fruitName);
```
slade13, line 17: Here the returned `Integer `is auto-unboxed to `int`. If the fruit is absent, this will fail with `NullPointerException`.
You should either validate presence before calling this method everywhere or make the DAO behavior safer.

src/main/java/core/basesyntax/db/StorageDaoImpl.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax.db;
+
+import core.basesyntax.model.FruitTransaction;
+import java.util.HashMap;
+import java.util.Map;
+
+public class StorageDaoImpl implements StorageDao {
+    private Map<String, Integer> fruits = new HashMap<String, Integer>();
```
slade13: Use interface types for fields and make them final where possible - for example, this line should be:

```suggestion
    private final Map<String, Integer> fruits = new HashMap<>();
```

src/main/java/core/basesyntax/service/impl/FileWriterImpl.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.service.FileWriter;
+import java.io.FileNotFoundException;
+import java.io.PrintWriter;
+
+public class FileWriterImpl implements FileWriter {
+
+    @Override
+    public void writeReport(String report, String fileName) {
```
slade13: `FileWriterImpl` should write to a relative path in resources or another explicit destination, not just a bare file name.
In Main you call:
```
fileWriter.writeReport(report, "Report.csv");
```

This writes to the current working directory, which may be inconsistent depending on where the app is run.
It is better to pass an explicit relative path.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,52 @@
...
+import core.basesyntax.strategy.impl.BalanceOperation;
+import core.basesyntax.strategy.impl.PurchaseOperation;
+import core.basesyntax.strategy.impl.ReturnOperation;
+import core.basesyntax.strategy.impl.SupplyOperation;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    public static void main(String[] args) {
```
slade13, line 32: These values:
```
"Data.csv"
"Report.csv"
```
should be extracted to constants in `Main`. This is a small issue, but it follows the “avoid hardcode” requirement better.

slade13, line 32: Method naming is not fully aligned with the task examples and intent:

- `readFile()` / `writeReport()` instead of more generic names like `read()` / `write()`
- `update()` in `StorageService `is not very expressive for processing transactions

src/main/java/core/basesyntax/strategy/impl/PurchaseOperation.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public PurchaseOperation(StorageDao dao) {
+        this.dao = dao;
+    }
+
+    @Override
+    public void execute(FruitTransaction fruit) {
+
+        if (!dao.checkFruit(fruit.getName())) {
+            throw new RuntimeException("The fruit " + fruit.getName() + " is not exist");
```
slade13:
```suggestion
            throw new RuntimeException("The fruit " + fruit.getName() + " does not exist");
```

src/main/java/core/basesyntax/strategy/impl/ReturnOperation.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public ReturnOperation(StorageDao dao) {
+        this.dao = dao;
+    }
+
+    @Override
+    public void execute(FruitTransaction fruit) {
+
+        if (!dao.checkFruit(fruit.getName())) {
+            throw new RuntimeException("The fruit " + fruit.getName() + " is not exist");
```
slade13:
```suggestion
            throw new RuntimeException("The fruit " + fruit.getName() + " does not exist");
```

src/main/java/core/basesyntax/model/FruitTransaction.java
```diff
@@ -0,0 +1,31 @@
...
+    private int quantity;
+    private Operation operation;
+
+    public FruitTransaction(String operation, String name, int quantity) {
+        if (name == null || name.isBlank()) {
+            throw new IllegalArgumentException("Fruit name cannot be null or blank");
+        }
+        this.name = name;
+        if (quantity < 0) {
+            throw new IllegalArgumentException("quantity must be positive");
```
slade13: "quantity must be positive" is misleading when `0` is allowed, because your condition only rejects values `< 0`
Error messages should be grammatically correct and match actual validation rules.

src/main/java/core/basesyntax/strategy/DataOperation.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.strategy;
+
+import core.basesyntax.model.FruitTransaction;
+
+public interface DataOperation {
```
slade13: The task expects Strategy pattern. You do have separate operation handlers, which is good, but `DataOperation `is not a very clear name for a strategy interface. Something like `OperationHandler `would better reflect its responsibility and align better with the task description.

src/main/java/core/basesyntax/service/impl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,29 @@
...
+public class ReportGeneratorImpl implements ReportGenerator {
+    private StorageDao dao;
+
+    public ReportGeneratorImpl(StorageDao dao) {
+        this.dao = dao;
+    }
+
+    @Override
+    public String generateReport() {
+        Map<String, Integer> fruits = dao.getData();
```
slade13: `ReportGeneratorImpl `depends on map iteration order, which is not guaranteed.
Since storage uses `HashMap`, the order of fruits in the report may vary.
The task example shows a stable report structure. It is safer to either:

- use `LinkedHashMap`
- sort entries before building the report

src/main/java/core/basesyntax/db/StorageDaoImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public void updateQuantity(String fruitName, int quantity) {
+        fruits.put(fruitName, quantity);
+    }
+
+    @Override
+    public Map<String, Integer> getData() {
+
+        return new HashMap<>(fruits);
+    }
```
slade13, line 28: Remove empty lines:
```suggestion
    public Map<String, Integer> getData() {
        return new HashMap<>(fruits);
    }
```

slade13 (changes requested):
There are some minor issues, and one critical issue in `SupplyOperation`, please review my comments, and update your code accordingly.

src/main/java/core/basesyntax/db/StorageDaoImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public void updateQuantity(String fruitName, int quantity) {
+        fruits.put(fruitName, quantity);
+    }
+
+    @Override
+    public Map<String, Integer> getData() {
+        return new LinkedHashMap<>();
```
slade13: This always returns an empty map instead of the actual storage contents.

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax.service.impl;
+
+import core.basesyntax.model.FruitTransaction;
+import core.basesyntax.service.DataConverter;
+import java.util.List;
+
+public class DataConverterImpl implements DataConverter {
+    @Override
+    public List<FruitTransaction> convertToTransaction(List<String> transactions) {
```
slade13, line 9: Your `DataConverterImpl` still does not use `parseTransaction()`.

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return i;
+    }
+
+    private FruitTransaction parseTransaction(String[] s) {
+        if (s.length != 3) {
+            throw new IllegalArgumentException("wrong transaction length");
+        }
+        for (String line : s) {
+            line = line.trim();
```
slade13: Trimming in `parseTransaction()` still does not work. `String` is immutable, and line is just a local variable.
Use indexing, for example:
```
for (int i = 0; i < s.length; i++) {
    s[i] = s[i].trim();
}
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class Main {
+    private static final String INPUT_FILE = "Data.csv";
+    private static final String OUTPUT_FILE = "Report.csv";
+
+    public static void main(String[] args) {
+        Map<Operation, OperationHandler> operations = new HashMap<>();
+        StorageDao dao = new StorageDaoImpl();
```
slade13: These fields should be final and private.

```suggestion
        private final Map<Operation, OperationHandler> operations = new HashMap<>();
        private final StorageDao dao = new StorageDaoImpl();
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,55 @@
...
+        StorageService storageService = new StorageServiceImpl(operations);
+        storageService.process(fruitTransactions);
+        // 4. Generate report
+        ReportGenerator generator = new ReportGeneratorImpl(dao);
+        String report = generator.generateReport();
+        // 5. Save report to file
+        FileWriter fileWriter = new FileWriterImpl();
+        fileWriter.write(report, OUTPUT_FILE);
+    }
+
```
slade13: Remove this empty line.
```suggestion
```

slade13 (changes requested):
Please review my comments and add your changes.

src/main/java/core/basesyntax/db/StorageDaoImpl.java
mateuszwojtkowiak, line 16: Please validate that the fruit exists in the Map before returning its quantity. 
`fruits.get(fruitName)` may return `null`, and auto-unboxing to `int` will cause a NullPointerException. 
Throw a descriptive RuntimeException when the fruit is missing.

src/main/java/core/basesyntax/model/FruitTransaction.java
mateuszwojtkowiak, line 8: The constructor should not accept the operation as a raw String. 
Parsing CSV codes (like "b", "s", "p", "r") belongs to the converter layer, not to the domain model. 
`FruitTransaction` should receive an already parsed `Operation` enum to keep the model clean and aligned with SRP. 
Please move the code-to-enum conversion into the DataConverter and pass `Operation` directly to this class.

src/main/java/core/basesyntax/service/impl/DataConverterImpl.java
mateuszwojtkowiak, line 26: `parseTransaction()` validates only the array length but does not validate empty fruit names or malformed operation codes.

mateuszwojtkowiak, line 15: `parseInteger()` contains an unnecessary temporary variable. 
You can return the parsed value directly inside the try block, which makes the method cleaner and easier to read. 
Also, when rethrowing the exception, include the original cause (`e`) so the stack trace is preserved:

```
try {
    return Integer.parseInt(s);
    throw new IllegalArgumentException("Invalid integer: " + s, e);
}
```

This improves clarity and provides more useful debugging information.

src/main/java/core/basesyntax/Main.java
```diff
@@ -26,10 +26,11 @@
 public class Main {
     private static final String INPUT_FILE = "Data.csv";
     private static final String OUTPUT_FILE = "Report.csv";
+    private static final Map<Operation, OperationHandler> operations = new HashMap<>();
+    private static final StorageDao dao = new StorageDaoImpl();
 
     public static void main(String[] args) {
```
mateuszwojtkowiak, line 32: Just like mentor said: your method names are too generic like `read() or write()`.

Wojtas1996PL (approved):
Good job! The task is accepted, but please read my comments about some things that could be improved and apply these changes.

PR #1471, opened 2026-05-14
did

Polya555:


src/main/java/core/basesyntax/serviceimpl/ReportGeneratorImpl.java
```diff
@@ -0,0 +1,25 @@
...
+public class ReportGeneratorImpl implements ReportGenerator {
+    private final Storage storage;
+    public ReportGeneratorImpl(Storage storage) {
+        this.storage = storage;
+    }
+
+    @Override
+    public String getReport() {
+        StringBuilder report = new StringBuilder("fruit,quantity\n");
+        for (Map.Entry<String, Integer> entry : storage.getAll().entrySet()) {
```
fetis-off:
```suggestion
        StringBuilder report = new StringBuilder(REPORT_HEADER);
        for (Map.Entry<String, Integer> entry : storage.getAll().entrySet()) {
```
try to avoid using magic strings in your code. Let's better define constant

src/main/java/core/basesyntax/BalanceOperation.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class BalanceOperation implements OperationHandler {
```
fetis-off:
```suggestion
public class BalanceOperationHandler implements OperationHandler {
```
Rename this component for better code readability. The same in other OperationHandler implementations

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,39 @@
...
+
+        ReportGenerator reportGenerator = new ReportGeneratorImpl(storage);
+        String resultingReport = reportGenerator.getReport();
+
+        FileWriter fileWriter = new FileWriter() {
+            @Override
+            public void write(String data, String fileName) {
+
+            }
+        };
```
fetis-off:
```suggestion
        FileWriter fileWriter = new FileWriterImpl();
```
why do you need anonymous class here?

Did you test your solution? It does not write data to an file, does it?

fetis-off (changes requested):
See comments

fetis-off (approved):
Good

